PR #2, opened 2014-10-07
config: add modulifile option to customize moduli file location (bug#2277)

alonbl:
Currently all files can be customized via sshd_config, however, the
moduli file cannot.

Running sshd in unprivileged context requires customization of all
resources, especially when some distributions sets the moduli as world
unreadable.


djmdjm:
Please use https://bugzilla.mindrot.org/ instead of github PRs. We only use this site as a mirror and seldom check the PR queue.

PR #3, opened 2014-10-07
AuthorizedKeysCommand: add an option for alternate owner (bug#2276)

alonbl:
Currently the owner of AuthorizedKeysCommand must be root.

A setup in which sshd is running as non root, can enjoy a complete
and secure environment even if the AuthorizedKeysCommand is owned by a
different user.

This patch adds AuthorizedKeysCommandOwner option to control the
ownership check of the AuthorizedKeysCommand. Default is root, so no
change is done without explicit request.

### DIFF
--- a/auth2-pubkey.c
+++ b/auth2-pubkey.c
@@ -529,10 +529,11 @@ user_key_command_allowed2(struct passwd *user_pw, Key *key)
 	FILE *f;
 	int ok, found_key = 0;
 	struct passwd *pw;
+	uid_t owneruid = 0;
 	struct stat st;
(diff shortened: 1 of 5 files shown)


alonbl:
Rebased

djmdjm:
We don't use pull requests at the moment - please file patches
on bugzilla.

alonbl:
On Tue, Mar 3, 2015 at 7:06 PM, Damien Miller <notifications@github.com> wrote:
>
> We don't use pull requests at the moment - please file patches
> on bugzilla.

Done.

PR #4, opened 2014-10-10
document evaluation of {Allow|Deny}{Users|Groups}

calestyo:
• Document what the evaluation order of AllowUsers, DenyUsers, AllowGroups and
  DenyGroups actually means.
  Fixes bug #2292.

Signed-off-by: Christoph Anton Mitterer mail@christoph.anton.mitterer.name

### DIFF
--- a/sshd_config.5
+++ b/sshd_config.5
@@ -116,6 +116,8 @@ The allow/deny directives are processed in the following order:
 .Cm DenyGroups ,
 and finally
 .Cm AllowGroups .
+The first one that matches determines whether the login is allowed or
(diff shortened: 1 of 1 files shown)


calestyo:
Forwarded to https://lists.mindrot.org/pipermail/openssh-unix-dev/2015-February/033465.html

djmdjm:
IMO the extra text is superfluous.

Please use https://bugzilla.mindrot.org/ instead of github PRs in future. We only use this site as a mirror and seldom check the PR queue.

PR #7, opened 2014-10-14
clarify how IgnoreUserKnownHosts works

calestyo:
Based on the previous documentation of the IgnoreUserKnownHosts directive, the
average user could easily think that the default value “no” is the more secure
choice (in the sense of “do not even check in ~/.ssh/known_hosts”).

• Clarify in sshd_config(5), that a value of “yes” in the IgnoreUserKnownHosts
  directive, makes sshd(8) only trust the global known hosts list (/etc/ssh/
  ssh_known_hosts).

Signed-off-by: Christoph Anton Mitterer mail@christoph.anton.mitterer.name

### DIFF
--- a/sshd_config.5
+++ b/sshd_config.5
@@ -627,7 +627,9 @@ should ignore the user's
 during
 .Cm RhostsRSAAuthentication
 or
-.Cm HostbasedAuthentication .
+.Cm HostbasedAuthentication
+and instead only trust the systemwide
+.Pa /etc/ssh/ssh_known_hosts .
 The default is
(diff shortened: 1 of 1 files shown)


calestyo:
Forwarded to https://lists.mindrot.org/pipermail/openssh-unix-dev/2015-February/033468.html

djmdjm:
committed as 62562ceae61e4f7cf

Please use https://bugzilla.mindrot.org/ instead of github PRs. We only use this site as a mirror and seldom check the PR queue.

PR #11, opened 2014-11-11
Options cleanup

calestyo:


calestyo:
Forwarded to https://lists.mindrot.org/pipermail/openssh-unix-dev/2015-February/033470.html

daztucker:
As it says in the README we don't normally use github pull requests.  Anyway applied upstream, thanks.

PR #12, opened 2014-11-19
Update contrib/redhat/openssh.spec file

nkadel:
Updates are in the %changelog, and include:
- Add '--mandir' and '--with-mantype' for RHEL 5 compatibility
- Add 'dist' option to 'ver' so package names reflect OS at build time
- Always include x11-ssh-askpass tarball in SRPM
- Add openssh-x11-aspass BuildRequires for libXT-devel, imake, gtk2-devel
- Discard 'K5DIR' reporting, not usable inside 'mock' for RHEL 5 compatibility
- Discard obsolete '--with-rsh' configure option
- Update openssl-devel dependency to 0.9.8f, as found in autoconf

### DIFF
--- a/contrib/redhat/openssh.spec
+++ b/contrib/redhat/openssh.spec
@@ -193,11 +193,6 @@ environment.
 CFLAGS="$RPM_OPT_FLAGS -Os"; export CFLAGS
 %endif
 
-%if %{kerberos5}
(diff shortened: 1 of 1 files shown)


daztucker:
As mentioned in README we don't normally use github pull requests.  I've cherrypicked these changes and updated with what's changed since then.  Thanks.

PR #13, opened 2014-12-09
Make openssh compile with musl

doughdemon:
I hope this is the right place to send this.


daztucker:
(I'm looking at the sys/param.h one too, but it also won't be a merge it'll be an upstream commit)

PR #15, opened 2014-12-15
unknown hosts: display a list of accepted hosts using the same key

saaros:
When connecting to a host for which there's no known hostkey, check if the
selected key has been accepted for other hostnames.  This is useful when
connecting to a host with a dymamic IP address or multiple names.

An example where the local laptop is using DHCP and has 192.168.3.2 as its
current IP address:

```
% ssh 192.168.3.2
The authenticity of host '192.168.3.2 (192.168.3.2)' can't be established.
ECDSA key fingerprint is 9d:14:4f:ef:0c:0b:b4:1c:62:b0:89:af:e2:43:5c:4b.
You have previously accepted this key for the following hostnames:
        192.168.1.4
        localhost
Are you sure you want to continue connecting (yes/no)?
```

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -417,7 +417,7 @@ check_key_in_hostfiles(struct passwd *pw, Key *key, const char *host,
 	const struct hostkey_entry *found;
 
 	hostkeys = init_hostkeys();
-	load_hostkeys(hostkeys, host, sysfile);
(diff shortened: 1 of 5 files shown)


djmdjm:
OpenSSH 8.5 included a similar feature, the commit that added it was d5d05cdb3d

PR #16, opened 2015-01-02
Obfuscated OpenSSH rebased for v6.7p1

BenBE:
This is a rebase of https://github.com/brl/obfuscated-openssh for OpenSSH based on v6.7p1 It would be nice to see these changes as an additional feature set in master (off be default, but available to those who want them).

The rebase and porting was based on a request by https://twitter.com/a_z_e_t/status/544631114351280129

I know the rebase is not perfect and will probably need some cleanup. Especially with the way it operates on the buffers. Additionally a review of the porting will be necessary by someone more experienced with the OpenSSH source base. There is a tag in my repository which has the patches by brl ported onto the alleged tag, thus comparing and verifying the rebase should be much easier.

I'd appreciate if this feature was added to upstream.

TIA,
BenBE.


djmdjm:
We're not interested in merging this feature. It can be accomplished through ProxyCommand on the client and a small inetd-style wrapper on the server if necessary.

PR #20, opened 2015-05-05
bcrypt_pbkdf: Fix out-of-bounds read

cryptomilk:
We are overrunning the array 'out' of 32 bytes at byte offset 1022 using
index i.


daztucker:
We've looked at this again and believe it's a false positive.  There was a bug that might have caused such a report but that was fixed in https://github.com/openssh/openssh-portable/commit/01b63498801053f131a0740eb9d13faf35d636c8 but that is prior to the version that you have in the pull request.  Is it possible that your report is based on that prior version?

If you could give an example of a set of parameters that triggers it that would be very helpful.

Thanks.

PR #22, opened 2015-06-08
Expose authentication information to PAM

gdemarcsek:
Implements enhancement described in: https://bugzilla.mindrot.org/show_bug.cgi?id=2408. Exposes information about the previously successful authentication methods to PAM.


djmdjm:
Similar feature discussed and committed via https://bugzilla.mindrot.org/show_bug.cgi?id=2408

PR #23, opened 2015-06-11
Increase robustness of redhat/openssh.spec

CarstenGrohmann:
* remove configure --with-rsh, because this option isn't supported anymore
* replace last occurrence of BuildPreReq by BuildRequires
* update grep statement to query the krb5 include directory

Signed-off-by: Carsten Grohmann <mail@grohmann-online.de>


daztucker:
The remaining BuildPreReq is behind a flag for building on really old redhat systems.  The other two changes look ok.

daztucker:
on second thought if the system can't handle BuildRequires the spec file isn't going to work anyway, so sure.

daztucker:
Applied upstream, thanks.

PR #25, opened 2015-07-17
Allow HostKeyAlias to match a host certificate principal if HostName does not

charles-dyfis-net:
See mailing list discussion starting at http://lists.mindrot.org/pipermail/openssh-unix-dev/2015-February/033443.html (and, presently, terminating at http://lists.mindrot.org/pipermail/openssh-unix-dev/2015-February/033448.html).


djmdjm:
Please use https://bugzilla.mindrot.org/show_bug.cgi?id=2359 for this

PR #33, opened 2015-11-10
Add debugging information to ga_match() to show each attempted match

PaulWay:
Add debugging information to ga_match() to show each attempted group and pattern match.

### DIFF
--- a/groupaccess.c
+++ b/groupaccess.c
@@ -83,9 +83,12 @@ ga_match(char * const *groups, int n)
 	int i, j;
 
 	for (i = 0; i < ngroups; i++)
-		for (j = 0; j < n; j++)
+		for (j = 0; j < n; j++) {
+			debug3("Does groups_byname[%d] '%s' match groups[%d] '%s'",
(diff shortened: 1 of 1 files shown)


Jakuje:
Hi. Thank you for taking time to create patch and pull request. But please, note the guidelines of contribution to openssh project, which is in short-version available in [README](https://github.com/openssh/openssh-portable/blob/master/README). OpenSSH tracks their bugs and feature requests in their own [bugzilla](http://bugzilla.mindrot.org/) and accepts discussion usually using [Mailing list](https://lists.mindrot.org/mailman/listinfo/openssh-unix-dev). This github repository is just a mirror and pull requests are [not accepted](https://github.com/openssh/openssh-portable/pull/3#issuecomment-76987598) at the moment.

PaulWay:
Then do me a favour and copy this over into that bugzilla so I don't have to create another login :-)

djmdjm:
IMO this would be far too noisy and there exist plenty of ways to interrogate group membership.

PR #35, opened 2016-01-05
to use tun for ssh-vpn on DilOS

ikozhukhov:
i have DilOS (http://www.dilos.org)
it is illumos (http://illumos.org) based platform and i have made changes to use ssh-vpn on DilOS.
please review and integrate changes.
best regards,
-Igor

### DIFF
--- a/auth.h
+++ b/auth.h
@@ -81,6 +81,8 @@ struct Authctxt {
 
 	struct sshkey	**prev_userkeys;
 	u_int		 nprev_userkeys;
+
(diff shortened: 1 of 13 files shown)


djmdjm:
FYI we do not use github pull requests, as this repository is just a mirror of our main one.

This feature is also pretty intrusive, please file a bug at https://bugzilla.mindrot.org/ to discuss it further. At present the patch is a too intrusive to commit.

PR #36, opened 2016-01-07
Fixed comment about const qualifier on certain systems regarding PAM

mschwager:
Normally I wouldn't have given the comment a second thought, but it led me down a rabbit hole when compiling on AIX :)

If we look at `configure.ac` we see that `PAM_SUN_CODEBASE` is only defined for HP-UX 11, Solaris, and SunOS 4, not AIX.

### DIFF
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -67,7 +67,7 @@
 
 /* OpenGroup RFC86.0 and XSSO specify no "const" on arguments */
 #ifdef PAM_SUN_CODEBASE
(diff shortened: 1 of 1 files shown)


Jakuje:
Please, as mentioned many times before, openssh team is using github only as a mirror and all the proposed patches are handled using https://bugzilla.mindrot.org/ or at least drop a message to mailing list.

daztucker:
Applied upstream (as mentioned, we don't use github for pull requests, it's only a mirror).

PR #39, opened 2016-01-14
wolfSSL support

kaleb-himes:
Alternate open source crypto solution


djmdjm:
This patchset is pretty intrusive and doesn't seem to be that cleanly implemented. We're not interested in supporting more crypto libraries unless there is a really good justification.

PR #41, opened 2016-05-28
Move MAX_DISPLAYS to a configuration option, fix a quirk or bug in MAX_DISPLAYS

agreene5:
- Fixed what appeared to be a bug or a quirk in how MAX_DISPLAYS is handled

When MAX_DISPLAYS is 1000 and X11DisplayOffset is 10, only 990 displays
were actually attempted because the X11DisplayOffset wasn't taken
into consideration. I didn't notice any clear documentation on how this
is supposed to work since it's an internal variable.

- Moved MAX_DISPLAYS to a default value and added MaxDisplays

Added a new MaxDisplays configuration option that override MAX_DISPLAYS

This is very useful in environments where OpenSSH is used as a
gateway for a large number of users, as in the environment where
I manage OpenSSH. The current hard limit of 1000 is not enough
in some cases, and is not configurable.

### DIFF
--- a/channels.c
+++ b/channels.c
@@ -148,9 +148,6 @@ static int all_opens_permitted = 0;
 
 /* -- X11 forwarding */
 
-/* Maximum number of fake X11 displays to try. */
-#define MAX_DISPLAYS  1000
-
 /* Saved X11 local (client) display. */
 static char *x11_saved_display = NULL;
 
@@ -3890,7 +3887,8 @@ channel_send_window_changes(void)
  */
 int
 x11_create_display_inet(int x11_display_offset, int x11_use_localhost,
-    int single_connection, u_int *display_numberp, int **chanids)
+	int max_displays, int single_connection, u_int *display_numberp, 
+	int **chanids)
 {
 	Channel *nc = NULL;
 	int display_number, sock;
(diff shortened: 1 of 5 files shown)


agreene5:
I realize this might not be useful enough to be accepted upstream but I ask for your consideration. It's quite painful in my environment having to hand edit a hardcoded value when deploying a new version.

Jakuje:
This seems reasonable feature, but there are few comments to the code:
 * The option should certainly have some documentation, basically item in manual page for `sshd_config` should be enough. Also note about the option in manual page for `sshd` is a good practice. If we will accept this in `Match` blocks, then it should be mentioned also there.
 * The constant `MAX_DISPLAYS` should probably be `DEFAULT_MAX_DISPLAYS` for consistency.
 * I would leave option parsing on `goto parse_int;` as in different cases. Using `a2port` seems confusing.
 * There is used `M_CP_INTOPT(max_displays);`, but the option is not marked as available in `Match` blocks (`SSHCFG_GLOBAL` flag). This needs also to clear up.
 * I would consider prefixing the option with `X11_`, as the other X11 related options are
 * There is also some whitespace noise in the changed lines, that would be nice to clean up.

Please, consider the notes and address the issues above. OpenSSH team does not accept contributions on github, so when we will have patch ready, it would be good to fill upstream bug on https://bugzilla.mindrot.org/

agreene5:
Thanks for the notes, very helpful. I will make the suggested changes and additions and then resubmit to the mindrot mailing list for the OpenSSH team to review.

agreene5:
Thanks for the input Jakuje, I've fixed the issues you mentioned up and added documentation. I will attach the patch to https://bugzilla.mindrot.org/show_bug.cgi?id=2580

PR #42, opened 2016-05-30
- fixed login in pkcs11_open_session

danielkucera:
Hi,
there is a bug in pin detection: if no pin is supplied to function, the exit condition is skipped.
Also if no pin is supplied, login is skipped even when card requires login.

### DIFF
--- a/ssh-pkcs11.c
+++ b/ssh-pkcs11.c
@@ -366,19 +366,16 @@ pkcs11_open_session(struct pkcs11_provider *p, CK_ULONG slotidx, char *pin)
 
 	f = p->function_list;
 	login_required = p->slotinfo[slotidx].token.flags & CKF_LOGIN_REQUIRED;
-	if (pin && login_required && !strlen(pin)) {
-		error("pin required");
-		return (-1);
-	}
+
 	if ((rv = f->C_OpenSession(p->slotlist[slotidx], CKF_RW_SESSION|
 	    CKF_SERIAL_SESSION, NULL, NULL, &session))
 	    != CKR_OK) {
 		error("C_OpenSession failed: %lu", rv);
 		return (-1);
 	}
-	if (login_required && pin) {
+	if (login_required) {
 		rv = f->C_Login(session, CKU_USER,
-		    (u_char *)pin, strlen(pin));
+		    (u_char *)pin, (pin != NULL) ? strlen(pin) : 0);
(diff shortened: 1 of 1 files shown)


etatus:
I agree. 

ssh doesn't allow entering a PIN as an argument, so it make no sense a condition that requires the PIN was entered when it should be asked to open the session. The proposed changes are mandatory to support HSMs that requires a login to read public keys.

Jakuje:
No, this is already discussed in the upstream bugzilla:

https://bugzilla.mindrot.org/show_bug.cgi?id=2430

This change would bother you to login, even if you would not need to use the token at all, not only in ssh-keygen, but also in the ssh process itself.

Most of the tokens these days allow you to list public objects before login. There is no way how to detect whether the loaded token is one of them or not, but the closest approximation is to try load keys without login and if there would be nothing usable, try to login and retry.

etatus:
Ok, you are right. The ideal behaviour is what you say: try to load keys without login and if there is no keys try to login by asking the PIN.

If you are thinking in local tokens/smarcards connected via USB, it could be acceptable to read public objects over and over without login, but if you connect to a network device through PKCS#11, for instance LAN HSM or even Cloud HSM (Amazon Cloud HSM), you should always login the HSM to achieve a better access control (you might be charged for each access to the device).

Do you know if there is any solution or workaround to this? I tried to put a public key in ~/.ssh/id_rsa.pub, but it seems ssh doesn't look for the corresponding private key in the PCKS11 provider if the public key was found in the filesystem.

Jakuje:
No, there is no workaround at this moment for ssh nor ssh-keygen. But if you load the pkcs11 module to ssh-agent, it is logging in in the first place, because it is already provided with a pin.

etatus:
I've already tried the ssh-agent and yes it works (partially). At first, it asks for the PIN to retrieve and convert the public keys to openssh. Then, you can make ssh connections and as ssh-agent provides the public keys and send a request to sign the challenge with the corresponding private key through PKCS#11 (it is already logged in) the connection to the server succeeds.

The problem is that after a while, when the PKCS#11 session expires, ssh-agent cannot request the signature because it is not logged in and it should ask for the PIN again instead returning an error. If you are lucky and your PKCS#11 has a KeepAlive=true option, maybe you can solve this with the corresponding risks of keeping the access to the device open forever.

Another problem of ssh-agent is that you have to keep it running all the time. If you want to make a ssh connection from a laptop, the ssh-agent configuration dissapears when the computer is turned off.

PR #46, opened 2016-06-22
Set BROKEN_STRNVIS on macOS

pom:
macOS 10.12 ships with its own copy of strnvis().


daztucker:
As the README says, we don't use github for pull requests so unfortunately this missed the 7.3p1 release.
Instead of continuing to whack this particular mole I added a test for the bad behaviour:
https://anongit.mindrot.org/openssh.git/commit/?id=20e5e8ba9c5d868d897896190542213a60fffbd2

Please file a bug at bugzilla.mindrot.org if this does not fix the issue on OS X.

PR #47, opened 2016-07-01
Expose authentication information to PAM [Bug 2408]

Feandil:
This PR is a code proposal to resolve [bug 2408](https://bugzilla.mindrot.org/show_bug.cgi?id=2408)
It contains the same code as the [patch attached to the bug on 2016/07/01](https://bugzilla.mindrot.org/attachment.cgi?id=2846), but as atomic patches, for a simpler review.


djmdjm:
Something similar has been committed

PR #48, opened 2016-09-18
Make it build using OpenSSL 1.1.0

kroeckx:


sshkey.c
```diff
@@ -2760,16 +2945,49 @@ sshkey_private_deserialize(struct sshbuf *buf, struct sshkey **kp)
...
+				goto out;
+			}
+			if ((r = rsa_generate_additional_parameters(k->rsa, iqmp)) != 0) {
+				BN_free(iqmp);
+				goto out;
+			}
+		}
 		break;
 	case KEY_RSA_CERT:
+#if 0
```
Jakuje: This is intentionally left out or like todo for future readers?

Jakuje: Thank you. I noticed your message on the mailing list, but forgot to check it properly before commenting. Adding it here clears that up. Thank you.

sshkey.c
```diff
@@ -2401,12 +2498,34 @@ sshkey_certify(struct sshkey *k, struct sshkey *ca, const char *alg)
...
+			BIGNUM *p, *q, *g, *pub_key;
+
+			p = BN_new();
+			q = BN_new();
+			g = BN_new();
+			pub_key = BN_new();
+
+			if (p == NULL || q == NULL || g == NULL ||
+			    pub_key == NULL ||
+			    (ret = sshbuf_put_bignum2(cert, p)) != 0 ||
```
Jakuje: This is certainly wrong. In this case, you should not allocate&put&set. It should be get&put. The same for below RSA case.

kroeckx: You're right, new commit added.

Jakuje:
I noticed you didn't port the SSH1 code to OpenSSL 1.1.0 -- we don't care about this protocol, but in clients it will be among us for some time. I added a the missing (only client side) bits in the answer on the mailing list [1].

Also I noticed you created a new compat filename, even though, there is already `openbsd-compat/openssl-compat.{c,h}`, which should be able to accommodate these changes, though the final decisions is probably up to the upstream developers which place to choose.

Also the missing part under `#if 0`, which is using zero BIGNUMs for missing parts.

[1] http://lists.mindrot.org/pipermail/openssh-unix-dev/2016-November/035454.html

noloader:
Please expedite this patch. Lack of OpenSSL 1.1.0 support is harming users. From [OpenSSL 1.1.0 integrtion](https://groups.google.com/forum/#!topic/comp.security.ssh/hPsJSM45p4A) on the user mailing list:

> The OpenSSL 1.0.2 downgrade means users lose a number of features from OpenSSL, like Poly1305, curve25519 and the project's improved security posture and testing procedures. (Other improvements are available at https://www.openssl.org/news/openssl-1.1.0-notes.html).

cookiengineer:
@Jakuje What is the work that is left to be done in order to get this PR merged? Following this thread it's a bit unclear for me as a third-party guy.

Jakuje:
@cookiengineer The most of the discussion about this topic was held on the mailing list thread linked from the previous comment. The bottom-line can be found here:

http://lists.mindrot.org/pipermail/openssh-unix-dev/2016-November/035497.html

There is no work need to be done on this PR (ok, there are few problems in this PR resolved by the [patch carried in Fedora](http://lists.mindrot.org/pipermail/openssh-unix-dev/2016-November/035454.html)) or upstream, but some compatibility between OpenSSL releases was requested which is very unclear if it can realistically happen.

cipher.c
```diff
@@ -372,7 +372,7 @@ cipher_init(struct sshcipher_ctx **ccp, const struct sshcipher *cipher,
 		ret = SSH_ERR_ALLOC_FAIL;
 		goto out;
 	}
-	if (EVP_CipherInit(cc->evp, type, NULL, (u_char *)iv,
+	if (EVP_CipherInit(cc->evp, type, (u_char *)key, (u_char *)iv,
```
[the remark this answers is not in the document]
kroeckx: I can't even remember why I did change (it's been more than a year). But I guess the same applies to the IV?

libcrypto-compat.c
```diff
@@ -0,0 +1,410 @@
...
+}
+
+unsigned char *EVP_CIPHER_CTX_iv_noconst(EVP_CIPHER_CTX *ctx)
+{
+    return ctx->iv;
+}
+
+EVP_MD_CTX *EVP_MD_CTX_new(void)
+{
+    return OPENSSL_zalloc(sizeof(EVP_MD_CTX));
```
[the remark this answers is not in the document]
kroeckx: And I can just get rid of the OPENSSL_zalloc() function in that case. For some reason I missed EVP_MD_CTX_create() / EVP_MD_CTX_destroy() at the time I wrote this.

cipher.c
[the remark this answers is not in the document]
kroeckx: Oh, I did make a comment about this:

    I've replaced the 2 EVP_CipherInit() calls in cipher_init() with 1. OpenSSL now clears
    everything when you call EVP_CipherInit() again, so what was passed in the first but not
    in the second call, and what the function calls between them did, was lost.

I wonder if this is actually a problem in OpenSSL.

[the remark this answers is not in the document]
davidben: I think the IV can be passed in at either time? OpenSSL doesn't have a mechanism for ciphers with variable-length IVs as far as I know.

If the `EVP_CIPHER` has a fixed length key [and you already took care that what you're passing as the right size, since the `EVP_CipherInit` API isn't bounds-checked], you don't need to do `EVP_CIPHER_CTX_set_key_length` first. If you have a variable-length key cipher [and you're not using whatever the default key size is], you do.

Granted, it seems the only such ciphers, barring `ENGINE` mischief, are Blowfish, CAST, RC2, RC4, and RC5 so this pattern is only of so much interest... :-) Though the existing `EVP_CIPHER_CTX_set_key_length` logic did also act as a bounds check I think.
https://git.openssl.org/gitweb/?p=openssl.git&a=search&h=HEAD&st=grep&s=EVP_CIPH_VARIABLE_LENGTH
https://git.openssl.org/gitweb/?p=openssl.git&a=search&h=HEAD&st=grep&s=EVP_CIPH_CUSTOM_KEY_LENGTH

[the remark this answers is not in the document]
davidben: Aha! I think what might actually be going on here is that openssh should have been using `EVP_CipherInit_ex`, not `EVP_CipherInit`. `EVP_CipherInit` was for folks who didn't call `EVP_CIPHER_CTX_init` (which `EVP_CIPHER_CTX_new` did implicitly).

But in 1.0.2, there was this weird quirk where `EVP_CipherInit` would skip `EVP_CIPHER_CTX_init` if `cipher` was NULL:
https://git.openssl.org/gitweb/?p=openssl.git;a=blob;f=crypto/evp/evp_enc.c;h=be577bac767fe9cf2c163be6967b20d1f6cc9014;hb=refs/heads/OpenSSL_1_0_2-stable#l94

OpenSSL 1.1.0 lost this behavior:
https://git.openssl.org/gitweb/?p=openssl.git;a=blob;f=crypto/evp/evp_enc.c;h=d21a0c7e54d54a5a77a2d5a6593e4851cf2f5291;hb=HEAD#l50

So I believe you want to revert this and merely switch those `EVP_CipherInit` calls to `EVP_CipherInit_ex`, which should work fine in 1.0.2 and 1.1.0. (Unless 1.1.0 additionally changed behavior here in ways I'm missing.) You may also wish to consider at minimum marking `EVP_CipherInit` as deprecated and documenting this behavior change, and possibly restoring the 1.0.2 behavior to avoid breaking folks.

kroeckx:
I've fixed the merge conflicts, rebased it against current master, and fixed various issues.

ssh-pkcs11.c
```diff
@@ -525,17 +540,18 @@ pkcs11_fetch_keys_filter(struct pkcs11_provider *p, CK_ULONG slotidx,
...
-			} else if ((rsa = RSAPublicKey_dup(evp->pkey.rsa))
+			} else if ((rsa = RSAPublicKey_dup(EVP_PKEY_get0_RSA(evp)))
 			    == NULL) {
 				error("RSAPublicKey_dup");
 			}
 			if (x509)
 				X509_free(x509);
 		}
-		if (rsa && rsa->n && rsa->e &&
+		RSA_get0_key(rsa, &n, &e, NULL);
```
Jakuje: This needs to be rewritten to something like the following, otherwise it will segfault for non-RSA keys (ECC on yubikey):
```
if (rsa && RSA_get0_key(rsa, &n, &e, NULL) && n && e &&
```
I just noticed it with the version I run on Fedora.

Jakuje: nope ... the `RSA_get0_key()` is void function so it will not work this way. But clearly we can call this function only if we are sure the `rsa` is not NULL.

noloader: > the RSA_get0_key() is void function so it will not work this way. But clearly we can call this function only if we are sure the rsa is not NULL

Just my 2¢, but if `RSA_get0_key` can't handle a NULL parameter, then it sounds like an OpenSSL bug. The [`RSA_get0_key` document](https://www.openssl.org/docs/man1.1.0/crypto/RSA_get0_key.html) does not state the parameter *cannot* be NULL; and it also states other parameters can be NULL.

rcombs:
Anything still blocking here, other than the `RSA_get0_key` issue?

kroeckx:
I've fixed the RSA_get0_key() call.

rcombs:
Any remaining blockers here? It seems like the primary issue that's been raised is compatibility with both the 1.0 and 1.1 APIs, which is handled here. It appears that all review comments have been addressed. Is this just pending additional review?

noloader:
From the TLS Working Group mailing list at [3rd WGLC: draft-ietf-tls-tls13]():

```
All,

This is the 3rd working group last call (WGLC) announcement for draft-ietf-tls-t
ls13; it will run through January 26th.  This time the WGLC is for version -23 (
https://datatracker.ietf.org/doc/draft-ietf-tls-tls13/).  This WGLC is a targete
d WGLC because it only address changes introduced since the 2nd WGLC on version
-21, i.e., changes introduced in versions -22 and -23.  Note that the editor has
 kindly included a change log in s1.2 and the datatracker can also produce diffs
 (https://www.ietf.org/rfcdiff?url1=draft-ietf-tls-tls13-21&url2=draft-ietf-tls-
tls13-23).  In general, we are considering all other material to have WG consens
us, so only critical issues should be raised about that material at this time.

Cheers,

spt
```

laomaiweng:
On my Gentoo box (using OpenSSL 1.1.0g configured with _--api=1.0.0_ and no `OPENSSL_API_COMPAT` #defined at build time), this PR is still missing some stuff:
* here is a first patch (against OpenSSH 7.6p1) to fix the remaining OpenSSL API changes that prevent building on my machine: https://592578.bugs.gentoo.org/attachment.cgi?id=518526
* here is a second patch (still against OpenSSH 7.6p1) to rename `pthread_{create,cancel,exit,join}()` in _auth-pam.c_, since OpenSSL 1.1.0 with threads support on Linux unconditionally #includes <pthread.h>, which collides with these functions: https://bugs.gentoo.org/attachment.cgi?id=518528

Hope that helps!

okias:
Will be OpenSSH 7.7 has issue with OpenSSL 1.1.x files?

okias:
Will be this issue fixed in OpenSSH 7.7?

I migrated to OpenSSL 1.1.1, because of TLS 1.3 support, which is really nice to have!

Jakuje:
No. The latest statement on the mailing list said that OpenSSH targets LibreSSL (to optionally bundle) rather than OpenSSL anymore.

https://lists.mindrot.org/pipermail/openssh-unix-dev/2017-October/036346.html

kroeckx:
And libressl now actually supports most (all?) of the new
functions, so maybe most of this patch can be applied to the
OpenBSD version instead of the portable version.

okias:
well, I really like idea of LibreSSL, but sadly TLS 1.3 support is at this moment available only with OpenSSL 1.1.1.

kanavin:
I was a bit disappointed that the new openssh release does nothing about the whole openssl1.0/openssl1.1/libressl situation. I am now moving the Yocto project to openssl 1.1, and openssh is the sole remaining holdout that requires 1.0 (see [1]). Perhaps the project should just officially announce that going forward, libressl is the only supported dependency, and anyone who wants opessh, has to package libressl as well.

[1] http://git.yoctoproject.org/cgit/cgit.cgi/poky-contrib/log/?h=akanavin/openssl-1.1.1

berolinux:
FWIW there is a working patch that is rebased to the current release at
http://www.linuxfromscratch.org/patches/downloads/openssh/openssh-7.7p1-openssl-1.1.0-1.patch

okias:
kinda sad, there has to be support for new releases of security library managed by external patches, can be this patch included, when someone already did majority of work?

shoujii:
Is there any patch to make it work with 1.1.1 pre5?
And yes, it's sad trhat there is no merge...

okias:
Well, **OpenSSH devs**:
I'd like to ask you, if you could give some statement. At least simple words like
- **"we accept patch with 1.1.1 support"** or
- **"we don't want openssl support in, keep it outside of openssh"**

For second scenario would be good, if there will be **unified patch** maintained on **a top of** development of actual **OpenSSH**.

Thank you

kanavin:
I would not want to use any third party patch. It's far too complex and invasive, and there is no guarantee that it will be maintained going forward. For Yocto project I went ahead and packaged libressl (for the sole purpose of supporting openssh), but I am not at all happy with how openssh devs do nothing to resolve the situation, and push it over to distributors. 
http://git.yoctoproject.org/cgit/cgit.cgi/poky-contrib/log/?h=akanavin/openssl-1.1.1

ericcurtin:
@okias there are clues in the repo:

https://github.com/openssh/openssh-portable/blob/master/INSTALL

> libcrypto (LibreSSL or OpenSSL >= 1.0.1 < 1.1.0)

> Note that because of API changes,
> OpenSSL 1.1.x is not currently supported.

It feels like openssh don't want to support openssl 1.1.x but their hand may be forced if more platforms don't start moving off openssl.

But that's just a guess on my part!

pprindeville:
Is there an obstacle with providing conditional compilation so it can be built against either LibreSSL *or* OpenSSL 1.0.x *or* OpenSSL 1.1.x?

pprindeville:
@kroeckx Kurt, can you please rebase so the conflicts go away?

Jakuje:
@pprindeville no, it works for all of them to my best understanding. If you are interested in rebased version, I keep around the changes in my fork so I would be able to build OpenSSH on my machine and it is rebased on master from time to time:

https://github.com/Jakuje/openssh-portable/commit/44dc0e6

berolinux:
We apply a ported variant of the patch in our distro packaging:

https://raw.githubusercontent.com/OpenMandrivaAssociation/openssh/master/openssh-7.7p1-openssl-1.1.0-1.patch

Works perfectly with 7.7p1

AnimusPEXUS:
as far as I can recall, openssh provides only executables. no shared objects. libssh and libssh2 build with openssl-1.1.x not bad. probably, now it is good opportunity for somebody to start concurrent project to openssh.. ;)

xnox:
https://www.openssl.org/blog/blog/2018/09/11/release111/

> Since 1.1.1 is our new LTS release we are strongly advising all users to upgrade as soon as possible.

> Our previous LTS release (OpenSSL 1.0.2) will continue to receive full support until the end of this year. After that it will receive security fixes only. It will stop receiving all support at the end of 2019. Users of that release are strongly advised to upgrade to OpenSSL 1.1.1.

emaste:
OpenSSL 1.1.x compatibility started arriving about a week ago: https://github.com/openssh/openssh-portable/commit/482d23bcacdd3664f21cc82a5135f66fc598275f

I cherry-picked the 1.1.x-compat commits on top of V_7_8_P1 (resolving trivial conflicts) and pushed the result here: https://github.com/emaste/openssh-portable/tree/v7.8p1-OpenSSL1.1

daztucker:
OpenSSL 1.1.x is now supported. Thanks.

PR #50, opened 2016-09-20
kex_send_ext_info: return error properly

Safari77:
Why it returns 0 on those errors?
Return r instead.


djmdjm:
Thanks. This was merged via upstream in 1622649b7a829fc8dc313042a43a974f0f3e8a99

PR #51, opened 2016-09-30
Linux getrandom support

Safari77:
also moved OpenSSL `RAND_bytes` call inside `getrnd`


djmdjm:
I don't think we want this. libcrypto will break horribly in other ways if its PRNG isn't seeded (example: ECDSA nonces), so this doesn't buy us much.

Maybe it might be worth doing getrandom() if libc has the syscall wrapper because it's so simple, but I don't want to go to the trouble of invoking the syscall directly.

PR #53, opened 2016-10-31
Certificateidentityhandling

aeijdenberg:
Creating pull request to link to in mailing list discussion.


djmdjm:
Please use https://bugzilla.mindrot.org/ instead of github PRs in future. We only use this site as a mirror and seldom check the PR queue (this the inactivity).

AFAIK there have been a few fixes to bare cert keys since this was reported.

PR #54, opened 2016-11-07
Bug 2377 - ssh-keygen: Add ssh-agent support for key signing

volans-:
Communicate with the ssh-agent to sign certificates if agent is present
and is loaded with the specified certificate authority.
In order for the agent to sign, the certificate authority public key
must be alongside the private key with the same file name but a .pub
extension.

Closes: https://bugzilla.mindrot.org/show_bug.cgi?id=2377

Patch developed against 7.1p by Meghana Bhat.
Patch adapted to master branch at 010359b (post 7.3p1) by @volans-.

### DIFF
--- a/authfd.c
+++ b/authfd.c
@@ -121,7 +121,7 @@ ssh_get_authentication_socket(int *fdp)
 }
 
 /* Communicate with agent: send request and read reply */
-static int
+int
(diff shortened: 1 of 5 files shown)


djmdjm:
Support for signing via the agent was added in a98339edbc1fc.

Please use https://bugzilla.mindrot.org/ in future, this is just a mirror of our main repository.

PR #56, opened 2016-11-22
Constify maybe_add_key_to_agent@sshconnect.h + dependencies

Feandil:
maybe_add_key_to_agent and its dependencies only read their variable arguments: they can be _constifyed_ to reflect it.


djmdjm:
I've committed most of this (except the SSHv1 bits) as d9e5cf07

Please use https://bugzilla.mindrot.org/ instead of github PRs in future. We only use this site as a mirror and seldom check the PR queue.

PR #57, opened 2016-11-23
sshconnect2.c: only re-order keys when resetting keys (instead of destroying the list and failing to rebuild it)

Feandil:
sshconnect2.c whipes all keys and close the agent fd between each authentication before trying to rebuild the list (but failing because building the list modifies its sources) and reconnecting to the agent.
However, during an authentication:
- Almost all access to the keys/key list do not modify them
- The only modification to them is in userauth_pubkey, which:
   * Re-order the key, increasing the `tries` count (up to 2)
   * Set the `isprivate` flag
   * All other functions called do not modify the keys

This PR proposes a solution by:
- Modifying userauth_pubkey so that it cleans up `isprivate` when it cleans up the key
- Adding a `pubkey_reset` function that:
   * Reverts the re-ordering (this could  even be skipped as the order does not seem to carry much importance...)
   * Resets the `tries` to 0

This PR also include some function argument constification, to make it simpler to see that these function do not modify their arguments

Should fix https://bugzilla.mindrot.org/show_bug.cgi?id=2642

### DIFF
--- a/authfd.c
+++ b/authfd.c
@@ -496,7 +496,7 @@ ssh_agent_sign(int sock, struct sshkey *key,
 
 #ifdef WITH_SSH1
 static int
-ssh_encode_identity_rsa1(struct sshbuf *b, RSA *key, const char *comment)
+ssh_encode_identity_rsa1(struct sshbuf *b, const RSA *key, const char *comment)
 {
 	int r;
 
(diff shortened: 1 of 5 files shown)


djmdjm:
AFAIK this was committed in b9844a45c7f016

PR #59, opened 2016-12-14
Fix X11 forwarding on macOS

jschlyter:
X11 forwarding does not work on macOS, as $DISPLAY is a path to socket set by launchd. The patch is based on work by Apple and simplified by me.

This bug has a long history, see ref: https://bugzilla.mindrot.org/show_bug.cgi?id=2341 for more information.


daztucker:
Patch was applied in https://bugzilla.mindrot.org/show_bug.cgi?id=2341 and was included in the 7.5p1 release.  Thanks.

PR #60, opened 2017-01-10
Add SyslogFacility setting to 'ssh' client program.

erahn:
Add SyslogFacility setting to 'ssh' client program. Also update ssh_config.5 man page to mention the new command.

I tested this on my local machine to make sure that it worked and everything seemed to work correctly.

I mentioned this + that I would post it on github on the openssh-unix-dev@mindrot.org mailing list and Darren Tucker said it seemed like a reasonable feature.

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -152,7 +152,7 @@ typedef enum {
 	oGlobalKnownHostsFile, oUserKnownHostsFile, oConnectionAttempts,
 	oBatchMode, oCheckHostIP, oStrictHostKeyChecking, oCompression,
 	oCompressionLevel, oTCPKeepAlive, oNumberOfPasswordPrompts,
-	oUsePrivilegedPort, oLogLevel, oCiphers, oProtocol, oMacs,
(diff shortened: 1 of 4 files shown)


Jakuje:
Please, fill a new bug in https://bugzilla.mindrot.org/

As already said, the github is only a mirror and the patches are tracked in the bugzilla. From the [README](https://github.com/openssh/openssh-portable/blob/master/README):

> Please read https://www.openssh.com/report.html for bug reporting
instructions and note that we do not use Github for bug reporting or
patch/pull-request management.

readconf.c
```diff
@@ -246,6 +246,7 @@ static struct {
 	{ "keepalive", oTCPKeepAlive },				/* obsolete */
 	{ "numberofpasswordprompts", oNumberOfPasswordPrompts },
 	{ "loglevel", oLogLevel },
+        { "syslogfacility", oLogFacility },
```
Jakuje: There should be tab instead of spaces. The same in the bellow chunks.

daztucker:
This was applied in https://github.com/openssh/openssh-portable/commit/68d3a2a05.

Thanks.

PR #62, opened 2017-03-01
Update servconf.c:  sAuthorizedKeysCommand and sAuthorizedPrincipalsC…

discodance1:
…ommand to change the command path format

Currently AuthorizedPrincipalsCommand expects the command path to start with `/` which is not the case for windows as the path starts with C:/Programdata/<blah>.
similarly for AuthorizedKeysCommand.


daztucker:
right now the only supported windows platform is cygwin and you can use /cygdrive/c/Programdata/...

anyway, as the README says we don't use github for pull requests, please see https://www.openssh.com/report.html

PR #63, opened 2017-03-16
Enable OpenSSH for Windows

manojampalam:
Seeking feedback on changes in OpenSSH sources added to support Windows platforms. The changes are not complete yet (code reviews and significant testing are still pending) but since most of the main line functionality is in place, we thought we would check with experts on changes done so far and if they seem feasible for a possible integration upstream sometime in future. Please focus on changes within the root folder and in openbsd-compat.

Following is done to keep most of Unix-based OpenSSH source code intact and to make it work on Windows. 
  - POSIX SDK and single threaded POSIX to Win32 adapter that implements
    - network, file, pipe and terminal IO (interrupt-able blocking and non-blocking). 
    - file descriptors and select
    - Signals
    - Other misc POSIX calls applicable on Windows. Ex dirent and passwd routines among many others.
  - Unicode translation surface that handles UTF-8 <--> UTF-16 conversion between POSIX to Win32 API calls. 
    - Windows command line UTF-16 input is captured on wmain entry point, converted to UTF-8 that gets fed to OpenSSH main() routines
    - POSIX and C runtime routines that exist on Windows are overridden to support UTF-8 based I/O.
  - A handful of Unix specific calls, that are not applicable on Windows are implemented as No-OPs.
	

That said, following lead to inclusion of some Windows specific code in OpenSSH sources
  - At an architectural level, Windows cannot support Unix like privilege separation model that heavily relies on fork() and setuid() calls. Fork() does not exist on Windows and setuid() does not map with how Windows security works. Also we have concluded that running ssh-agent in user session has some security implications (applicable in Unix too but more so on Windows). Based on this reasoning, we have rearchitected OpenSSH components in a way that's secure on Windows with minimal deviation in source code layout. 
    - Sshd is implemented as a service running in low privileged mode.
    - A Windows version of ssh-agent is implemented as a service running as ROOT/SYSTEM. This apart from serving *key* requests (like Unix version does), also acts as a privileged broker for sshd. For Windows, ssh-agent (on the daemon side) also serves client authentication. 
  - Fork() calls are replaced with a custom spawn_child call. In sshd, additional state information is added to identify and differentiate logic in parent (listener) and child (connection worker) processes.
  - Client authentication - Auth on Windows ends up with a user security token. Subsystem processes are run in the context of client user. 
  - POSIX calls based on uid and gid that cannot be mapped to Windows that has a more complex identity representation.
  - Differences between Unix and Windows file/dir paths. 

The changes done so far ensured that the original code for Unix remains untouched. This is so as to not cause any regressions in Unix land and to ensure easier review of the initial set of changes.


brynet:
If you're looking for feedback from OpenSSH developers, I think you might have more luck using the official mailing lists and bug tracker.

https://www.openssh.com/report.html
https://www.openssh.com/list.html
https://bugzilla.mindrot.org/
https://marc.info/?l=openssh-unix-dev

djmdjm:
I've finally found some time to start looking at this. Some high-level comments first. Please keep in mind that I'm not very familiar with Windows programming and haven't done any for quite a few years.

I'm unclear why you re-used ssh-agent to perform what seems to be a very different role in the server. It seems to acting in a role that is analogous to the privilege separation monitor process on Unix. I can understand why you'd want a separate persistent process to handle parts of the authentication flow, but why built it in to ssh-agent? Could you bring back something more similar to OpenSSH Unix style privsep by making sshd.c:server_accept_loop() spawn two processes (monitor and unprivileged child) instead of one?

The authentication callouts to the ssh-agent process seem to get back a token pointer that is stored in authctxt->methoddata and later used in setting up the session. Given the token pointer is passed as a 32 bit value, how does this work on 64 bit systems?

Removing privilege separation and sandboxing is quite a loss IMO. I guess splitting some functions into ssh-agent provides some of the benefits of privilege separation but it still seems like a subset of the protections it provides, e.g. in Unix privilege separation the host's private keys are not directly available to the low-privilege process. This doesn't seem to be the case with the current ssh-agent arrangement (though it could be). Also, I think (remember that my Win32 experience is years rusty though) that sandboxing in Unix OpenSSH more drastically limits what the low-privilege process can do - e.g. rendering it unable to open network connections and probe kernel attack surface for local privilege escalations.

One specific comment: process_authagent_request() in authagent-request.c doesn't check opn_len is sufficient before calling memcmp()

Generally, I don't see us being able to merge this patch any time soon because there are way too many ifdefs for us to maintain upstream given our general lack of experience developing Windows applications - we'd end up accidentally breaking Windows support too often for it to be viable.

That being said, I think we could work towards minimising large parts of it. A good place to start might be in subprocess handling - there are a few cases where we need to fire off subprocesses (e.g. sftp and scp). If we made a more convenient API to do so then we could probably avoid a bunch of ifdefs. We could use something like your spawn_child API as a basis...

manojampalam:
Thanks @djmdjm for looking into this. 

Here's some back ground. In Windows, it’s a standard practice to run any Network facing service under the context of [NetworkService](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684272(v=vs.85).aspx). This is a special low privilege machine account that (at least in theory) should do no harm to the system if its compromised from a security standpoint. 
That said, this account still has some privileges (ex. one to act on client's behalf). Otherwise, Windows services rarely need to run as root. Authentication is carried over by talking to [SSPs](https://en.wikipedia.org/wiki/Security_Support_Provider_Interface) (GSSAPI provider equivalents) that are hosted by [security subsystem](https://technet.microsoft.com/en-us/library/cc961760.aspx).
The end result of client authentication is a security token that represents the client user.

When I approached privilege separation for Windows, I had to weigh in its benefits vs its complexity. It does provide a more controlled and sandboxed environment to process untrusted network data. But due to the fact that there isn't a fork() equivalent in Windows, having a side by side implementation (using a spawn_child approach) is going to make it more complex from a code management stand point. By disabling the current "privilege separation code" entirely for Windows, I felt we can keep the assumptions clean and simple while introducing and enabling Windows support.  Perhaps, we could look into complexity introduced by solving this problem for sshd initial fork (for connection worker) and reevaluate the feasibility of introducing privilege separation for Windows at a later point.

Otherwise, running the daemon as a less privileged service itself would provide us with reasonable security to start with. At this point we had to consider how to secure "host keys". For Windows, running ssh-agent as a root service solved this problem without introducing any common code changes or complexity. Ssh-agent for Windows can persist key registrations by securely storing them using [user's or system's security context](https://msdn.microsoft.com/en-us/library/ms995355.aspx). To keep host keys secure on Windows, it is recommended to register them with ssh-agent. That said, the ssh-agent model itself is still intact and users can use other agents of their choice (by setting SSH_AUTH_SOCK).

The next problem to consider is authentication. To generate client user tokens for SSH key based authentication, one needs to be root. As you indicated, If privilege separation was enabled for Windows, and sshd was running as root, this could be a simple case of delegating a privileged operation to the monitor. But considering that this was just one and the only requirement for privilege escalation, I decided to go with something Windows specific - a high privileged key authentication agent. Since ssh-agent version for Windows already runs as root, I added the key authentication capability into it. I understand the confusion this resulted in. I will separate out the "key-management" agent and "key-authentication" agent end points. From sshd standpoint, it will open separate connections to these 2 agents. "key-management" agent will continue to be controlled by SSH_AUTH_SOCK while the "key-authentication" agent is going to be internal and Windows specific.

Regarding encoding of tokens as 32-bit values, this is a Windows artifact. Windows handles (Unix fd equivalents) though a (void*) type - they are guaranteed and can only contain 32-bit values. This is a requirement stemming from having the ability to pass handles between 32-bit and 64-bit processes.

AkihiroSuda:
What's the current status?

manojampalam:
We are continuing to add E2E test coverage on Windows fork while constantly integrating upstream changes (to maintain Windows fork and openssh/master in sync).

We will soon start on refactoring efforts to address Windows specific #ifdefs.

manojampalam:
Closing this PR. Windows fork has moved on significantly since this version and is now architecturally aligned with Unix (specifically around privilege separation). Its POSIX compat library is now mostly compatible and we are following up with individual PRs to address Unix/Windows platform differences via platform abstraction.

PR #64, opened 2017-03-23
Set ldns=yes in configure if using ldns-config

fornwall:
If the configure script detects ldns-config it forgets to set ldns=yes.

I also attached this patch in a bugzilla issue at https://bugzilla.mindrot.org/show_bug.cgi?id=2697, but I'm unsure if you prefer github pull requests or not for these kind of patches.


daztucker:
We prefer bugzilla (it's listed in the README).  I have applied your patch.

Thanks!

PR #66, opened 2017-05-15
sftp-server: Add support for BSD style advisory file locking operations

kalvdans:
Only non-blocking shared or exclusive locks are supported for now. Allocate a new extension string for the flock operations.


kalvdans:
The corresponding changes in sshfs that takes advantage of this protocol extension can be viewed at: https://github.com/libfuse/sshfs/pull/55

estan:
We're waiting eagerly for this feature. It would be great if a maintainer could take a look at this PR.

Jakuje:
If you are interested in these changes in OpenSSH, you should fill the bug report on http://bugzilla.mindrot.org/ -- the github is used only as a mirror, as already said many times.

estan:
@Jakuje Aha, did not know that, sorry for the noise. @kalvdans would you mind filing a bug with the patch at Bugzilla?

kalvdans:
Thanks @Jakuje, a bug has been posted to https://bugzilla.mindrot.org/show_bug.cgi?id=2828

djmdjm:
Sorry, but we have no desire to add file locking to sftp/sftp-server. In fact, addition of file locking feature to the sftp protocol was why we never implemented versions of the protocol after 3.

djmdjm:
Also, please use https://bugzilla.mindrot.org/ instead of github PRs in future. We only use this site as a mirror and seldom check the PR queue.

PR #67, opened 2017-05-26
Fix invalid condition for packet type

allanjude:
The condition as written did not allow for packet types between SSH2_MSG_LOCAL_MIN and SSH2_MSG_LOCAL_MAX as intended

### DIFF
--- a/packet.c
+++ b/packet.c
@@ -1612,7 +1612,7 @@ ssh_packet_read_poll2(struct ssh *ssh, u_char *typep, u_int32_t *seqnr_p)
 		goto out;
 	if (ssh_packet_log_type(*typep))
 		debug3("receive packet: type %u", *typep);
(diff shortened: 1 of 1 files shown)


Jakuje:
Please, open a bug in https://bugzilla.mindrot.org/ or send the patch to [openssh mailing list](http://lists.mindrot.org/pipermail/openssh-unix-dev/). As you can notice, OpenSSH developers [do not use pull requests in github](https://github.com/openssh/openssh-portable/blob/master/README#L3).

djmdjm:
AFAIK this is intentional, as we don't support any local-use message. Moreover, if we relaxed this condition then the effect would be identical as we don't set handlers for these numbers and the default dispatch.c behaviour is to treat that case as a protocol error too.

FYI we don't use github pull requests as this is just a mirror. Please use https://bugzilla.mindrot.org/ instead.

PR #69, opened 2017-06-05
Removed dependency on sysinternals and psexec. Added PS utility to do host key managment

manojampalam:
Removed dependency on sysinternals and psexec. Added PS utility to do host key managment


manojampalam:
Pull request created by mistake on wrong repo.

PR #70, opened 2017-06-14
[WIP] sshd_config: allow directories in AuthorizedKeysFile=

lucab:
[This PR is for reviews/comments only, patch will be submitted to bugzilla if/once ready]

This commit enhances AuthorizedKeysFile= to accept directory paths
in addition to single files.
It provides an include semantics mimicking `.d` / `run-parts(8)`
approach, offering a consistent way for different entities to add
public keys to a given account without single-file contention.


lucab:
Forwarded to https://bugzilla.mindrot.org/show_bug.cgi?id=2755.

lucab:
@djmdjm I'va rebased and updated this with the comments from Bugzilla, can you please have a look?

bgilbert:
[ssh-key-dir](https://github.com/coreos/ssh-key-dir) is an `AuthorizedKeysCommand` implemented as a workaround for the missing `~/.ssh/authorized_keys.d` support.  It's been shipping in Fedora CoreOS by default for a couple years now.

djmdjm:
I don't love implicitly reading everything in a directory - it feels like it would be easy to make mistakes with this.

Instead, we could implement wildcards in `AuthorizedKeysFile` and `AuthorizedPrincipalsFile` - this would achieve pretty much the same outcome but would require explicit configuration from the admin.

https://github.com/djmdjm/openssh-wip/pull/48 implements this

travier:
This has been implemented in OpenSSH 10.0 (https://bugzilla.mindrot.org/show_bug.cgi?id=2755 & https://www.openssh.com/releasenotes.html & https://github.com/openssh/openssh-portable/commit/85f0c1e75e8f6c5d83b8070918ee2f6ab16d403e) and we have tested that in https://github.com/coreos/ssh-key-dir/issues/188 so this can be closed. Thanks!

PR #74, opened 2017-07-25
permitopen but for -R

biagioni84:
restricts which ports are available for a given user on a remote server when opening remote forwarding ports 
use case: NAT traversing limited to a specified port for each user
on the remote server, on the users  .ssh/authorized_keys , add: permitopen="host:port" and user's public key. it helps mitigate a DoS in case a user's private key is lost 
** if no permitopen is found for the user, all ports are allowed as usual
useful to limit tunneling for nat traversing to a specified port on a by user base

### DIFF
--- a/channels.c
+++ b/channels.c
@@ -3444,6 +3444,27 @@ channel_request_rforward_cancel(struct Forward *fwd)
 }
 
 /*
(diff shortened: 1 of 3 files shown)


biagioni84:
wrong place, using bugzilla instead

PR #75, opened 2017-07-30
Don't require a working DISPLAY for confirmation

tcely:
Anyone who uses 'ssh-add -c' (which appears to be few people) on a non-graphical system is greatly inconvenienced until they replace ssh-askpass with something that can use the terminal and sets DISPLAY before launching ssh-agent.

This appears to be a mistake from refactoring back in 2004. If certain cases need to force the use of SSH_ASKPASS instead of trying the tty the callers and this function will need to be changed to pass their own flags again.

I hit this again recently when using bash on Windows 10.

### DIFF
--- a/readpass.c
+++ b/readpass.c
@@ -177,7 +177,7 @@ ask_permission(const char *fmt, ...)
 	vsnprintf(prompt, sizeof(prompt), fmt, args);
 	va_end(args);
 
-	p = read_passphrase(prompt, RP_USE_ASKPASS|RP_ALLOW_EOF);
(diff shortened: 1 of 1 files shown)


tcely:
For reference the commit that caused this problem is:
https://github.com/openssh/openssh-portable/commit/ce327b62ac28cb6a605fd827a97d978ea31e2860

djmdjm:
OpenSSH 8.4 included support for $SSH_ASKPASS_REQUIRE that allows the $DISPLAY requirement to be bypassed.

PR #76, opened 2017-08-04
Fix incorrect recursion on Hostname via %h

tb3088:
The %n token is defined as the original host value as specified on the command-line. It stands to reason that `hostname %h` is gratuitous and benign by virtue of being a NOOP, `hostname %h.domain` makes no sense. Only `hostname %n.domain` complies with the documentation. Perhaps when %n was added to the list of tokens, nobody went back to update the code. Also ssh.c and readconf.c were not kept in sync despite the comment in both locations.

Given the age of this defect, and the world-wide use of '%h' we may want to run both incarnations of percent_expand back to back so as to not break a billion configuration files.

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -562,7 +562,7 @@ match_cfg_line(Options *options, char **condition, struct passwd *pw,
 	} else if (options->hostname != NULL) {
 		/* NB. Please keep in sync with ssh.c:main() */
 		host = percent_expand(options->hostname,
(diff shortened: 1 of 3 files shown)


Jakuje:
This certainly makes sense to fix. But as you certainly read in the readme, OpenSSH team uses github only as a mirror and patches and bug reports should be sent to mailing list and bugzilla:

https://www.openssh.com/list.html
https://bugzilla.mindrot.org/

PR #78, opened 2017-09-22
Make username case sensitivity

jeddcuison:
The opensource project [Jwala](https://github.com/cerner/jwala) is currently using Cygwin and OpenSsh to manage web and application servers. The application is currently deployed in production environments and has experienced ssh login problems due to the username changing case from time to time . Initial deployment of Cygwin and OpenSsh in a Windows environment is initially fine, ssh accepts any case. Later on, sshd will complain that auth failed due to the username comparison failure since the username that **getpwnam** returns in now in a specific case which does not necessarily equal that of active directory's (It  would have been okay if it did equal active directory's username case since we'll just follow it). As a summary, the pull request's goal is to give users of OpenSsh in Cygwin/Windows the option of having the username case sensitivity configurable to mitigate the above mentioned issue without changing anything in Cygwin or Windows.

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -580,12 +580,22 @@ getpwnamallow(const char *user)
 #endif
 #ifdef HAVE_CYGWIN
 	/*
-	 * Windows usernames are case-insensitive.  To avoid later problems
-	 * when trying to match the username, the user is only allowed to
-	 * login if the username is given in the same case as stored in the
-	 * user database.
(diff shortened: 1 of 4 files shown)


djmdjm:
I don't think we'll make this configurable, but please let's continue the discussion at the open bug. https://bugzilla.mindrot.org/show_bug.cgi?id=2764

PR #82, opened 2018-01-15
Fix scp in case of RemoteCommand in ssh_config

CamJN:
This is based off of the `PermitLocalCommand=no` above the new lines, and seems to work for me.


djmdjm:
This was been committed as 77e05394af21d3f5

Please use https://bugzilla.mindrot.org/ for bugs in future; we only use github as a mirror.

PR #83, opened 2018-01-29
Fail when can't create public key file

WillerZ:
ssh-keygen should exit with a failure result code (non-zero) if the public key cannot be written to the public identity file or if a trailing newline cannot be written to the public identity file. Either of these conditions prevents the user from generating a valid `authorized_keys` file with the de-facto standard `cat id_whatever.pub >> authorized_keys` mechanism.

Setting correct exit codes is important for the benefit of automation which invokes ssh-keygen.

I have chosen to log and not fail if the comment cannot be written because it's an optional field.


Jakuje:
If you are interested in these changes in OpenSSH, you should fill the bug report on http://bugzilla.mindrot.org/ -- the github is used only as a mirror, as already said many times.

PR #84, opened 2018-02-06
Remove redundant null tests before RSA_free.

coypoop:
It does null test.


daztucker:
Thanks for the patch.  As it says in README we don't use generally use github pull requests.  Anyway, the change you are proposing has been done as part of a larger cleanup: https://github.com/openssh/openssh-portable/commit/7cd31632

Thanks.

PR #89, opened 2018-03-27
fix typo causing compilation with -DDEBUG_PK

matthewaveryusa:
fix typo causing compilation with -DDEBUG_PK -- siglen replaced with slen.


Jakuje:
> Please read https://www.openssh.com/report.html for bug reporting
instructions and note that we do not use Github for bug reporting or
patch/pull-request management.

https://github.com/openssh/openssh-portable/blob/master/README#L3

daztucker:
Thanks, this was fixed in 086cc614f from upstream.

PR #90, opened 2018-03-31
Update auth2-hostbased.c

RomanEmelyanov:
Compilation error with option DEBUG_PK


Jakuje:
Isn't this the same as #89 ?

RomanEmelyanov:
U A right!

PR #92, opened 2018-05-15
Unify return values in functions from sftp-client.

malyzelenyhnus:
Affected functions used both 0/-1 and SSH2_FX_* constants as return
values. It caused that errors were sometimes not properly detected.


djmdjm:
This was fixed in f6a59a22b0c1 some time ago

PR #93, opened 2018-06-03
build for iOS on Mac OSX (eg. arm64/armv7)

0neday:


bogardon:
Seems odd that this was an issue. How does apple compile this themselves?

PR #97, opened 2018-08-22
Spelling fixes for 7.8p1

knweiss:
Another `codespell` pass before the 7.8p1 release.


daztucker:
Thanks, these have all been committed either in portable or upstream where appropriate (the latter will take a little time to trickle down to this tree).

Thanks.

PR #98, opened 2018-08-28
Implementing RFC 4254 section 6.9 : Signals

JawaGL:
Hello every one,

Following the discussion “trying to resurrect discussion about "Cannot signal a process over a channel (rfc 4254, section 6.9)"” on the mailing list (link: https://marc.info/?t=153077282000003&r=1&w=2), I would like to put an eventual patch here so it is not forgotten.

The code is a gathering of other codes that were proposed on Bugzilla with some modifications.

This patch allows the ssh client to send “signal” messages (Section 6.9 of RFC 4254) via the mechanism of escape characters, eg : ‘~ST’ to send a SIGTERM to the remote process.

Upon reception of such message, the server will send the chosen signal to the process group of the last forked process or the process on the foreground depending if it is in exec or shell mode.

There is also a rudimentary test for these functionalities in the test suite.

Best regards

### DIFF
--- a/clientloop.c
+++ b/clientloop.c
@@ -152,6 +152,7 @@ static time_t control_persist_exit_time = 0;
 /* Common data for the client loop code. */
 volatile sig_atomic_t quit_pending; /* Set non-zero to quit the loop. */
 static int last_was_cr;		/* Last character was a newline. */
+static int last_was_sr;	  /* Last character sequence was signal request*/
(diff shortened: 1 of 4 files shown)


djmdjm:
The server side of this landed in openssh-7.9, but we didn't commit the client side. If you have a use-case for that then we could consider it.

PR #99, opened 2018-09-03
getgrouplist can return with ngroups > _SC_NGROUPS_MAX

darrenmoffat:
On Solaris, and maybe on other platforms too, getgrouplist() can return with ngroups set to a value greater than the value that sysconf returned for _SC_NGROUPS_MAX.  This can happen when the name service has more entries available.

For example Solaris 11.4 _SC_NGROUPS_MAX is 1024 but if I put 1028 entries for a user into the nameservice (files is fine LDAP isn't needed for this) then after the call to getgrouplist() ngroups will
be 1028.

The current code expects that can can log "getgrouplist: groups list too small" in this case.  However
ngroups is now set to a value larger than was allocated with xcalloc, this can result in ga_free() attempting to free more entries than we actually allocated.

A possible fix is to reset the global ngroups to be the value allocated so that ga_free() will
have the correct size array.

### DIFF
--- a/groupaccess.c
+++ b/groupaccess.c
@@ -40,6 +40,7 @@
 #include "log.h"
 
 static int ngroups;
+static int ngroups;
 static char **groups_byname;
 
 /*
@@ -52,13 +53,14 @@ ga_init(const char *user, gid_t base)
 	gid_t *groups_bygid;
 	int i, j;
 	struct group *gr;
+	int allocated_ngroups;
 
 	if (ngroups > 0)
 		ga_free();
 
 	ngroups = NGROUPS_MAX;
 #if defined(HAVE_SYSCONF) && defined(_SC_NGROUPS_MAX)
-	ngroups = MAX(NGROUPS_MAX, sysconf(_SC_NGROUPS_MAX));
+	allocated_ngroups = ngroups = MAX(NGROUPS_MAX, sysconf(_SC_NGROUPS_MAX));
 #endif
 
 	groups_bygid = xcalloc(ngroups, sizeof(*groups_bygid));
@@ -66,6 +68,8 @@ ga_init(const char *user, gid_t base)
 
 	if (getgrouplist(user, base, groups_bygid, &ngroups) == -1)
 		logit("getgrouplist: groups list too small");
+	if (ngroups > allocated_ngroups)
+		ngroups = allocated_ngroups;
 	for (i = 0, j = 0; i < ngroups; i++)
 		if ((gr = getgrgid(groups_bygid[i])) != NULL)
 			groups_byname[j++] = xstrdup(gr->gr_name);
--- a/groupaccess.c
+++ b/groupaccess.c
@@ -40,7 +40,6 @@
 #include "log.h"
 
 static int ngroups;
-static int ngroups;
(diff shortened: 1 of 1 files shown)


daztucker:
"can return with ngroups set to a value greater than the value that sysconf returned for _SC_NGROUPS_MAX."  For small values of "MAX", I guess.  I can see that doing so would cause problems.  I'd rather make it work as expected instead of truncating the list, eg:

       while (getgrouplist(user, base, groups_bygid, &ngroups) == -1) {
               groups_bygid = xreallocarray(groups_bygid, ngroups,
                   sizeof(*groups_bygid));
               groups_byname = xreallocarray(groups_byname, ngroups,
                   sizeof(*groups_byname));
       }

(this is a -portable only issue, the OpenBSD code uses a purely statically sized array).

darrenmoffat:
On 04/09/2018 06:56, Darren Tucker wrote:
>
> "can return with ngroups set to a value greater than the value that 
> sysconf returned for _SC_NGROUPS_MAX." For small values of "MAX", I 
> guess. I can see that doing so would cause problems. I'd rather make 
> it work as expected instead of truncating the list, eg:
>
> |while (getgrouplist(user, base, groups_bygid, &ngroups) == -1) { 
> groups_bygid = xreallocarray(groups_bygid, ngroups, 
> sizeof(*groups_bygid)); groups_byname = xreallocarray(groups_byname, 
> ngroups, sizeof(*groups_byname)); }|

|For this particular case where OpenSSH code is itself doing the 
matching to the array expanding it out beyond _SC_NGROUPS_MAX to the 
number returned by getgrouplist will work just fine (at least on Solaris 
which is all I've tested).  In the more general case where the group 
list is being set in the users kernel cred _SC_NGROUPS_MAX should be 
respected, and at least on Solaris we have a hard limit of 1024 in 
kernel for the maximum number of groups a user can be in.

Does the while loop guard need to verify that ngroups is actually bigger 
than we
previously allocated?  At least in the Solaris implementation 
getgroupslist could return -1
in the case of invalid arguments (shouldn't happen here) or failures in 
malloc (that it or functions it calls perform) and it won't update 
ngroups in those cases.

|||||
>
> (this is a -portable only issue, the OpenBSD code uses a purely 
> statically sized array).
>
Yes I saw that, thanks for confirming this is a -portable only issue.

Darren

daztucker:
Fair point.  I did check the man pages for both Linux and OpenBSD and those only mentioned undersized array as the only reason it'd return -1.  (Solaris doesn't have man page for it, at least on my system.) I think I'm going to go allow only one retry, eg:

        groups_bygid = xcalloc(ngroups, sizeof(*groups_bygid));
        while (getgrouplist(user, base, groups_bygid, &ngroups) == -1) {
                if (retry++ > 0)
                        fatal("getgrouplist: groups list too small");
                groups_bygid = xreallocarray(groups_bygid, ngroups,
                    sizeof(*groups_bygid));
        }
        groups_byname = xcalloc(ngroups, sizeof(*groups_byname));

daztucker:
Should be resolved by https://github.com/openssh/openssh-portable/commit/26788330.

Thanks.

PR #102, opened 2018-10-09
get / put struct passwd

dag-erling:
Add `sshbuf_{get,put}_passwd()` for `struct passwd`.  Since both `size_t` and `time_t` can vary in size from one system to another, add wrappers for both which use either `u32` or `u64` depending on what autoconf says.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3498,10 +3498,10 @@ fi
 AC_CHECK_TYPES([long long, unsigned long long, long double])
 
 # Check datatype sizes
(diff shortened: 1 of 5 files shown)


daztucker:
Applied the configure.ac one, thanks.  The other one probably need to go up into OpenBSD at least in part.  I'll take a look.

djmdjm:
This was fixed in b2bcec13f1 and shipped in V_8_5

PR #103, opened 2018-10-09
Ignore missing keys unless explicitly configured.

dag-erling:
When adding a host key file to the list, if `file` refers to our `"[default]"` constant and `path` does not exist, ignore it. This silences warnings on systems where certain keys are intentionally not present without requiring the administrator to explicitly list those that are. We will still get an error if no keys were found at all.


djmdjm:
Similar patch applied in 928f1231f65f88

PR #105, opened 2018-10-27
close stderr of persistent proxy command if not in debug mode

sprohaska:
The patch should likely be first applied to upstream.  I've sent it to openssh-unix-dev@mindrot.org, but haven't yet received a confirmation that the message has been accepted.

If the parent becomes a new persistent connection master daemon, stderr
of the proxy command should be detached, too, similar to the master
daemon's stderr, as changed in
openssh-portable@d2d6bf864e52af8491a60dd507f85b74361f5da3,
upstream@4fb726f0fdcb155ad419913cea10dc4afd409d24 and discussed in
bz#1988.

### DIFF
--- a/sshconnect.c
+++ b/sshconnect.c
@@ -78,6 +78,7 @@ static int matching_host_key_dns = 0;
 static pid_t proxy_command_pid = 0;
 
 /* import */
+extern int debug_flag;
 extern Options options;
 extern char *__progname;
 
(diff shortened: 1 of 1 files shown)


djmdjm:
I've committed a similar change in ccef7c4faf9

sprohaska:
Thanks.

PR #107, opened 2018-11-09
Windows support: Account for path differences

manojampalam:
Issue: 
Unix and Windows have different file system path rules. While Unix paths start with '/', Windows start with a drive letter and ':'. There are a handful of places where this checks for "absolute paths" are made. If we can route these checks via a unitity function, we could overload the utility for Windows and keep the rest of the calls clean (from #ifdef logic). 

Fix:
Changes include utility function that could be overloaded for Windows in one place.


djmdjm:
I've committed in 2a35862e66 a variant of this that caught a few more cases that should use path_absolute()

PR #108, opened 2018-11-09
Add option to compile with no zlib support

manojampalam:
Issue: In Windows, zlib isn't officially supported yet. Truly compiling without zlib requires a lot of changes, I avoided these by creating a dummy zlib library for Windows. We still need some changes in core code for the client and server stacks to not advertise zlib support on the wire.

Fix: Add a macro to opt out zlib support. By default, if the macro is not defined, zlib compression support is included, so it makes no difference when compiling/building for Unix flavors

### DIFF
--- a/myproposal.h
+++ b/myproposal.h
@@ -180,7 +180,11 @@
 
 #endif /* WITH_OPENSSL */
 
+#ifdef NO_ZLIB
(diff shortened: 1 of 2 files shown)


manojampalam:
zlib is being added to Windows fork. This change is no longer necessary.

PR #109, opened 2018-11-09
Normalize incoming user before processing username relying logic

manojampalam:
This is to primarily support scenarios in Windows. The changes otherwise have no impact on behavior in Unix. 
  - user names are case insensitive in Windows with multiple valid domain scoping formats (user@domain, domain\user, etc). 
  - Changes include normalizing user name (done by getpwnam on Windows) upfront before processing user name based logic.

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -567,10 +567,10 @@ getpwnamallow(const char *user)
 	auth_session_t *as;
 #endif
 #endif
-	struct passwd *pw;
+	struct passwd *pw = getpwnam(user);
 	struct connection_info *ci = get_connection_info(1, options.use_dns);
 
-	ci->user = user;
+	ci->user = pw ? pw->pw_name : user;
 	parse_server_match_config(&options, ci);
 	log_change_level(options.log_level);
 	process_permitopen(ssh, &options);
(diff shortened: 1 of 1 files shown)


manojampalam:
The underlying problem solved by getpwnam on Windows is that the input username can come in various valid forms and we would like to resolve these something unique per user. Please note that, getpwnam is not a Windows primitive, its implemented in POSIX compat library in our fork. 
In summary it does lowercasing and name format translation. In detail, it does the following:

For local accounts, getpwnam(input) returns pw_name = "localuser" for the following valid input:
- hostname\localUSeR
- hostIP\localuser (could be IPV4 or IPv6)
- localuSer@hostname
- localusER@hostIP
- LocalUser

for domain accounts, getpwnam(input) returns pw_name = "domain\user" for the following valid input:- 
- domain\UseR
- Domain.company.com\uSeR
- user@domaiN
- user@domain.company.com
- User (when "user" is only found in "domain")

As detailed [here](https://github.com/PowerShell/Win32-OpenSSH/wiki/sshd_config#allowgroups-allowusers-denygroups-denyusers), to avoid confusion and simplify sshd admin experience, we mandated the username to conform to the following rules in sshd_config:
- local accounts: no domain qualification
`AllowUsers localuser@192.168.2.23` 
- domain accounts: strictly in NetBIOS format (domain\user) 
`DenyUsers contoso\admin@192.168.2.23`


This approach also helped us to keep the internal user name based match processing rules simple and unambiguous.

-------------

As for root/toor, I'm inferring that we get different passwd entries. Would these entries have the same pw_name ?

PR #112, opened 2018-12-04
add in debug showing public key of user attempting login

catskul:
For debug and security/forensic reasons it's sometimes valuable to know the public key of users attempting to log in.

This patch adds in full public key printing at debug1.


djmdjm:
A similar capability was included in OpenSSH 8.0

PR #113, opened 2018-12-13
Don't pass loginmsg by address now that it's an sshbuf*

kadler:
In 120a1ec74, loginmsg was changed from the legacy Buffer type to struct sshbuf*, but it missed changing calls to sys_auth_allowed_user and sys_auth_record_login which passed loginmsg by address. Now that it's a pointer, just pass it directly.

This only affects AIX, unless there are out of tree users.


daztucker:
merged, thanks.

PR #114, opened 2018-12-30
kexc25519.c fixed all-zero test

janmojzis:
Hello,
there is bug in "curve25519-sha256" implementation.
In the file kexc25519.c, in function kexc25519_shared_key is 'all-zero-pk' check which checks
if the public-key is not zero. But the check is not complete.
You shoud check if the public key is not zero, p (2^255-19), 2p, ...

Is much better to check the scalar product after scalar multiplication (shared-secret) because if pk modulo p is zero, then the shared secret is also zero.

You can check it using script (autopkgtest for TinySSH server):
https://salsa.debian.org/debian/tinyssh/blob/master/debian/tests/02handshake

run (before and after patch) e.g.:
./02handshake 127.0.0.1 22

Jan.

### DIFF
--- a/kexc25519.c
+++ b/kexc25519.c
@@ -64,14 +64,16 @@ kexc25519_shared_key(const u_char key[CURVE25519_SIZE],
     const u_char pub[CURVE25519_SIZE], struct sshbuf *out)
 {
 	u_char shared_key[CURVE25519_SIZE];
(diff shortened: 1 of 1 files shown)


djmdjm:
This got submitted upstream and was in the openssh-8.0 release.

PR #116, opened 2019-02-10
ForwardIdentityAgent - a new ssh option to set a dedicated ssh-agent for ForwardAgent

prbinu:
# New Feature Request

## Problem Statement
We have a requirement to use `ForwardAgent` to enable SSH `sudo` authentication using identities cached in our local `ssh-agent`. However the agent forwarding has an unintended security consequence. With agent forwarding (active session) anyone with sufficient permissions (e.g. sudo users or an attacker who managed to compromised the host) on the remote host can potentially impersonate the user and connect to a 3rd host by performing authentication using login identities (keys) from your `ssh-agent` (lateral movement).

To mitigate this problem, one option is to use a separate identity for `sudo`, different from SSH login identity. But `openssh` client forwards the same authentication agent used for SSH login, exposing our login identities to remote hosts. 

As a workaround, it is technically possible to setup `RemoteForward` to local `ssh-agent` for sudo authentication, but that requires some hacks/less-ideal changes in the server side.

The proposed feature will allow us to separate login and ssh authentication identities by using two different `ssh-agent` in the workstation.

More info: see [here](https://medium.com/@prbinu/touch2sudo-enable-remote-sudo-two-factor-authentication-using-mac-touch-id-df638b7da594)

##  `ForwardIdentityAgent`

This option specifies the UNIX-domain socket used to communicate with the forwarded authentication agent. `ForwardIdentityAgent` can be used in conjunction with `ForwardAgent` to configure a separate `ssh-agent` for agent forwarding. This is similar to `IdentityAgent`, but for `ForwardAgent`.

This feature is helpful in cases where a user needs to enable `ForwardAgent` (e.g. for SSH `sudo` authentication using a dedicated identity), but doesn't want to expose **cached** SSH login identities (certificates) to remote hosts.

## Usage

```bash
# default ssh-agent that contains login identity key
$ env | grep SSH_AUTH
SSH_AUTH_SOCK=/tmp/ssh-z1Lsp0G9s6wh/agent.27887

# another ssh-agent instance that contains sudo authentication identity key
$ SUDO_SSH_AUTH_SOCK=/tmp/ssh-A5JsI9UpPWzg/agent.27955
$ ssh -o "ForwardIdentityAgent $SUDO_SSH_AUTH_SOCK" -A bob@172.31.2.67
...
bob@ip-172-31-2-67:~$ sudo whoami
root

```

### DIFF
--- a/authfd.c
+++ b/authfd.c
@@ -82,18 +82,16 @@ decode_reply(u_char type)
 		return SSH_ERR_INVALID_FORMAT;
 }
 
-/* Returns the number of the authentication fd, or -1 if there is none. */
+/* Returns socket fd, or -1 if there is none. */
 int
-ssh_get_authentication_socket(int *fdp)
+get_authentication_socket(const char *authsocket, int *fdp)
 {
-	const char *authsocket;
 	int sock, oerrno;
 	struct sockaddr_un sunaddr;
 
 	if (fdp != NULL)
 		*fdp = -1;
 
-	authsocket = getenv(SSH_AUTHSOCKET_ENV_NAME);
 	if (authsocket == NULL || *authsocket == '\0')
 		return SSH_ERR_AGENT_NOT_PRESENT;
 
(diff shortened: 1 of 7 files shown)


prbinu:
Hi @djmdjm, I would appreciate if you could review this request (when you find some time)

djmdjm:
I think I have a patch somewhere that makes ForwardAgent a multistate option: "yes" | "no" | $environment | /path/to/agent/sock

I'll try to dig it out when I get back from travelling next week.

prbinu:
thank you!

prbinu:
@djmdjm Just checking to see if you have any update to share on this request.

prbinu:
@djmdjm  happy new year!
please let me know if you need any help with merging/enabling this feature.

thank you!

djmdjm:
Something similar was committed in 40be78f50 - the ssh_config ForwardAgent directive can now accept a path (or environment variable name) to forward instead of $SSH_AUTH_SOCK

prbinu:
thanks @djmdjm 
One QQ - What would be the new behavior for `-A` SSH option, which used to be a boolean until now.

djmdjm:
-A is still a boolean and that can't easily be changed, as optional
processing of getopt() flags is unreliable. OTOH,
-oForwardAgent=$OTHER_SSH_AUTH_SOCK would work on the commandline

PR #117, opened 2019-02-16
Add RHEL 7 build settings to contrib/redhat/openssh.spec

nkadel:
Use %global instead of %define univerally.
Add epoch numbers for openssl dependencies on RHEL 7
Require /usr/bin/login intead of /bin/login for RHEL compatibility


daztucker:
I had a look at these prior to the release, but the commits don't explain *why* some of the changes were made, in particular the define->global one, and most of the others depend on that.

nkadel:
Oh! %global is now considered standard for all .spec files in every RHEL and Fedora release.

nkadel:
Would you like the switch to '%global" as a different pull request?

nkadel:
I'll re-merge after pull request 156, when %define gets switched to the RHEL default %global.

PR #119, opened 2019-02-23
pledge: setting errno instead of aborting

devnexen:


daztucker:
Why would you want to do that?.  If the sandbox process is compromised terminating it noisily is the safest thing to do.

PR #120, opened 2019-02-23
Few mem leak fixes

devnexen:


emaste:
Coverity noted the issue in auth-options.c in FreeBSD

PR #121, opened 2019-03-06
Avoid integer overflow when compiling for 32bit

flippy84:
When hashing >= 512 MiB of data context->count will be updated with an invalid value resulting in the wrong hash being calculated.


daztucker:
Thanks, this was fixed in adcaf40fd0a180e6cb5798317fdf479b52e3c09a.  The same potential problem existed in sha2, which took more work to update (09159594a3bbd363429ee6fafde57ce77986dd7c and 11cba2a4523fda447e2554ea457484655bedc831).

PR #122, opened 2019-03-13
New feature: Set PAM service name in sshd_config file

CorvusCorax:
This allows different pam auth configurations based on
used AuthenticationMethod and Match directives.

See https://pastebin.com/SuCG6dft  for the problem this solves.
There are 2 new config directives:

From Manpage:
```

     **PAMServiceName**
             Specifies the service identifier to be used for pluggable authentication modules
             (PAM).  If set to none, the sshd executable name (usually sshd) is used.  If set
             to name, authentication options can be configured in a matching file in
             /etc/pam.d/name.  The default is none.

     **PasswordPAMServiceName**
             Specifies the service identifier to be used for pluggable authentication modules
             (PAM) for PasswordAuthentication only. This is only used for the authentication
             (auth) PAM aspect. For account and session management, as well as
             ChallengeResponseAuthentication, the service name set in PAMServiceName is
             applied. This can be helpful to implement 2 factor authentication.  Example:

                   AuthenticationMethods "publickey,keyboard-interactive
                   password,keyboard-interactive"
                   PAMServiceName sshd_2factor
                   PasswordPAMServiceName sshd_password

             In this example, the challenge response authentication would ask the user for the
             second factor, for example an OATH token, while the first factor could be either
             provided by ssh-key or password. This requires two different PAM configurations to
             be used, in this example /etc/pam.d/sshd_2factor and /etc/pam.d/sshd_password. If
             set to none, the value of PAMServiceName is used.  The default is none.
```

Both new config options are allowed within **Match** directives, which allows very sophisticated setup, where different hosts can undergo completely different authentication paradigms.
If neither of the new options is set (or set to **none**) the behavior is as it used to, setting the pam service either to the program name or to the fixed override given with
./configure --with-pam-service=name

Implementation:
Instead of hardcoding the service name to SSHD_PAM_SERVICE, the pam handle is now initiated with a dynamic service name. Since this can change unexpectedly when used in Match directives, pam is now reinitialized every time pam_init() is called, even if the user is identical.

sshpam_auth_passwd() which implements PAM based PasswordAuthentication now uses its own separate PAM handle used for authentication only (the regular one is used for account and session management)

I'm looking forward to your feedback.

### DIFF
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -239,6 +239,7 @@ pthread_join(sp_pthread_t thread, void **value)
 
 
 static pam_handle_t *sshpam_handle = NULL;
+static pam_handle_t *sshpam_password_handle = NULL;
 static int sshpam_err = 0;
 static int sshpam_authenticated = 0;
 static int sshpam_session_open = 0;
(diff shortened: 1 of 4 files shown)


TommiTallgren:
I hope that Softagram visual analysis of this PR could help the reviewers. This was made originally as a demo to SSH.COM f2f meeting.

### Softagram Impact Report for pull/1 (head commit: f3abe4ee116f)


### :star: Change Overview


Showing the changed files, dependency changes and the impact - click for full size


#### :star: Details of Dependency Changes


details of dependency changes - click for full size
*(Open in [Softagram Desktop](https://softagram.com/desktop) for full details, need to login to opensource.softagram.com first)*


#
Give feedback of this report to support@softagram.com

daztucker:
PAMServiceName was added in OpenSSH 9.8p1
https://www.openssh.com/txt/release-9.8

PR #123, opened 2019-03-19
Fix SSH incorrect command line conversion

fcharlie:
When I run ssh, if the command line of the exec request contains a space, the command line parsing on the server is inconsistent with the expectation.

Example:

Client:

```shell
ssh git@localhost echo "There are spaces in the statement" done
```

We print the command line on the server, and the command line received on the server is actually as follows:

```shell
echo There are spaces in the statement done
```
You can find that SSH handles command line parameter errors.

Old code:

```c
//https://github.com/openssh/openssh-portable/blob/9edbd7821e6837e98e7e95546cede804dac96754/ssh.c#L1061
		/* A command has been specified.  Store it into the buffer. */
		for (i = 0; i < ac; i++) {
			if ((r = sshbuf_putf(command, "%s%s",
			    i ? " " : "", av[i])) != 0)
				fatal("%s: buffer error: %s",
				    __func__, ssh_err(r));
		}
```

This PR escapes a string containing special characters, spaces, and constructs a string that can be handled correctly by the shell.

 Thansk --> [Windows: syscall.EscapeArg](https://github.com/golang/go/blob/50bd1c4d4eb4fac8ddeb5f063c099daccfb71b26/src/syscall/exec_windows.go#L26)

### DIFF
--- a/ssh.c
+++ b/ssh.c
@@ -575,6 +575,96 @@ set_addrinfo_port(struct addrinfo *addrs, int port)
 	}
 }
 
+static const char* escape_argument(char *buf, int bufsize, char *arg){
+	int len = strlen(arg);
+	if (len == 0){
+		return "\"\"";
+	}
+	if(len+2>=bufsize){
+		return arg;
+	}
+	int hasspace, i ,n;
+	hasspace = 0;
+	n = len;
+	for (i=0; i<len; i++){
+		switch(arg[i]){
+			case '"':
+			case '\\':
+			n++;
+			break;
+			case ' ':
+			case '\t':
+			hasspace =1;
+			break;
+			default:
+			break;
+		}
+	}
+	if(hasspace){
+		n+=2;
+	}
+	if (n == len||bufsize+1<n){
+		return arg;
+	}
+	int j=0;
+	int slashes=0;
+	if(hasspace){
+		buf[j]='"';
+		j++;
+	}
+	for(i=0; i<len; i++){
+		switch(arg[i]){
+			case '\\':
+				slashes++;
+				buf[j]=arg[i];
+			break;
+			case '"':{
+				for(;slashes>0;slashes--){
+					buf[j]='\\';
+					j++;
+				}
+				buf[j]='\\';
+				j++;
+				buf[j]=arg[i];
+			}
+			break;
+			default:
+				slashes=0;
+				buf[j]=arg[i];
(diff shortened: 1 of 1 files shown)


daztucker:
Unfortunately ssh has no knowledge of what the shell is at the other end or what its quoting rules are, and the protocol (RFC4254 section 6.5) only specifies an unstructured string for the "exec" channel.

Making assumptions like you are proposing will likely cause problems in some cases.  If it matters to you then you need to either do the (double) quoting yourself or use a construction like:

$ ssh localhost /bin/sh <<EOD
echo "There  are  spaces  in  the  statement" done
EOD

which prevents some of these problems by avoiding one level of quote processing on the client and specifying the exact shell on the server side.

fcharlie:
@daztucker The SSH RFC specification does have a lot of deficiencies, but when openssh starts a child process on the server, it uses `-c` which means it will follow the command-line escaping rules.

```
      byte      SSH_MSG_CHANNEL_REQUEST
      uint32    recipient channel
      string    "exec"
      boolean   want reply
      string    command

   This message will request that the server start the execution of the
   given command.  The 'command' string may contain a path.  Normal
   precautions MUST be taken to prevent the execution of unauthorized
   commands.
```

'echo' is really just an example that doesn't matter. 

But we do have some troubles, such as cloning a repository with a space on the server.
Example:

```shell
git clone ssh://git@example.com/some%20user/repo.git
# ssh git@example.com git-upload-pack "some user/repo.git"
```

Unfortunately, SSH is not perfect.

daztucker:
>  when openssh starts a child process on the server, it uses -c

that's an implementation detail of OpenSSH's sshd.  Other servers, especially on other platforms, could do anything.

>  it will follow the command-line escaping rules
... of whatever the user's login shell happens to be (on a Unix-ish system, anyway).  So it might be a Bourne-compatible shell (sh bash ksh) or it might be something else (csh tcsh fish) or it might be some custom shell or non-Unix system with entirely different behaviour.

It would be possible to do it in an "exec"-like vendor extension that takes a list of string args.  There's some discussion around this in http://bugzilla.mindrot.org/show_bug.cgi?id=2283 but the cost/benefit tradeoff is not clear.

fcharlie:
@daztucker If there is SSH-3.0 in the future, I hope this problem can be solved. thank you for your reply.

PR #125, opened 2019-04-19
regress: update utimensat for Linux semantics

chutz:
The current test checks that utimensat() fails when run on a symlink
with AT_SYMLINK_NOFOLLOW. These are not the semantics of this flag on
Linux, from the man page:

    AT_SYMLINK_NOFOLLOW
        If pathname specifies a symbolic link, then update the timestamps
        of the link, rather than the file to which it refers.

This modified the test (gated under `#ifdef __linux__`) to check for the
Linux semantics, so this test will not always fail on Linux.

### DIFF
--- a/openbsd-compat/regress/utimensattest.c
+++ b/openbsd-compat/regress/utimensattest.c
@@ -88,8 +88,30 @@ main(void)
 	if (symlink(TMPFILE2, TMPFILE) == -1)
 		fail("symlink", 0, 0);
 
+#ifdef __linux__
+	/*
+	 * The semantics of the original test are wrong on Linux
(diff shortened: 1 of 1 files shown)


daztucker:
The compat code doesn't have a way to update the times on the symlink, so it fails instead of following a symlink when explicitly told not to.  The test is just intended to ensure that it doesn't follow the symlink, not that the failure is the desired behaviour.   I think we should change the test to instead check that the times on the destination file are not updated rather the checking for an explicit failure.

daztucker:
I've updated the test as described and it now passes with the Linux native implementation.  Thanks.

PR #126, opened 2019-04-26
pthread_create(3) returns positive values on failure.

enh:
Found by inspection after finding similar bugs in other code used by
Android.


daztucker:
Applied, thanks.

PR #128, opened 2019-04-30
Fixing compilation issue for OPENSSH with no support for OPENSSL_HAS_ECC 

SK851831:
All the KEX_ECDH_SHA2 related cases should come under OPENSSL_HAS_ECC macro since the definitions of the functions called under these switch cases are available only under OPENSSL_HAS_ECC macro causing OPENSSH to fail in compilation without OPENSSL_HAS_ECC macro


daztucker:
This was already fixed by97370f6c2c3b825f8c577b7e6c00b1a98d30a6cf (after 8.0) which adds no-op implementations of those functions without adding the added maintenance of those ifdefs.

PR #129, opened 2019-05-03
Use more accurate "unresponsive" instead of "inactive"

jblaine:
The ClientAliveInterval setting is grossly misunderstood by administrators to be some sort of "idle session timeout" tunable. I think a good deal of this confusion comes from the usage of the overloaded word "inactive" in this man page when "unresponsive" is meant.


daztucker:
Applied upstream, thanks.  I also made the same change to ssh_config.5.

PR #130, opened 2019-05-04
Added source port feature

Zibri:
ssh is since a long time missing the feature to set an arbitrary source port for the connection.
thus forcing people to use "ProxyCommand" option and use nc or ncat to force the source port.
I added the feature in the client directly so, from root user you can use -Z portnum to specify
the source port.

### DIFF
--- a/ssh.c
+++ b/ssh.c
@@ -167,6 +167,8 @@ char *config = NULL;
  */
 char *host;
 
+char *source_port = NULL;
+
 /* Various strings used to to percent_expand() arguments */
 static char thishost[NI_MAXHOST], shorthost[NI_MAXHOST], portstr[NI_MAXSERV];
 static char uidstr[32], *host_arg, *conn_hash_hex;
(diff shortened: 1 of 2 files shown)


Zibri:
Note: this should be put in options and used also in scp/sftp

Zibri:
Added -Z portnum option in:
ssh
scp
sftp

matthieu-labas:
> Perhaps it could be implemented by extending the existing BindAddress option, e.g.
> `ssh -oBindAddress=*:1234`
> would try to use source port 1234. What do you think?

It can be useful for further firewall filtering, but I agree it's not common so extending `BindAddress` would be nice.

Zibri:
I think that like most of shh options I hate that syntax…  but the important is that it will be implemented. My patch to ssh/sct/sftp works perfectly in the meanwhile. 😝  From: Matthieu LabasSent: Friday, March 27, 2020 14:23To: openssh/openssh-portableCc: Zibri; AuthorSubject: Re: [openssh/openssh-portable] Added source port feature (#130) Perhaps it could be implemented by extending the existing BindAddress option, e.g.ssh -oBindAddress=*:1234would try to use source port 1234. What do you think?It can be useful for further firewall filtering, but I agree it's not common so extending BindAddress would be nice.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

PR #131, opened 2019-05-24
allow s390 specific ioctl for ecc hardware support

hfreude:
Adding another s390 specific ioctl to be able to support ECC hardware acceleration
to the sandbox seccomp filter rules.

Now the ibmca openssl engine provides elliptic curve cryptography support with the
help of libica and CCA crypto cards. This is done via jet another ioctl call to the zcrypt
device driver and so there is a need to enable this on the openssl sandbox.

Code is s390 specific and has been tested, verified and reviewed.

Please note that I am also the originator of the previous changes in that area.
I posted these changes to Eduardo and he forwarded the patches to the openssl
community.

Signed-off-by: Harald Freudenberger <freude@linux.ibm.com>
Reviewed-by: Joerg Schmidbauer <jschmidb@de.ibm.com>


daztucker:
Applied, thanks.

PR #132, opened 2019-05-26
openssl-devel is obsoleted by libssl-devel

sorin25:
openssl-devel is no longer installable via the cygwin setup and it's hidden by default, so you can't see the replacement very easy.


daztucker:
Applied, thanks.

PR #133, opened 2019-06-24
Update README doc to include missing test cases

jitendrasharma1989:
Readme regress document is missing various individual tests,
which are supported currently. Update README to
include those test cases.


jitendrasharma1989:
@djmdjm Could you please review this PR.

daztucker:
Applied, thanks.  I'm wondering if we'd be better off replacing the list with a shell one-liner that won't get stale.

PR #134, opened 2019-06-25
Remove unnecessary unlink

eriksjolund:
Remove unnecessary

    unlink(xauthfile);

Just before that line writing to the string _xauthfile_ failed.

Signed-off-by: Erik Sjölund <erik.sjolund@gmail.com>


daztucker:
Applied upstream, thanks.

PR #135, opened 2019-07-05
Bug #3030: window change events not passed when using session multipl…

darrenmoffat:
When the ssh client code calls platform_pledge_mux() on Solaris it will call solaris_drop_privs_root_pinfo_net_exec().  This function drops too many privileges.

It is supposed to be equivalent to pledge("stdio proc tty", ...)

Note "proc" is kept because of the need to call kill(2).  The Solaris implementation of this is currently dropping PRIV_PROC_SESSION which is needed to be able to send signals to processes the user owns but are outside of the current session.


djmdjm:
Applied - thanks!

PR #136, opened 2019-07-08
Remove perm_ok argument from sshkey_load_private_type

jitendrasharma1989:
Remove perm_ok, which could potentially be used to check
for private identification file permissions.
But as sshkey_load_private_type would return
SSH_ERR_KEY_BAD_PERMISSIONS for bad permission.
Hence perm_ok seems redundant.

### DIFF
--- a/authfile.c
+++ b/authfile.c
@@ -164,10 +164,9 @@ sshkey_perm_ok(int fd, const char *filename)
 	return 0;
 }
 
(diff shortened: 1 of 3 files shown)


jitendrasharma1989:
@djmdjm , @daztucker could you please help in reviewing this patch.
The idea of this patch came from TODO list, which says to fix XXX comments.

I have submitted this patch in openssh-unix-dev mailing list also.
In continuation to this patch i have also submitted another patch in mailing list.(I have not yet created Pull request for that another patch.)

PR #141, opened 2019-07-23
Correct typo in the sshd_config(5) manual page

hnj2:
One comma in the listing of algorithms used by default for CASignatureAlgorithms was a dot.


daztucker:
Applied to OpenBSD, thanks.  I'll get pulled into -Portable next time we do a sync.

PR #142, opened 2019-08-22
sandbox-seccomp-filter: allow mprotect syscall

yegortimoshenko:
This allows to use portable OpenSSH with OpenBSD malloc and GrapheneOS hardened_malloc.

See:
- https://github.com/openbsd/src/blob/df69c215c7c66baf660f3f65414fd34796c96152/lib/libc/stdlib/malloc.c#L489
- https://github.com/GrapheneOS/hardened_malloc/blob/a32e26b8e9987177ef7573594b2aab2104ca9d3f/memory.c#L51
- GrapheneOS/hardened_malloc#97

### DIFF
--- a/sandbox-seccomp-filter.c
+++ b/sandbox-seccomp-filter.c
@@ -202,6 +202,9 @@ static const struct sock_filter preauth_insns[] = {
 #ifdef __NR_mmap2
 	SC_ALLOW(__NR_mmap2),
 #endif
+#ifdef __NR_mprotect
+	SC_ALLOW(__NR_mprotect),
+#endif
 #ifdef __NR_mremap
 	SC_ALLOW(__NR_mremap),
 #endif


thestinger:
The mprotect call you're linking in OpenBSD malloc is technically only called during initialization, before the sandbox is entered, but these are called at runtime:

For malloc_freeunmap (U):

https://github.com/openbsd/src/blob/df69c215c7c66baf660f3f65414fd34796c96152/lib/libc/stdlib/malloc.c#L719-L720
https://github.com/openbsd/src/blob/df69c215c7c66baf660f3f65414fd34796c96152/lib/libc/stdlib/malloc.c#L783-L784
https://github.com/openbsd/src/blob/df69c215c7c66baf660f3f65414fd34796c96152/lib/libc/stdlib/malloc.c#L800-L801

For malloc(0):

https://github.com/openbsd/src/blob/df69c215c7c66baf660f3f65414fd34796c96152/lib/libc/stdlib/malloc.c#L899-L901
https://github.com/openbsd/src/blob/df69c215c7c66baf660f3f65414fd34796c96152/lib/libc/stdlib/malloc.c#L1112-L1113

For malloc_guard (G):

https://github.com/openbsd/src/blob/df69c215c7c66baf660f3f65414fd34796c96152/lib/libc/stdlib/malloc.c#L1149-L1154

The hardened_malloc implementation is *mostly* using mmap with MAP_FIXED to both purge and protect regions at the same time, but it uses `mprotect` to make regions readable/writable. It *could* technically use `mmap` with MAP_FIXED for all of these cases but `mprotect` makes more sense and is probably faster.

thestinger:
I think glibc malloc also uses mprotect, but it just isn't getting triggered by OpenSSH at the moment. It might only be used with threads.

brynet:
It's also worth clarifying that there is no portable version of OpenBSD malloc(3), and that people using it on platforms other than OpenBSD are kind of doing their own thing.

OpenSSH on OpenBSD would not be using the sandbox-seccomp-filter, but instead sandbox-pledge.

yegortimoshenko:
> It's also worth clarifying that there is no portable version of OpenBSD malloc(3), and that people using it on platforms other than OpenBSD are kind of doing their own thing.

There is a really old unofficial port, that said, you're right. However, it does make a point that other malloc implementations might also want to use `mprotect(2)` for the same reasons OpenBSD implementation does.

> However, like mmap(2), the PROT_EXEC flag isn't allowed by default.

Is that a W^X restriction? As far as I can tell, seccomp filter will currently pass through `PROT_EXEC` flag for `mmap(2)`:

https://github.com/openssh/openssh-portable/blob/b3a77b25e5f7880222b179431a74fad76d2cf60c/sandbox-seccomp-filter.c#L199-L201

In which case, passing through `PROT_EXEC` for `mprotect(2)` seems fair.

yegortimoshenko:
Looks like this has been addressed in https://github.com/openssh/openssh-portable/commit/f6906f9bf12c968debec3671bbf19926ff8a235b. Thank you!

djmdjm:
Yeah, I don't want to allow PROT_EXEC mappings as these seem useful in bootstrapping a memory fault exploit to full RCE.

I've added a SC_ALLOW_ARG_MASK() to allow permission of syscalls whose arguments match a particular mask. In this case mprotect() with only PROT_READ|PROT_WRITE|PROT_NONE is permitted.

brynet:
@djmdjm In which case, does it make sense to use SC_ALLOW_ARG_MASK()  for the __NR_mmap/__NR_mmap2 cases?

yegortimoshenko:
@djmdjm: Somewhat offtopic, would it make sense to limit `PROT_EXEC` for `mmap(2)` in that case?

djmdjm:
I was just working on that :)

adrelanos:
OpenSSH feature request: [test compatibility with hardened memory allocator Hardened Malloc](https://bugzilla.mindrot.org/show_bug.cgi?id=3245)

PR #143, opened 2019-08-24
Add PermitListen support for port 0

GuyLewin:
As part of a project I'm working on I want to limit an SSH user to only be able to create remote forwards with a randomly binded port (0).
Current implementation doesn't let the user permit port=0 on remote forwards, even though it's a valid input in remote forwards (not in local forwards).
With this PR the context of what flow is being passed into the port validation, taking it into account.


GuyLewin:
@djmdjm is there anything I should fix?

GuyLewin:
@djmdjm bump :)

PR #144, opened 2019-09-05
kex: avoid NULL deref upon kex_assemble_names misuse

emaste:
No functional change, kex_assemble_names is never called with NULL
listp.

Reported by Coverity against FreeBSD, CID 1395656


djmdjm:
I've committed something similar in 00865c2969000

djmdjm:
ugh, comment got manged but it had a link to this PR

emaste:
Thanks.

Also, another issue we spotted in Coverity already has a fix waiting in https://github.com/openssh/openssh-portable/pull/120

PR #150, opened 2019-10-06
Initial (pre-alpha) implementation of SSH w/lz4

eborisch:
This is not ready to be merged, but just thought I would put it out there to gauge interest.

Replaces zlib with lz4 as the compression used if available on both sides; otherwise backwards compatible.

Needs more work, just posting this up in case someone else is interested in helping out / has more experience with openssh internals.

Seems to function at this point. Textual input actually grows due to lots of tiny packets, and lz4 streaming properties. Need to consider what to do about this (send tiny packets uncompressed with just a size header? May not be worth the extra complexity since these very small packets will be dominated by other (network latency) timings anyway...

But this has very low (CPU) overhead compared to zlib for many reasonable compression on many streams. A quick test on my 2.3 GHz i5 is 4x faster when CPU bound.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1260,6 +1260,47 @@ AC_CHECK_FUNC([getspnam], ,
 AC_SEARCH_LIBS([basename], [gen], [AC_DEFINE([HAVE_BASENAME], [1],
 	[Define if you have the basename function.])])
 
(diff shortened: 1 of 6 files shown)


WSLUser:
Please add this upstream instead

PR #153, opened 2019-10-11
Switch .spec files to use %global instead of %define

nkadel:
The newer standard is to use %global wherever possible in .spec files.


nkadel:
Branched from wrong repo.

PR #154, opened 2019-10-12
Deny (non-fatal) ipc in preauth privsep child.

jeremyd2019:
As noted in openssh/openssh-portable#149, i386 does not have have
_NR_shmget etc.  Instead, it has a single ipc syscall (see man 2 ipc,
https://linux.die.net/man/2/ipc).  Add this syscall, if present, to the
list of syscalls that seccomp will deny non-fatally.


daztucker:
Applied to both master and the V_8_1 branch and will be in the next release.

Thanks.

PR #156, opened 2019-10-13
Switch %define to %global for redhat/openssh.spec

nkadel:
Simple, lightweight pull request to just address the obsolete use of '%define' in RHEL .spec files.


daztucker:
Applied thanks.

PR #160, opened 2019-11-04
U2f mnemonic

Jordandev678:
Update the U2F `-x` option to use a string argument instead of raw flag numbers. Also fixes an issue with "x" being defined twice in ```getopt```. The flag is named `no_up` to disable the user-presence requirement when signing.


djmdjm:
We have something similar now via the -O flag. Thanks

PR #161, opened 2020-01-08
seccomp: Allow clock_gettime64() in sandbox.

kraj:
This helps sshd accept connections on mips platforms with
upcoming glibc ( 2.31 )


daztucker:
Applied, to both master and the V_8_1 branch and will be in the next release.  Thanks.

jim0421:
@kaaj I'm confused why we need to allow clock_gettime64() to in sandbox when we just use openssh on x86 platform. But original openssh 8.0p1 will fail on glibc-2.31 without these change. Hoping for your response!

PR #162, opened 2020-01-14
Allow ssh(1) to connect through UNIX domain sockets.

pjd:
On systems where getaddrinfo(3) supports AF_LOCAL family allow ssh(1) to connect through UNIX domain sockets.

This is useful in combination with connections forwarding over UNIX domain
sockets.

Let's say I have system A behind a NAT and system B accessible in the Internet.
I want to enable some users of system B to connect to sshd on system A.
I can leverage stream forwarding and run on system A the following command:

ssh -f -N -R /tmp/ssh.sock:127.0.0.1:22 -l pjd B.example.org

Now, on the system B only I will be able to connect to system A over UNIX domain socket.
On system B with this patch I can run:

ssh -v /tmp/ssh.sock
[...]
debug1: Connecting to /tmp/ssh.sock [/tmp/ssh.sock].
debug1: Connection established.
[...]
debug1: Authenticating to /tmp/ssh.sock as 'root'
[...]
Authenticated to /tmp/ssh.sock.

% tail -1 ~/.ssh/known_hosts
/tmp/ssh.sock ecdsa-sha2-nistp256 [...]

### DIFF
--- a/ssh-sk.c
+++ b/ssh-sk.c
@@ -23,7 +23,9 @@
 
 #include <dlfcn.h>
 #include <stddef.h>
-#include <stdint.h>
+#ifdef HAVE_STDINT_H
+# include <stdint.h>
+#endif
 #include <string.h>
 #include <stdio.h>
 


cgf1:
I know I'm a couple of years late with this but I don't see how this patch could possibly work.  In the above example, /tmp/ssh.sock will be interpreted as a host name, which will fail with:
ssh: Could not resolve hostname /tmp/ssh.sock: Name or service not known

Is there a missing part of this patch?

arogozhnikov:
May I advocate for support of unix domain sockets in ssh as an alternative to using TCP ports? 

I found that management of different connections between docker containers and host machine is more convenient when it is organized using unix domain sockets, not TCP ports.

This way, multiple docker containers can co-exist on the same host and each expose a number of services. These services are accessible as unix domain sockets, not ports. For one thing, I'm less worried about security issues as now communication between them is basically IPC. For the second, and more important, there is no fight over ports and no need to memorize which port is what and from which container as now that's a part of filesystem where name conflicts are simply resolved by mounting volumes to different folders... that's so convenient. 

In total, that's a very strong case to have a broader support of unix domain sockets ... and most tools I use already have it, but ssh server and ssh client require additional steps with `socat` to make it work. Yes, it is possible with additional steps, but I believe more people would be into using unix sockets if there was a canonical way to deal with it in ssh.

In particular, I believe the following parts are missing:

1. serve ssh server on unix sockets 
2. connect with ssh client to unix socket on this machine (this PR)
3. connect with ssh client to unix socket on another machine. That's a combination of proxy connect to TCP port and previous point, but it needs some explicit way to encode this in CLI and ssh config. Potentially `ssh -j machine -v /socket` is the right way for CLI

jirib:
A use case could also be a kind of "hosting" where main-system `sshd` would forward users connections to the final user `sshd` which is systemd socket activated; that is system `sshd` would use `ForceCommand` to `ssh` to unix socket. (This way user `sshd` could run with `systemd.exec(5)` options as an alternative to having per user chroots in traditional way.)

rwmjones:
Was this actually fixed, and if so on what branch?

arogozhnikov:
doesn't seem to be fixed / implemented, PR likely was auto-closed due to force push by @pjd

rwmjones:
Oh that's a shame.  I might have a go at resubmitting this once I'm back at work because doing ssh over Unix domain sockets would be very useful to us.

oliverkurth:
The change seems to have disappeared (commits and "files changes" has no entries). Is it it available somewhere? I would like to experiment with it.

kalvdans:
@oliverkurth gitlab logged the hash before the force push, see https://github.com/pjd/openssh-portable/compare/429170f273ce1b0140f8111a45ba69390d98de3a...cfc423f06cd6cc0dac9721f9d5a20c59b92021be

oliverkurth:
> @oliverkurth gitlab logged the hash before the force push, see [pjd/openssh-portable@429170f...cfc423f](https://github.com/pjd/openssh-portable/compare/429170f273ce1b0140f8111a45ba69390d98de3a...cfc423f06cd6cc0dac9721f9d5a20c59b92021be)

Thank you! I found it meanwhile also here on @pjd's fork: https://github.com/openssh/openssh-portable/compare/master...pjd:openssh-portable:unix_socket

kalvdans:
> Oh that's a shame. I might have a go at resubmitting this once I'm back at work because doing ssh over Unix domain sockets would be very useful to us.

I resubmitted it now, see: https://github.com/openssh/openssh-portable/pull/431

PR #166, opened 2020-01-20
Fix missing prototype warning

rubenk:
This function is only used in this file, and only on Cygwin,
so make it static and hide it behind HAVE_CYGWIN

Fixes:
session.c:978:1: warning: no previous prototype for function 'copy_environment' [-Wmissing-prototypes]
copy_environment(char **source, char ***env, u_int *envsize)


daztucker:
Applied, thanks.

PR #168, opened 2020-01-20
auth-pam.c: fix missing prototype

rubenk:
The function is only used in this file, so make it static.
Fixes:
1 warning generated.
auth-pam.c:304:1: warning: no previous prototype for function 'sshpam_password_change_required' [-Wmissing-prototypes]
sshpam_password_change_required(int reqd)
^
1 warning generated.


daztucker:
Applied, thanks.

PR #169, opened 2020-01-20
configure.ac: fix ldns test

rubenk:
When running ./configure --with-ldns, if ldns-config cannot be found, we
add -Iyes/include to CPPFLAGS and -Lyes/lib to LDFLAGS. Fix that.


daztucker:
Applied, thanks.

PR #172, opened 2020-02-20
Fix some typos and textual mistake

mkontani:
I read docs and found some typos and textual mistake.


daztucker:
Applied upstream to openbsd or here to portable as appropriate.  Thanks.

PR #173, opened 2020-02-23
Fix typo in ssh-keygen.1, s/expirty/expiry/

mkontani:
SSIA


daztucker:
applied upstream, thanks.

PR #178, opened 2020-04-19
Fix infinity loop

eagleoflqj:
i is size_t (used to be ssize_t), which makes the original comparisons always true.


daztucker:
Thanks.  We've made a slightly different change that's a bit more idiomatic but should also fix it.

PR #181, opened 2020-04-28
sshd.8: fix marker (singular) known_hosts field description

scop:


djmdjm:
Thanks - this has been merged

PR #184, opened 2020-05-07
Refresh primary group ID after successful authentication.

hpjansson:
In certain environments (e.g. using winbind), getpwnam() will return a placeholder primary group ID until the user has been successfully authenticated. This causes sshd to set the wrong GID for the session. As a workaround, we refresh this value with an additional call to getpwnam() after a successful authentication.

For a longer explanation of why this is needed, see [the Samba 4.6 release notes, winbind section](https://wiki.samba.org/index.php/Samba_4.6_Features_added/changed#winbind_changes).

### DIFF
--- a/sshd.c
+++ b/sshd.c
@@ -1502,6 +1502,7 @@ main(int ac, char **av)
 	int keytype;
 	Authctxt *authctxt;
 	struct connection_info *connection_info = NULL;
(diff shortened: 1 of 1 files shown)


hpjansson:
I'm withdrawing this since it's been several years without any response from the dev list.

PR #185, opened 2020-05-18
indentation

cyrilst:
This is only cosmetic. Some lines have non-coherent indentation.


djmdjm:
Thanks, I fixed these and others upstream. Merged as 31d8d231e with additional portable ones in 57ed647ee

PR #188, opened 2020-05-29
Added option RSAMinimumModulusSize <int> to lower the (now) hard limit

sinihappo:
Added option RSAMinimumModulusSize <int> to lower the (now) hard limit
SSH_RSA_MINIMUM_MODULUS_SIZE when necessary.  Although it is usually
possible and extremely advisable to generate new longer server keys,
there are older devices where this is not possible.  It is, however,
practical still use SSH with them even if it is not as secure as it
should be.  It is still safer to use SSH with 800 bit keys that telnet
without any encryption.

Background:

I have struggled with older network gear, where either it is not possible because of the lack of new FW or lack of permit to upgrade.  If you think that having this option needs more safeguards, please give ideas on what kind of extra checks or options or anything.


ssh_config.5
```diff
@@ -1486,6 +1486,12 @@ an OpenSSH Key Revocation List (KRL) as generated by
 .Xr ssh-keygen 1 .
 For more information on KRLs, see the KEY REVOCATION LISTS section in
 .Xr ssh-keygen 1 .
+.It Cm RSAMinimumModulusSize
+Specifies the minimum accepted RSA modulus size in different
+contexts.  This is only be used with older SSH servers when it is
```
mirabilos, line 1491: Line break after a full stop, so the spacing (american or proper) is applied automatically. See: https://man.openbsd.org/roff.7#Sentence_Spacing

ssh_config.5
```diff
@@ -1486,6 +1486,12 @@ an OpenSSH Key Revocation List (KRL) as generated by
...
+impossible to have or generate longer keys for them.  This should not
+be used in any other purposes except perhaps testing.  There is still
+hard limit, 512.  To use shorter RSA keys than that, OpenSSH must be recompiled.
```
mirabilos, line 1494: `s/hard/a &/`

sinihappo:
Thank you, I agree.

How we could get the patch merged to the repository?  As I have stated, there is
a need for this.  And I tried to implement it in such a way that it would be
used only when needed.

    /alo

mirabilos wrote:

mirabilos:
sinihappo dixit:

>How we could get the patch merged to the repository?  As I have stated, there is
>a need for this.  And I tried to implement it in such a way that it would be
>used only when needed.

Sorry, not an OpenSSH developer :/

djmdjm:
Sorry, we're not interested in enabling support for unsafe crypto. If you need to connect to such devices, then please compile your own ssh with SSH_RSA_MINIMUM_MODULUS_SIZE redefined.

PR #189, opened 2020-06-03
auth_log: dont log partial successes as failures

Feandil:
Before this patch, when auth_log is called with `authenticated=0` and `partial=1` without `authctxt->postponed` being set (which is normal on partial authentications) then:
- if method is `password`, `keyboard-interactive` or `challenge-response`, then `record_failed_login` is called
- `audit_event` is called with an event from `audit_classify_auth` which always returns a failure SSh event (or unknown).

However partial authentications are not failures! Looking a bit into history, it seems that the custom logs part of `auth_log` (within `ifdef`) was missed out when the `partial` parameter was added to `auth_log` (15b05cfa17592da7470d7bd4b2de063188697471 in openssh-portable).

The patch is modifying a bit the existing `ifdef` logic to remove the duplicated `if (authenticated == 0 && !authctxt->postponed`, which I understand was flawed (missing the partial check). I'm not sure if it's better, because it makes it a bit more complicated to follow all cases (depending on the defined variables), but at the same time it simplifies theif statements. Let me know what's best.

I'm not too sure why the `if (authenticated)` part for `WITH_AIXAUTHENTICATE`is within `CUSTOM_FAILED_LOGIN`, as it's not logging a failed login, but I didn't want to change this logic...

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -352,23 +352,26 @@ auth_log(struct ssh *ssh, int authenticated, int partial,
 
 	free(extra);
 
-#ifdef CUSTOM_FAILED_LOGIN
-	if (authenticated == 0 && !authctxt->postponed &&
-	    (strcmp(method, "password") == 0 ||
-	    strncmp(method, "keyboard-interactive", 20) == 0 ||
-	    strcmp(method, "challenge-response") == 0))
-		record_failed_login(ssh, authctxt->user,
-		    auth_get_canonical_hostname(ssh, options.use_dns), "ssh");
-# ifdef WITH_AIXAUTHENTICATE
+#if defined(CUSTOM_FAILED_LOGIN) || defined(SSH_AUDIT_EVENTS)
+	if (authenticated == 0 && !(authctxt->postponed || partial)) {
+		/* Log failed login attempt */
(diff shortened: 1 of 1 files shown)


daztucker:
> I'm not too sure why the if (authenticated) part for WITH_AIXAUTHENTICATE is within CUSTOM_FAILED_LOGIN

It looks like that was added by me in 26d4e19caa3013f57dc3c1462847eceaac6a1d7d.  CUSTOM_FAILED_LOGIN was an attempt to somewhat standardize handling of platforms that did their own thing for bad passwords (97363a8b24601bad631f6f187c487a166f7eb959) although the only other user of it (UNICOS) has since been removed.  WITH_AIXAUTHENTICATE always sets CUSTOM_FAILED_LOGIN (in port-aix.h) so that could be moved outside the ifdef and it'd still work.

Looks like the whole thing could use some cleanup, although it'll require a bit of thought.

daztucker:
the closing curly was inside SSH_AUDIT_EVENTS so it'd only compile if that was defined, otherwise looked ok to both me and djm so applied with that minor change.

thanks.

PR #190, opened 2020-06-05
skip ensure_minimum_time_since if use pam

FirstLoveLife:
some pam modules will set fail delay, e.g. ,pam_tally*, pam_faillock.
When ensure_minimum_time_since is invoked, these delays are not
guaranteed to be correct.

### DIFF
--- a/auth2.c
+++ b/auth2.c
@@ -336,7 +336,7 @@ input_userauth_request(int type, u_int32_t seq, struct ssh *ssh)
 		debug2("input_userauth_request: try method %s", method);
 		authenticated =	m->userauth(ssh);
 	}
-	if (!authctxt->authenticated)
+	if (!authctxt->authenticated && !options.use_pam)
 		ensure_minimum_time_since(tstart,
 		    user_specific_delay(authctxt->user));
 	userauth_finish(ssh, authenticated, method, NULL);


daztucker:
Why is ensuring a minimum time a problem?  I'd expect if something set pam_fail_delay the code in sshd just wouldn't do anything since the minimum time should have already passed.

FirstLoveLife:
> Why is ensuring a minimum time a problem? I'd expect if something set pam_fail_delay the code in sshd just wouldn't do anything since the minimum time should have already passed.

Sorry for my late reply. I have a minimum demo to demostrate:

```
uos@uos-PC:~$ sudo journalctl -b 0 | egrep "ensure_minimum_time_since|pam_faildelay"
6月 08 14:37:06 uos-PC sshd[14903]: pam_faildelay(sshd:auth): setting fail delay to 13000000
6月 08 14:37:09 uos-PC sshd[14899]: debug3: ensure_minimum_time_since: elapsed 13329.696ms, delaying 9770.912ms (requested 5.640ms) [preauth]
6月 08 14:37:19 uos-PC sshd[14903]: debug3: ensure_minimum_time_since: elapsed 13590.364ms, delaying 9510.244ms (requested 5.640ms) [preauth]
uos@uos-PC:~$ grep -i debug /etc/ssh/sshd_config 
LogLevel DEBUG3
uos@uos-PC:~$ head -2 /etc/pam.d/sshd 
# PAM configuration for the Secure Shell service
auth  optional  pam_faildelay.so  delay=13000000 debug
```

`OpenSSH_7.9p1 Deepin-10, OpenSSL 1.1.1c  28 May 2019`


As u can see, I have set faildelay to `13s`, but `ensure_minimum_time_since` still try to increase it to nearly `24s`, which is my actual waiting time.

FirstLoveLife:
@daztucker Hi, is there other question?

djmdjm:
What if the user doesn't have faildelay enabled? sshd has no way of knowing...

FirstLoveLife:
@djmdjm Sorry for late reply. IMO no matter whether faildelay is used, once pam is used, the minimum time should be decided by pam, instead sshd itself.

PR #191, opened 2020-06-12
Fix RFC number in PROTOCOL.sshsig

haruyama:
s/RFC4243/RFC4253/


djmdjm:
Thanks! Applied upstream and imported as 7fafaeb5

PR #192, opened 2020-06-29
parse_user_host_path, empty path should be allowed

kircherlike:
when parse_user_host_path, empty path should be allowed

when using "sftp root@[2001::16%eth0]", the error output:
ssh: Could not resolve hostname [2001::16%eth0]: Name
or service not known
Connection closed.
Connection closed

fix sftp ipv6 login failed accidental like this:
File "/root/!" not found.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -681,6 +681,8 @@ colon(char *cp)
 			flag = 1;
 		if (*cp == ']' && *(cp+1) == ':' && flag)
 			return (cp+1);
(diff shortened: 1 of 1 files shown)


kircherlike:
@djmdjm

djmdjm:
I can't replicate this problem - `sftp 'djm@[fe80::1%lo0]'` works as expected on Linux and OpenBSD

PR #194, opened 2020-07-15
Use SFTP internally for SCP command

Jakuje:
Based on the discussion on mailing list [1], this is initial implementation of SCP mode using SFTP. Comments, suggestions, reviews, feedback welcomed.

[1] https://lists.mindrot.org/pipermail/openssh-unix-dev/2020-June/038594.html

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -399,6 +399,7 @@ int Tflag, pflag, iamremote, iamrecursive, targetshouldbedirectory;
 char cmd[CMDNEEDS];		/* must hold "rcp -r -p -d\0" */
 
 enum scp_mode_e {
+	MODE_DEFAULT,
 	MODE_SCP,
 	MODE_SFTP
 };
@@ -423,7 +424,7 @@ main(int argc, char **argv)
 	extern char *optarg;
 	extern int optind;
 	/* For now, keep SCP as default */
-	enum scp_mode_e mode = MODE_SCP;
+	enum scp_mode_e mode = MODE_DEFAULT;
 	char *sftp_direct = NULL;
 
 	/* Ensure that fds 0, 1 and 2 are open or directed to /dev/null */
@@ -559,6 +560,10 @@ main(int argc, char **argv)
 	/* Do this last because we want the user to be able to override it */
 	addargs(&args, "-oForwardAgent=no");
 
+	/* The server can not work in SFTP mode so keep default SCP there */
+	if (mode == MODE_DEFAULT) {
+		mode = iamremote ? MODE_SCP : MODE_SFTP;
+	}
 	if (mode != MODE_SFTP && sftp_direct != NULL) {
 		fatal("SFTP direct can be used only in SFTP mode");
 	}
--- a/scp.c
+++ b/scp.c
@@ -123,11 +123,15 @@
 #include "progressmeter.h"
 #include "utf8.h"
 
+#include "sftp-common.h"
+#include "sftp-client.h"
+
 extern char *__progname;
 
 #define COPY_BUFLEN	16384
 
-int do_cmd(char *host, char *remuser, int port, char *cmd, int *fdin, int *fdout);
+int do_cmd(char *program, char *host, char *remuser, int port, char *cmd,
+    int *fdin, int *fdout);
 int do_cmd2(char *host, char *remuser, int port, char *cmd, int fdin, int fdout);
 
 /* Struct for addargs */
@@ -161,6 +165,12 @@ char *ssh_program = _PATH_SSH_PROGRAM;
 
 /* This is used to store the pid of ssh_program */
 pid_t do_cmd_pid = -1;
+
+/* Needed for sftp */
+volatile sig_atomic_t interrupted = 0;
+
+int remote_glob(struct sftp_conn *, const char *, int,
+    int (*)(const char *, int), glob_t *); /* proto for sftp-glob.c */
 
 static void
 killchild(int signo)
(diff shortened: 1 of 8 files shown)


sftp-client.c
```diff
@@ -1956,3 +1956,52 @@ path_append(const char *p1, const char *p2)
...
+	if ((a = do_stat(conn, path, 1)) == NULL)
+		return(0);
+	if (!(a->flags & SSH2_FILEXFER_ATTR_PERMISSIONS))
+		return(0);
+	return(S_ISDIR(a->perm));
+}
+
+
+int
+is_dir(const char *path)
```
djmdjm, line 1988: since this will be (more) public, it would probably better as local_is_dir() for clarity

sftp-client.h
```diff
@@ -142,4 +142,17 @@ int upload_dir(struct sftp_conn *, const char *, const char *, int, int, int,
...
+char *make_absolute(char *, const char *);
+
+/* Check if remote path is directory */
+int remote_is_dir(struct sftp_conn *conn, const char *path);
+
+/* Check if local path is directory */
+int is_dir(const char *path);
+
+/* Check whether path returned from glob(..., GLOB_MARK, ...) is a directory */
+int pathname_is_dir(const char *pathname);
```
djmdjm, line 156: likewise, s/pathname_is_dir/globpath_is_dir/

scp.c
```diff
@@ -976,19 +1044,37 @@ toremote(int argc, char **argv)
 			}
 			addargs(&alist, "--");
 			addargs(&alist, "%s", host);
-			addargs(&alist, "%s", cmd);
+			/* This will work only if the first remote scp supports sftp mode */
+			addargs(&alist, "%s%s", cmd, mode == MODE_SFTP ? " -M sftp" : "");
```
djmdjm, line 1048: Will this work? I thought it would treat "cmd -M sftp" as a single argument, though scp's horrible argument passing might avoid that.

Unless there's some reason to do it on a single line, an explicit "if (mode == MODE_SFTP) addargs(...)" would be better IMO

scp.c
```diff
@@ -1225,6 +1379,95 @@ rsource(char *name, struct stat *statp)
...
+
+	if (!getcwd(pwd, sizeof(pwd))) {
+		error("Couldn't get local cwd: %s", strerror(errno));
+		err = -1;
+		goto out;
+	}
+
+	remote_path = do_realpath(conn, ".");
+	if (remote_path == NULL) {
+		error("Need cwd");
```
djmdjm, line 1403: maybe /* TODO - gracefully degrade by using relative paths */ or somesuch here

scp.c
```diff
@@ -1225,6 +1379,95 @@ rsource(char *name, struct stat *statp)
...
+		abs_dst = NULL;
+	}
+
+out:
+	free(abs_src);
+	globfree(&g);
+	if (err == -1) {
+		fatal("Failed to download file '%s'", src);
+	}
+	//return(err);
```
djmdjm, line 1467: remove C++ comment

scp.c
Jakuje: Thank you. Indeed you are right. The arguments should come separately. Fixed now.

scp.c
Jakuje: added

Jakuje:
Thank you for the comments. I addressed your suggestions in the current version.

I will be able to follow-up no earlier than in August on the fallback to scp (thanks for suggesting the mux -- it could actually work that way).

mfriedl:
I had a quick first look and I think this looks great!
It also looks like we could put this in soon and keep defaulting to the scp protocol for a while, but
get people to use it and figure out problems.
I'll have a detailed look at the code soon!
thanks!

Jakuje:
Updated to resolve conflicts.

I tried also to play with the fallback to scp from sftp, but the current code in sftp-client is really not ready for this as it calls fatal() in `send_msg()` from `do_init()`, which is the place we should just figure out that sftp subsystem request failed so it would require some more tweaking. Putting this aside for now.

Jakuje:
I rebased on current master and change default to SFTP to give it more testing. Any update from your side?

mfriedl:
I'd like to work on this if spare time permits.

scp.c
```diff
@@ -526,6 +560,18 @@ main(int argc, char **argv)
...
+	/* The server can not work in SFTP mode so keep default SCP there */
+	if (mode == MODE_DEFAULT) {
+		mode = iamremote ? MODE_SCP : MODE_SFTP;
+	}
+	if (mode != MODE_SFTP && sftp_direct != NULL) {
+		fatal("SFTP direct can be used only in SFTP mode");
+	}
+
+	if (mode == MODE_SFTP && iamremote) {
+		fatal("The server can not be ran in SFTP mode");
```
suominen, line 572: s/ran/run/ (grammar)

God-knows-all:
Please fix it for me. I don’t undetected properly how to fix it? So please
fix it for me.

God-knows-all (approved):

God-knows-all (approved):

DemiMarie:
Can we please make quoting bug-compatibility optional?  I really really would like to be able to avoid it in scripts.

djmdjm:
All but the switch-default change have been committed (in af5d8094d and 197e29f1c).

My plan is to leave the protocol as scp for at least one release and then look at switching.

Thanks Jakub for writing this!

PR #195, opened 2020-07-30
Fix delay in case of ssh's ConnectTimeout option usage

kuhanov:
For example, ConnectionTimeout=5 produces 5 sec delay in any case.
ssh -o BatchMode=yes -o ServerAliveInterval=5 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=5 user@target uname
In real connection takes about 200ms


kuhanov:
don't apply changes

PR #196, opened 2020-08-05
Add forgotten free

dns2utf8:
Hi all

This is my first PR here.
**EDIT:** I overlooked the `autoreconf` step, it did build.

Cheers,
Stefan


dns2utf8:
I overlooked that there is a check for the variant where the argument is not NULL, so this PR is not needed.

Thank you anyways

PR #197, opened 2020-08-05
Pass the -4 and -6 flags on with -J

dns2utf8:
Hi all

I had the problem that I could not enforce the `-6` flag for a dual stack target that had broken IPv4.
This patch fixes that by passing the flag on if present.

I hope the formatting is okay. If not, I am happy to reformat it.

Cheers,
Stefan

### DIFF
--- a/ssh.c
+++ b/ssh.c
@@ -1265,6 +1265,7 @@ main(int ac, char **av)
 	if (options.jump_host != NULL) {
 		char port_s[8];
 		const char *jumpuser = options.jump_user, *sshbin = argv0;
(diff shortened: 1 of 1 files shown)


djmdjm:
Hi, we've deliberately chosen not to do this because it precludes the possibility of forcing IPv4 on the first connection of a jump while leaving the proxied connection itself free to choose the address family. It's a little more hassle, but I'd recommend using the AddressFamily keyword in a couple of ~/.ssh/config stanzas to achieve the same effect.

God-knows-all:
Yes

PR #200, opened 2020-09-01
configure.ac: fix libfido2 back-compat

martelletto:
- HAVE_FIDO_CRED_PROD -> HAVE_FIDO_CRED_PROT;
- check for fido_dev_get_touch_begin(), so that HAVE_FIDO_DEV_GET_TOUCH_BEGIN gets defined.


daztucker:
applied, thanks.

PR #201, opened 2020-09-13
Fix serverside SetEnv inside Match blocks

manuelm:
I noticed that serverside `SetEnv` keywords inside `Match` blocks isn't working. Looking at the commit history (28013759f09ed3ebf7e8335e83a62936bd7a7f47) it looks like adding the array to copy macro simply has been forgotten.

Further more it seems multiple statements inside the same context are not supported for no apparent reason. So I removed that as well.

### DIFF
--- a/servconf.c
+++ b/servconf.c
@@ -1945,12 +1945,11 @@ process_server_config_line_depth(ServerOptions *options, char *line,
 		break;
 
 	case sSetEnv:
-		uvalue = options->num_setenv;
 		while ((arg = strdelimw(&cp)) && *arg != '\0') {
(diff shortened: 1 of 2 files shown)


djmdjm:
Thanks, I have merged the first commit (via our OpenBSD upstream) but not the second.

I think ignoring previous successful setenv causes Match block setenv to be appended to ones set in the global configuration rather than the desired behaviour of overriding them. I need to think about it a little more....

djmdjm:
It turns out that the first patch alone is sufficient to fix SetEnv inside match, as parse_server_match_config() always works on a new, empty ServerOptions object.

Thanks!

PR #202, opened 2020-09-14
change type of nid to int

thingsconnected:
This is in line with the return type of sshkey_ecdsa_nid_from_name()
and the argument type of sshkey_curve_nid_to_bits(int nid).

When casting to u_int, the possible -1 error return value of
sshkey_ecdsa_nid_from_name() will go unchecked.


djmdjm:
merged via OpenBSD as b12b835dc022b - thanks!

PR #206, opened 2020-10-01
Fix `EOF: command not found` error in ssh-copy-id

Fallmay:
At the moment, the ssh-copy-id utility is not working properly.

```sh
$ ssh-copy-id user@hostname
/usr/bin/ssh-copy-id: INFO: attempting to log in with the new key(s), to filter out any that are already installed
/usr/bin/ssh-copy-id: INFO: 2 key(s) remain to be installed -- if you are prompted now it is to install the new keys
/usr/bin/ssh-copy-id: line 251: warning: here-document at line 251 delimited by end-of-file (wanted `EOF')
/usr/bin/ssh-copy-id: line 250: warning: here-document at line 250 delimited by end-of-file (wanted `EOF')
/usr/bin/ssh-copy-id: line 254: /dev/null`: Permission denied
/usr/bin/ssh-copy-id: line 260: EOF: command not found
```

OpenSSH 8.4/8.4p1 (2020-09-27).

### DIFF
--- a/contrib/ssh-copy-id
+++ b/contrib/ssh-copy-id
@@ -247,7 +247,7 @@ installkeys_sh() {
   #    the -z `tail ...` checks for a trailing newline. The echo adds one if was missing
   #    the cat adds the keys we're getting via STDIN
   #    and if available restorecon is used to restore the SELinux context
-  INSTALLKEYS_SH=$(tr '\t\n' ' ' <<-EOF)
+  INSTALLKEYS_SH=$(tr '\t\n' ' ' <<-EOF
 	cd;
 	umask 077;
 	mkdir -p $(dirname "${AUTH_KEY_FILE}") &&
@@ -258,6 +258,7 @@ installkeys_sh() {
 	  restorecon -F .ssh ${AUTH_KEY_FILE};
 	fi
 EOF
+  )
 
   # to defend against quirky remote shells: use 'exec sh -c' to get POSIX;
   printf "exec sh -c '%s'" "${INSTALLKEYS_SH}"


3eka:
Hi,
this might depend on OS (just guessing). 
On Linux /bin/sh is sym-linked to /bin/bash which is not case everywhere.
For example, I have to change /bin/sh to /bin/bash on OpenIndiana to get ssh-copy-id working (no errors; last one from 8.4p1).

Regards.

daztucker:
ssh-copy-id is contributed by Phil Hands, its upstream is at https://gitlab.com/phil_hands/ssh-copy-id/-/tree/main

phil-hands:
Which shell/OS was it that threw this error?

I'm asking becuase it occurs to me that it _may_ be better to detect that shell in a way similar to the initial 'if false " ...' test for the old Solaris shell, and then hunt for something better.

However, I suspect that the suggested patch is pretty portable, so probably easier to go with that.

BTW /bin/sh is not always a symlink to /bin/bash on Linux -- that is dependant on the distribution.  Debian uses  dash as the default /bin/sh, for instance.

daztucker:
Aside:
> /usr/bin/ssh-copy-id: line 254: /dev/null`: Permission denied
if your /dev/null is not world writable (or isn't a character special file) then many things are going to be sad.

Jakuje:
> Aside:
> 
> > /usr/bin/ssh-copy-id: line 254: /dev/null`: Permission denied
> 
> if your /dev/null is not world writable (or isn't a character special file) then many things are going to be sad.

I don't think the `/dev/null` is not writable. I think this is an after-effect of wrong parsing of the script and in this attempt, it tries to *execute* the `/dev/null` as a program, which obviously on any sane system fails.

Fallmay:
> Какая оболочка / ОС вызвала эту ошибку?
> 
> Я спрашиваю, потому что мне пришло в голову, что _может_ быть лучше обнаружить эту оболочку способом, аналогичным первоначальному тесту «если ложь» ... »для старой оболочки Solaris, а затем искать что-нибудь получше.
> 
> Однако я подозреваю, что предлагаемый патч довольно переносим, ​​поэтому, вероятно, с ним проще работать.
> 
> Кстати, / bin / sh не всегда является символической ссылкой на / bin / bash в Linux - это зависит от дистрибутива. Например, Debian использует тире как / bin / sh по умолчанию.

Fedora 33 (Workstation Edition Prerelease).

Fallmay:
> Which shell/OS was it that threw this error?
> 
> I'm asking becuase it occurs to me that it _may_ be better to detect that shell in a way similar to the initial 'if false " ...' test for the old Solaris shell, and then hunt for something better.
> 
> However, I suspect that the suggested patch is pretty portable, so probably easier to go with that.
> 
> BTW /bin/sh is not always a symlink to /bin/bash on Linux -- that is dependant on the distribution. Debian uses dash as the default /bin/sh, for instance.

Fedora 33 (Workstation Edition Prerelease).

Jakuje:
Maybe more important what shell is your default and what is your `/usr/bin/sh` symlinked to.

Fallmay:
> Maybe more important what shell is your default and what is your `/usr/bin/sh` symlinked to.

```
$ echo "$SHELL"
/bin/bash
$ ls -l /bin/sh /usr/bin/sh
lrwxrwxrwx. 1 root root 4 Jul 27 16:18 /bin/sh -> bash
lrwxrwxrwx. 1 root root 4 Jul 27 16:18 /usr/bin/sh -> bash
```

Jakuje:
Oh, right, I see it also with default `bash` in my test, but I somehow managed to miss this issue.

georgmu:
I don't understand how this has to do with bash vs. sh. This is an obvious shell error and I want to see any shell which interprets this correct.

To reduce the problem, execute the following command:
```
echo $(cat <<-EOF)
```

It is wrong to close the subshell before the EOF end marker

georgmu:
Ok, I tries it with dash and there it works.

bash accepts
```
(cat << EOF)
foo
EOF
```

but not 

```
FOO=(cat <<EOF)
foo
EOF
```

dash accepts both.

phil-hands:
Ah, righto -- It seems that I've somehow managed to do _all_ of my testing with dash.  That was careless.

I'll add a few more shells to my tests, and then fix this as suggested.

Thanks for the report :-)

phil-hands:
This is now included in my repo:

  https://gitlab.com/phil_hands/ssh-copy-id/-/tree/main

so the pull request here is no longer needed.

Thanks again to you all for helping make ssh-copy-id better.

daztucker:
I have pulled this change (and a couple of subsequent ones) from Phil's repository.  Thanks all.

gustavorv86:
It's perfect!

```
/usr/bin/ssh-copy-id: INFO: Source of key(s) to be installed: "/home/gustavo/.ssh/id_rsa.pub"
/usr/bin/ssh-copy-id: INFO: attempting to log in with the new key(s), to filter out any that are already installed
/usr/bin/ssh-copy-id: INFO: 1 key(s) remain to be installed -- if you are prompted now it is to install the new keys
/usr/bin/ssh-copy-id: line 251: warning: here-document at line 251 delimited by end-of-file (wanted `EOF')
/usr/bin/ssh-copy-id: line 250: warning: here-document at line 250 delimited by end-of-file (wanted `EOF')
/usr/bin/ssh-copy-id: line 260: EOF: command not found
```

I fixed the problem in Cygwin with openssh 8.4p1-1 package.

Thank you!

PR #208, opened 2020-10-10
Portability: use `uname -n` instead of `hostname`.

ghen2:
hostname is not defined by POSIX and not present by default on some modern Linux distro's, eg. Arch or Fedora.

### DIFF
--- a/regress/percent.sh
+++ b/regress/percent.sh
@@ -10,8 +10,8 @@ fi
 
 USER=`id -u -n`
 USERID=`id -u`
-HOST=`hostname | cut -f1 -d.`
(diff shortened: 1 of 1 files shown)


geert-hendrickx-be:
Not sure what caused the automatic checks to be cancelled, it seems unrelated to my change, rather some timeout in the check infrastructure?

`make tests` still runs succesfully here, particularly `percent.sh`.

daztucker:
Applied upstream, thanks.

I am wondering if anything has hostname but not uname -n, but that wasn't true of any of the oddball systems I checked so if any show up we can deal with them at that time.

PR #210, opened 2020-10-19
Correct and clarify no-touch-required option for CAs and certificates.

JackOfMostTrades:
Based on [source](https://github.com/openssh/openssh-portable/blob/master/auth-options.c#L98) and testing, this certificate extension should be "no-touch-required". Also based on some testing, the certificate also has to have been loaded (in an authorized_keys files) with the no-touch-required option for this extension to be respected (otherwise all signatures will still require touch even if this extension is included in the certificate).

### DIFF
--- a/PROTOCOL.certkeys
+++ b/PROTOCOL.certkeys
@@ -280,12 +280,15 @@ their data fields are:
 
 Name                    Format        Description
 -----------------------------------------------------------------------------
-no-presence-required    empty         Flag indicating that signatures made
+no-touch-required       empty         Flag indicating that signatures made
                                       with this certificate need not assert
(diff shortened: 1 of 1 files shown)


djmdjm:
Thanks for pointing out the incorrect option name - I committed a fix for this in 1879ff4062

However, this file isn't the right place to sshd discussion configuration options. This is a description of the certificate contents intended for other SSH protocol implementors and not end users.

Instead I'll look at adding an example to the the sshd(8) authorized_keys example section

djmdjm:
c298c4da574 adjusts the sshd(8) authorized_keys examples to include a CA that sets no-touch-required

PR #212, opened 2020-11-08
CVE-2020-14871 remote attack mitigation

minceheid:
CVE-2020-14871 in Solaris is a buffer-overflow vuln in PAM. Oracle's SSH and OpenSSH pass the username as-is to the Solaris PAM, and can trigger a remote exploit as there is insufficient bounds checking in PAM.

While patches are available for Solaris 10 and Solaris 11, for those circumstances where an organisation is not entitled to Oracle Premier Support, mitigation in SSHD would help reduce the likelihood of a remote exploit, whilst acknowledging that the PAM module is still broken, and local exploits may still be possible.

Setting MAXUSERLEN to zero in the updated auth2.c results in default OpenSSH sshd behaviour, but setting it to a nonzero value will truncate any long usernames to the length specified. The Solaris PAM issue is with a 512 byte buffer, so any value of MAXUSERLEN less than 512 should prevent sshd from passing a string capable of breaking PAM.

### DIFF
--- a/auth2.c
+++ b/auth2.c
@@ -280,6 +280,17 @@ input_userauth_request(int type, u_int32_t seq, struct ssh *ssh)
 	if ((style = strchr(user, ':')) != NULL)
 		*style++ = 0;
 
+	/* MMS - CVE-2020-14781 Remote Exploit Mitigation */
+	/* If MAXUSERLEN is defined to be >0, then it introduces a maximum length
+	CVE-2020-14781 is a PAM buffer overflow vulnerability of a 512 byte buffer,
+	and so, setting 0<MAXUSERLEN<512 will shield PAM from inbound SSH attacks,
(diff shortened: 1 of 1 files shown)


daztucker:
Thanks.  I committed a variant of this in fcf429a4c69d30d8725612a55b37181594da8ddf.
Damien found the offending PAM code (https://github.com/joyent/illumos-joyent/blob/edd669a7ce20a2f7406e8f00489c426c0690f1bd/usr/src/lib/libpam/pam_framework.c#L622) and the size of the buffer in question is exposed in the PAM header, so we always enable the mitigation for Sun-derived PAM stacks with the limit specified in the PAM header, terminate (and log) the connection if found, and with the code in auth-pam.c with the rest of the PAM code.

PR #214, opened 2020-11-17
ssh-keygen.1 - State that the KDF algo is bcrypt

rafork:
The ssh-keygen.1 manpage now states that the default number of KDF rounds is 16, but it doesn't state what KDF algorithm is used. It's important to know the KDF algorithm when choosing the number of rounds. This isn't obvious (except perhaps to openbsd users?). This commit adds an explicit statement to the -a option section that the KDF function is bcrypt.


daztucker:
added but in a slightly differently way.  thanks.

PR #215, opened 2020-11-17
ssh.1 Fix -i (mention .pub files if key in agent)

rafork:
The -i option section in ssh.1 is incomplete. It states that the -i option argument is the private key file. This is one of two possibilities. If the private key only exists in ssh-agent, and not locally on disk, then the -i option argument needs to be the corresponding public key file instead. The reader is given no clue that this is a possibility. This commit adds a short sentence to make this clear.

### DIFF
--- a/ssh.1
+++ b/ssh.1
@@ -297,6 +297,12 @@ It is possible to have multiple
 .Fl i
 options (and multiple identities specified in
 configuration files).
+If the private key is in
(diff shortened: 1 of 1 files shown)


daztucker:
A variant of this was committed upstream (2282e83b0ff78d2efbe705883b67240745fa5bb2).  Thanks.

PR #216, opened 2020-11-20
Improve sending struct passwd

yuichiro-naito:
This pull request avoids sending pointer values in struct passwd.
It is possible to be a wild pointer in peer process, if a new pointer member is added.
And I believe that one ToDo comment will be finished.

Background:

I made a patch for FreeBSD's openssh. (c.f. [my FreeBSD patch](https://reviews.freebsd.org/D17128) to see what I fixed)
We discussed about pointer values should not be sent to another process.
It is necessary for openssh and I think the ToDo comment says as same as well.


djmdjm:
This was fixed in b2bcec13f

PR #217, opened 2020-11-21
Using explicit_memset for the explicit_bzero compatibility layer.

devnexen:
Favoriting the native implementation in this case.


daztucker:
Applied, thanks.

PR #218, opened 2020-12-08
Fix the hostkeys protocol documentation

dbussink:
The documentation was lacking the needed want-reply field in the initial global request. This was found implementing hostkeys support in another SSH implementation.

See https://gitlab.com/libssh/libssh-mirror/-/merge_requests/145 for where this is being added to libssh that lead to find this issue.


djmdjm:
merged upstream and synced as 63238f5ae. This will be in OpenSSH 8.5, due very soon - thanks!

PR #219, opened 2020-12-26
Support finding Kerberos via pkg-config

Ericson2314:
This makes cross compilation easier.

The diff renders very poorly due to indenting the existing code. View the second commit in isolation to see what the actual change is.


Ericson2314:
Ah there is. Sent to https://lists.mindrot.org/pipermail/openssh-unix-dev/2021-June/039440.html

daztucker:
Applied the ones you sent to the list (I tried to maintain the authorship info from these PRs though).  Thanks.

PR #220, opened 2020-12-29
Remove redundant `attrib_clear` from SFTP client

pmeinhardt:
The `stat_to_attrib()` call just below the removed line calls `attrib_clear()` itself.

https://github.com/openssh/openssh-portable/blob/46f4a853c4681153ef45b08dfb87df6a6caebe5a/sftp-common.c#L67-L82

### DIFF
--- a/sftp-client.c
+++ b/sftp-client.c
@@ -1853,7 +1853,6 @@ upload_dir_internal(struct sftp_conn *conn, const char *src, const char *dst,
 	if (print_flag)
 		mprintf("Entering %s\n", src);
 
-	attrib_clear(&a);
 	stat_to_attrib(&sb, &a);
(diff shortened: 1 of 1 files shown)


pmeinhardt:
🤔 I am unsure whether/how the CI failure relates to this single-line change. Any pointers would be appreciated. 🙇

pmeinhardt:
> 🤔 I am unsure whether/how the CI failure relates to this single-line change. Any pointers would be appreciated. 🙇

Ah. I branched off of `master` and the `master` branch is red. 🚥 I could rebase onto the last "green" commit?

pmeinhardt:
Looks like this is only failing because the `master` branch is "red". How do you suggest to proceed in this case?

daztucker:
Applied upstream, thanks.

PR #221, opened 2020-12-29
Small sftp-common.c refactoring

pmeinhardt:
Make the attrib-stat conversion functions more consistent.

1.  Extract `stat_clear()` function

    Make stat struct clearing consistent with how attributes are cleared, using the `attrib_clear()` function.

2.  Clear attributes using `memset()`, same as stats

    We are using `memset()` to clear stats. We can use the same approach to clear attribute structs.


pmeinhardt:
I branched off of `master` and the `master` branch is red, which is probably what causes the build for this PR to fail. 🚥 I could rebase onto the last "green" commit to make this PR pass. What do you think?

pmeinhardt:
Looks like this is only failing because the `master` branch is "red". How do you suggest to proceed in this case?

PR #222, opened 2021-01-11
Remove duplicated declaration in fatal.c .

anatasluo:
Declaration of sshfatal has been included in log.h. So there is no need to declare it again. And also, I check the code in [openbsd](https://github.com/openbsd/src/blob/2d2e8ab08946096eec777725afff21569594ceef/usr.bin/ssh/fatal.c), there is no such declaration in fatal.c. So maybe it is duplicated?


daztucker:
merged, thanks.

PR #223, opened 2021-01-14
Do not skip the second leading token backslash in argv_split.

eehakkin:


misc.c
```diff
@@ -1935,10 +1935,7 @@ argv_split(const char *s, int *argcp, char ***argvp)
 
 		/* Start of a token */
 		quote = 0;
-		if (s[i] == '\\' &&
-		    (s[i + 1] == '\'' || s[i + 1] == '\"' || s[i + 1] == '\\'))
-			i++;
```
eehakkin, line 1940: This used to skip the first backslash.

djmdjm: Do you have an example string that it gets wrong? I'll make a unit test for this function from it.

eehakkin: It replaces one, two and three backslashes with one backslash whereas I would have expected it to replace three (and four) backslashes with two backslashes.
```c
int argc;
char **argv;
argv_split("\\a", &argc, &argv);
assert(strspn(argv[0], "\\") == 1);
argv_split("\\\\a", &argc, &argv);
assert(strspn(argv[0], "\\") == 1);
argv_split("\\\\\\a", &argc, &argv);
assert(strspn(argv[0], "\\") == 2);  /* Fails because strspn returns 1. */
argv_split("\\\\\\\\a", &argc, &argv);
assert(strspn(argv[0], "\\") == 2);
```

djmdjm: ah, there's also a related bug in quote handling, e.g. the string ```foo" bar"``` gets turned into {"foo", "bar"} instead of {"foo bar"}

djmdjm: This has been committed - thanks!

PR #224, opened 2021-01-20
Allow overriding TERM with SetEnv directives in ssh_config

lf-:
This removes the necessity to change profile files that are potentially
immutable or unsupported while using terminals that are not in the remote
server's termcap.

My personal use for this is that I use `alacritty` which is `xterm` compatible, but I often have to log into old systems that don't have `alacritty` in termcap, so I want to override `TERM` to be `xterm`. The alternative to this change is writing a wrapper script which would have to parse the host specification and other details, which is not ideal.

Let me know if this is the right avenue to submit this change.

Related threads: https://lists.mindrot.org/pipermail/openssh-unix-dev/2018-November/037361.html; https://serverfault.com/q/986847/322245

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -509,6 +509,22 @@ default_ssh_port(void)
 }
 
 /*
+ * Gets the value of the TERM environment variable as set in the options,
+ * defaulting to using that of the calling environment.
+ */
+const char *
+get_term(const Options *options) {
+    const char *term = NULL;
+    for (int idx = 0; idx < options->num_setenv; idx++) {
+        if (!strncmp("TERM=", options->setenv[idx], 5)) {
+            // skip the =
+            term = options->setenv[idx] + 5;
+        }
+    }
+    return term ? term : getenv("TERM");
+}
+
+/*
  * Execute a command in a shell.
  * Return its exit status or -1 on abnormal exit.
  */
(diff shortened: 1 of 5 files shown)


Lazarus535:
Some time ago i wrote this question on stackoverflow:

https://serverfault.com/questions/986847/ssh-behavior-of-setenv-for-term-variable

My usecase/motivation is the same as @lf- and i think many people will profit from this.
I looked through the code and it appears fine to me, although i am not an expert in this codebase (or C) for that matter.

readconf.c
```diff
@@ -508,6 +508,22 @@ default_ssh_port(void)
...
 
+/*
+ * Gets the value of the TERM environment variable as set in the options,
+ * defaulting to using that of the calling environment.
+ */
+const char *
+get_term(const Options *options) {
+    const char *term = NULL;
+    for (int idx = 0; idx < options->num_setenv; idx++) {
+        if (!strncmp("TERM", options->setenv[idx], 4)) {
```
jchook, line 519: Should this be `strncmp("TERM=", options->setenv[idx], 5)` ?

lf-: Yes, that's indeed buggy on environment variables starting with TERM that aren't TERM. Good eye!

lf-: I've fixed this and force-pushed.

djmdjm:
committed something similar in f64f8c00d1

PR #225, opened 2021-01-25
Fix error when DEBUG_SK is on and Unify variable names in sk part

ZenithalHourlyRate:
## The error
```console
$ make CFLAGS="-DDEBUG_SK"
ranlib libssh.a
ssh-sk.c: In function ‘sshsk_sign’:
ssh-sk.c:698:19: error: ‘message’ undeclared (first use in this function)
  698 |  sshbuf_dump_data(message, sizeof(message), stderr);
      |                   ^~~~~~~
ssh-sk.c:698:19: note: each undeclared identifier is reported only once for each function it appears in
make: *** [Makefile:195: ssh-sk.o] Error 1
```

This error is introduced by 59d2de956ed29aa5565ed5e5947a7abdb27ac013
In this commit, variable `message` is removed from func `sshsk_sign` into `sk_sign` in `sk-usbhid.c`
but its debug code is not removed.

## Variable names

Also, variable names in func argument was renamed in definition but not
in declaration, so in this commit their names are unified.

### DIFF
--- a/sk-api.h
+++ b/sk-api.h
@@ -86,7 +86,7 @@ int sk_enroll(uint32_t alg, const uint8_t *challenge, size_t challenge_len,
     struct sk_option **options, struct sk_enroll_response **enroll_response);
 
 /* Sign a challenge */
-int sk_sign(uint32_t alg, const uint8_t *message, size_t message_len,
+int sk_sign(uint32_t alg, const uint8_t *data, size_t data_len,
     const char *application, const uint8_t *key_handle, size_t key_handle_len,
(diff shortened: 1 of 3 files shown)


djmdjm:
Thanks - I committed a modified version of this upstream and synced it back as 34c5ef6e2 and 88e3d4de3 - it will be included in the OpenSSH 8.5 release (due shortly). The one I committed prints the original unhashed message rather than adding a DEBUG_SK to sk-usbhid.c, as there is no such debugging there at present. I figure this is a very niche feature to begin with :)

PR #226, opened 2021-02-04
ssh: optional bind interface if bind address specified.

ak503:
Allows the -b and -B options to be used together.
For example, when the interface is in the VRF.

### DIFF
--- a/sshconnect.c
+++ b/sshconnect.c
@@ -390,6 +390,12 @@ ssh_create_socket(struct addrinfo *ai)
 		}
 		memcpy(&bindaddr, res->ai_addr, res->ai_addrlen);
 		bindaddrlen = res->ai_addrlen;
+
+		/*'-B' option*/
+		if (options.bind_interface != NULL) {
+			if (setsockopt(sock, SOL_SOCKET, SO_BINDTODEVICE, options.bind_interface, strlen(options.bind_interface) + 1))
+				debug("error: setsockopt SO_BINDTODEVICE");
+		}
 	} else if (options.bind_interface != NULL) {
 #ifdef HAVE_IFADDRS_H
 		if ((r = getifaddrs(&ifaddrs)) != 0) {


djmdjm:
I think this makes -B do two different things:

if -b has not been specified then -B will bind to the **address** of the specified interface

if -b has been specified, then -B will additionally bind to the named interface.

IMO this behaviour is potentially confusing as the second case is actually a different thing entirely.

Could you elaborate on that cases where setting both would be useful?

ak503:
Hello!
For example my link from client go to via VRF

```
HOST:/home/adm$ ip a s dev ethernet2
7: ethernet2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master vrf1 state UP group default qlen 1000
    link/ether 08:00:27:e1:0c:bf brd ff:ff:ff:ff:ff:ff
    inet 12.12.12.2/24 brd 12.12.12.255 scope global ethernet2
       valid_lft forever preferred_lft forever
HOST:/home/adm$ ip link show vrf vrf1
7: ethernet2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master vrf1 state UP mode DEFAULT group default qlen 1000
    link/ether 08:00:27:e1:0c:bf brd ff:ff:ff:ff:ff:ff
```

before patch: 
```
HOST:/home/adm$ ssh adm@12.12.12.1 
^C
HOST:/home/adm$ 

HOST:/home/adm$ ssh adm@12.12.12.1 -b 12.12.12.2
bind 12.12.12.2: Cannot assign requested address
ssh: connect to host 12.12.12.1 port 22: failure

HOST:/home/adm$ ssh adm@12.12.12.1 -B vrf1
ssh: connect to host 12.12.12.1 port 22: failure

HOST:/home/adm$ ssh adm@12.12.12.1 -B ethernet2
bind 12.12.12.2: Cannot assign requested address
ssh: connect to host 12.12.12.1 port 22: failure
```

after patch:
```
HOST:/home/adm$ ssh adm@12.12.12.1 -B vrf1 -b 12.12.12.2
The authenticity of host '12.12.12.1 (12.12.12.1)' can't be established.
ECDSA key fingerprint is SHA256:gT3ceU+3YWxBqgvoA5J2bq5M3l/pzg4YnAgFkdfZttc.
Are you sure you want to continue connecting (yes/no/[fingerprint])? 
```
its idea from https://www.kernel.org/doc/Documentation/networking/vrf.txt:
```Applications
------------
Applications that are to work within a VRF need to bind their socket to the
VRF device:

    setsockopt(sd, SOL_SOCKET, SO_BINDTODEVICE, dev, strlen(dev)+1);

or to specify the output device using cmsg and IP_PKTINFO.
```

djmdjm:
can you achieve the same result using `ip vrf exec vrf1 ssh adm@12.12.12.1` ?

Otherwise, perhaps a more consistent way to do this would be adding a RDomain keyword to ssh_config similar to the one in sshd_config...

PR #227, opened 2021-02-05
bz3259：Fix log deadlock occurs during sshd running

kircherlike:
We recommend removing the debug printing function in main_sigchld_handler to avoid possible deadlocks.

https://bugzilla.mindrot.org/show_bug.cgi?id=3259


daztucker:
Applied upstream, thanks.

PR #228, opened 2021-02-13
bz2217: connect via SRV record

LittleFox94:
Crude implementation of connection via SRV record. Currently only connects to the lowest-prio, highest-weight server, but this is still useful for non-standard ports.

SRV lookup is only done when a port is not manually given on either command line (`-p`) or config.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -281,6 +281,8 @@ clean:	regressclean
 	rm -f regress/unittests/bitmap/test_bitmap$(EXEEXT)
 	rm -f regress/unittests/conversion/*.o
 	rm -f regress/unittests/conversion/test_conversion$(EXEEXT)
+	rm -f regress/unittests/dns/*.o
+	rm -f regress/unittests/dns/test_dns$(EXEEXT)
 	rm -f regress/unittests/hostkeys/*.o
 	rm -f regress/unittests/hostkeys/test_hostkeys$(EXEEXT)
 	rm -f regress/unittests/kex/*.o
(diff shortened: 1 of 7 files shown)


LittleFox94:
Also posted this to openssh-unix-dev mailing list, but maybe not received via email by everybody due to DMARC fails. I hate email. Web archive has it, tho

Nicba1010:
I just want to say I'm fascinated how little feedback there is on this issue. And the openssh-unix-dev thread on this has been dead for a while. This is my biggest pain point to date for ssh.

LittleFox94:
I got some feedback off list and will try to feed it back here tomorrow or something

Iirc it boiled down to name resolution already being a mess (DNS, ssh_config, ....) and another indirection in that is really hard to add at the correct level as behavior is really fuzzy to define (different people want different behavior) and building something around should be relatively easy (was it ProxyCommand or something?)

And there just does not seem to be a lot of interest

LittleFox94:
For a flexible interpretation of "tomorrow", here the reply I got from @djmdjm:

> I didn't get a sense of strong demand for it in OpenSSH and I do worry about it complicating the already pretty fraught DNS/hostkey/canonicalisation logic
>
> Could SRV resolution work as a standalone ProxyCommand helper? This would let people experiment with it without committing OpenSSH to support it
>
> Note also that there is a ProxyUseFdpass that allows ProxyCommand programs to act solely as "dialers" that set up the connection before passing a connected fd to ssh. This lets them exit and removes the resource cost (and hassle of copying data back and forth).
That might be a good fit for this case if you're interested

And yes, that should probably work in some way - but I didn't yet look into that further.

karenmcd:
those looking for a temporary solution or workaround to this SRV record problem, might find this wrapper interesting: https://github.com/Crosse/sshsrv

As a user this is hugely helpful because I cannot remember all my non-standard port assignments for various users hosts behind my home IP (which I also never remember) while doing 'urgent' (/s) fixes to my game servers from termux on android device while traveling or visiting in-laws.

PR #229, opened 2021-02-23
disable warning for the sandboxing part and conceal it within the

devnexen:
related deprecated api.


djmdjm:
I don't think this is a good idea - hiding the warning doesn't solve the use of the deprecated API and in fact makes it less likely that someone will write the code to move us to the undeprecated one.

PR #234, opened 2021-03-18
ssh: pkcs11: pkcs11_del_provider was missing for ssh.

AdithyaBaglody:
C_Finalize API is needed to properly terminate a C_Initialize.
This can cause pkcs11 objects to be left exposed. This patch
does the required cleanup to keep the sanity of the HSM.

Signed-off-by: Adithya Baglody <adithya.nagaraj.baglody@intel.com>


djmdjm:
This has been merged - thanks!

PR #235, opened 2021-03-19
Added test for ssh client to make sure it does fail on termination signal

Alexxz:
A while ago there was a bug spotted and fixed on https://bugzilla.mindrot.org/show_bug.cgi?id=3281

I'm suggesting to add a functional regression test which covers this functionality to make sure correct behavior is covered by tests.


daztucker:
Applied upstream with a few tweaks.  Thanks!

PR #236, opened 2021-03-19
Fix Windows-generated fingerprint output

beldmit:
ssh-keygen -l mangles the output for the keys generated on Windows machine because of '\r' EOL. This patch fixes it.

### DIFF
--- a/ssh-keygen.c
+++ b/ssh-keygen.c
@@ -962,7 +962,7 @@ do_fingerprint(struct passwd *pw)
 	while (getline(&line, &linesize, f) != -1) {
 		lnum++;
 		cp = line;
-		cp[strcspn(cp, "\n")] = '\0';
+		cp[strcspn(cp, "\r\n")] = '\0';
 		/* Trim leading space and comments */
 		cp = line + strspn(line, " \t");
 		if (*cp == '#' || *cp == '\0')


WSLUser:
This should either be fixed downstream at Win32 within https://github.com/PowerShell/openssh-portable/tree/latestw_all/contrib/win32/win32compat or (since this file doesn't exist in that directory) this should be pushed upstream and verify Linux and BSD are unaffected by the change. Otherwise you'll need to add logic (through an `ifdef windows`) so that it only applies to Windows.

beldmit:
Let me disagree. As the public key file is a text file, it's always a chance that it will be opened with a windows text editor, so non-Linux version should be ready to deal with '\r\n' EOLs.

WSLUser:
I should call out that plain old Notepad built into Windows has line ending support for Linux (thanks to WSL's existence). So users will always have a way (plus most good text editors support multiple platforms such as VS Code, which can read line endings from Mac, Linux, and Windows).

glebfm:
Yesterday I reported the same issue here: [mindrot#3385](https://bugzilla.mindrot.org/show_bug.cgi?id=3385) (thanks @jakubjelinek for pointing out it's duplicate).

Let me copy my description here:

```$ rm -f test*; ./ssh-keygen -q -t rsa -C test@test -N '' -f test
$ ./ssh-keygen -l -f test.pub 
3072 SHA256:Fh8V9v/JyBFlGI0ZqvWMb6480Ldm9dF0XJGTZFnoosI test@test (RSA)
$ unix2dos -q test.pub
$ ./ssh-keygen -l -f test.pub 
 (RSA)HA256:Fh8V9v/JyBFlGI0ZqvWMb6480Ldm9dF0XJGTZFnoosI test@test
```
```
$ read -r key < test.pub
$ printf '%s\r9999 SHA256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdea good@key\n' "$key" > bad.pub
$ ./ssh-keygen -l -f bad.pub 
9999 SHA256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdea good@key (RSA)
$ ./ssh-keygen -l -f bad.pub | less
3072 SHA256:rDIBjjZQEjZTzR3GIp/KERSlnMtsxeFrAbeye6TRoI0 test@test^M9999 SHA256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdea good@key (RSA)
```

glebfm:
> (thanks @jakubjelinek for pointing out it's duplicate).
Sorry, I meant @Jakuje. :)

djmdjm:
Merged in 6993d9f09

PR #237, opened 2021-03-25
Ignoring comments at end of config file lines broke ProxyCommand.

tlsalmin:
If all lines are negated from the hash-sign onwards, any existing
options with embedded hash-signs are truncated. e.g. If there are
bash parts that manipulate strings.

The value is too free form to keep a good tab on when the
hash-sign is meant to be in the value and when not.

Unfortunately as 8.5 is released, there probably already exist
config files that have comments at the end of options.

Changed the comment remover to ignore only lines having a
hash-sign as the first non-space character. This will break 8.5
configurations with hash-comments at the end of functional lines.

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -2133,8 +2133,14 @@ read_config_file_depth(const char *filename, struct passwd *pw,
 		 * NB - preserve newlines, they are needed to reproduce
 		 * line numbers later for error messages.
 		 */
-		if ((cp = strchr(line, '#')) != NULL)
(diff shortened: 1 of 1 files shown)


djmdjm:
See https://bugzilla.mindrot.org/show_bug.cgi?id=3288 for the planned fix

djmdjm:
Fixes for this were committed in ea9e45c89a4 and a10f929d1c. They should be in the openssh-8.7 release, due in a few months

PR #238, opened 2021-03-29
Coverity fixes

beldmit:
This is a set of patches fixing the issues found during the coverity scan of the openssh 8.5.

### DIFF
--- a/channels.c
+++ b/channels.c
@@ -436,9 +436,9 @@ channel_register_fds(struct ssh *ssh, Channel *c, int rfd, int wfd, int efd,
 	} else if (nonblock) {
 		if (rfd != -1)
 			set_nonblock(rfd);
-		if (wfd != -1)
+		if (wfd >= 0)
 			set_nonblock(wfd);
-		if (efd != -1)
+		if (efd >= 0)
 			set_nonblock(efd);
 	}
 }
--- a/krl.c
+++ b/krl.c
@@ -1262,6 +1262,7 @@ is_key_revoked(struct ssh_krl *krl, const struct sshkey *key)
 		return r;
 	erb = RB_FIND(revoked_blob_tree, &krl->revoked_sha1s, &rb);
 	free(rb.blob);
+	rb.blob = NULL; /* make coverity happy */
 	if (erb != NULL) {
 		KRL_DBG(("revoked by key SHA1"));
 		return SSH_ERR_KEY_REVOKED;
@@ -1272,6 +1273,7 @@ is_key_revoked(struct ssh_krl *krl, const struct sshkey *key)
 		return r;
 	erb = RB_FIND(revoked_blob_tree, &krl->revoked_sha256s, &rb);
 	free(rb.blob);
+	rb.blob = NULL; /* make coverity happy */
 	if (erb != NULL) {
 		KRL_DBG(("revoked by key SHA256"));
 		return SSH_ERR_KEY_REVOKED;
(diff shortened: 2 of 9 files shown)


daztucker:
Some of these have been fixed (eg the addr.c and ssh-keygen.c moduli screening ones).  Could you please update and rebase?

beldmit:
@daztucker done.

beldmit:
Whoops, sorry, a typo :(

beldmit:
Fixup pushed

beldmit:
The failures seem irrelevant

serverloop.c
```diff
@@ -543,7 +543,7 @@ server_request_tun(struct ssh *ssh)
 		debug_f("invalid tun");
 		goto done;
 	}
-	if (auth_opts->force_tun_device != -1) {
+	if (auth_opts->force_tun_device >= 0) {
```
daztucker, line 546: what's this fixing?

monitor.c
```diff
@@ -1599,7 +1599,6 @@ mm_answer_pty(struct ssh *ssh, int sock, struct sshbuf *m)
 	s->ptymaster = s->ptyfd;
 
 	debug3_f("tty %s ptyfd %d", s->tty, s->ttyfd);
-
```
daztucker, line 1602: no whitespace-only changes please.

sshd.c
```diff
@@ -525,8 +525,10 @@ privsep_preauth(struct ssh *ssh)
 
 		privsep_preauth_child();
 		setproctitle("%s", "[net]");
-		if (box != NULL)
+		if (box != NULL) {
 			ssh_sandbox_child(box);
+			free(box);
```
daztucker, line 530: sent upstream, waiting for an ok.

daztucker: Damien points out that this will cause problems for other sandboxes where the sandbox has some persistent (non-kernel) state, eg the systrace sandbox since the contents of "box" is needed throughout the life of the preauth privsep supervisor process.

Ideally we'd add a ssh_sandbox_child_finish and call it near mm_send_keystate, but the "box" variable doesn't exist in that scope and getting it there would be messy.

One option would be to free "box" at the end of ssh_sandbox_child in any sandbox implementation that can do that, and do nothing for the ones that can't.  Another option would be to just ignore it, because while it is technically a leak, it's in a process that exists only during the authentication process and exits once the authentication succeeds or fails.

channels.c
```diff
@@ -436,9 +436,9 @@ channel_register_fds(struct ssh *ssh, Channel *c, int rfd, int wfd, int efd,
 	} else if (nonblock) {
 		if (rfd != -1)
 			set_nonblock(rfd);
-		if (wfd != -1)
```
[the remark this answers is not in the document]
beldmit: I think it's an artifact of our coverity run, they should be uniform.

krl.c
```diff
@@ -1261,6 +1261,7 @@ is_key_revoked(struct ssh_krl *krl, const struct sshkey *key)
 		return r;
 	erb = RB_FIND(revoked_blob_tree, &krl->revoked_sha1s, &rb);
 	free(rb.blob);
+	rb.blob = NULL; /* make coverity happy */
```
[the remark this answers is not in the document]
beldmit: Don't remember, can skip it

serverloop.c
beldmit: Sorry, don't remember. Better skip it.

beldmit:
Resolved, thanks. What's better - to remove the places where the changes are suspicious?

daztucker:
I think we've got most of this and the codebase has moved on quite a bit.  If you have any additional things or things we missed please make a new PR, preferably one per reason (ie "this PR fixes this particular Coverity warning here, here and here") which would make it easier to review and merge.

PR #239, opened 2021-03-31
fix control socket listener (didn't chdir /)

rafork:
When a control socket is in use, the process inherits and retains the existing current working directory. This is a problem when it happens to be in the file system of a removeable storage device like a USB stick. Later, when the user tries to umount the removeable storage, it fails because an ssh process is still using it as a current working directory. The ssh process has to be identified and terminated before the removeable storage can be umounted. This patch adds chdir("/") at the end of muxserver_listen() in mux.c to solve this problem.


rafork:
Replaced with a different pull request #243 with a different approach.

PR #241, opened 2021-04-03
Refer to KEX algorithms as such, not methods in man page

scop:


daztucker:
Applied upstream, thanks.

PR #242, opened 2021-04-03
Note availability of -/+/^ for CASignatureAlgorithms

scop:
Don't know if `^` should be documented here, as this is not a prioritized list, so it's pretty much the same as `+`, but incuding it matches the implementation.


djmdjm:
fixed in dcce2a2bcf0

PR #243, opened 2021-04-09
fix control socket listener (didn't chdir /)

rafork:
When a control socket is in use, the process inherits and retains the existing current working directory. This is a problem when it happens to be in the file system of a removeable storage device like a USB stick. Later, when the user tries to umount the removeable storage, it fails because an ssh process is still using it as a current working directory. The ssh process has to be identified and terminated before the removeable storage can be umounted. This patch changes daemon(1, 1) in the control_persist_detach() function to daemon(0, 1) so that daemon() will chdir /.

### DIFF
--- a/ssh.c
+++ b/ssh.c
@@ -1736,7 +1736,7 @@ control_persist_detach(void)
 	}
 	if (stdfd_devnull(1, 1, !(log_is_on_stderr() && debug_flag)) == -1)
 		error_f("stdfd_devnull failed");
-	daemon(1, 1);
+	daemon(0, 1);
 	setproctitle("%s [mux]", options.control_path);
 }
 


djmdjm:
the problem here is that the user may have specified cwd-relative files for particular things that are needed later (e.g. Unix domain socket forwardings) and these will break if the directory changes

rafork:
On Thu, Sep 02, 2021 at 10:11:18PM -0700, Damien Miller ***@***.***> wrote:

> the problem here is that the user may have specified cwd-relative
> files for particular things that are needed later (e.g. Unix domain
> socket forwardings) and these will break if the directory changes

And it's a problem if they haven't done so.

I'm surprised that anyone would specify cwd relative files.
I could understand $HOME relative files, but not cwd relative files.
Shows what I know. :-)

What's the best default? Assuming that cwd relative files are used,
or assuming that they haven't been used?

Perhaps the correct solution is to change the directory
to / for the control socket listener, and transform all
cwd relative file references into absolute file
references.

That way, both use cases work.

cheers,
raf

PR #244, opened 2021-04-09
Restore nonblock

beldmit:
OpenSSH incorrectly restores the standard mode (blocking mode) on standard output upon exiting. This causes the next shell scripts commands to potentially fail in EAGAIN.
The reproducer is:
 
```
#!/bin/sh
(
ssh localhost true
cat /dev/zero
) | sleep 30
 
```
 
Restoring the blocking modes happens with the duped file descriptors and too late.
 
The changes causing this problem was introduced in https://github.com/openssh/openssh-portable/commit/4d5456c7de108e17603a0920c4d15bca87244921
 
We introduce storing the real blocking state of the file descriptors and restoring it on closing the fds.
We also remove the too late restoring of the blocking mode


djmdjm:
A fix for this was committed in 7be4ac813662

PR #246, opened 2021-04-22
Correctly restoring channel's FDs nonblocking mode

beldmit:
OpenSSH incorrectly restores the standard mode (blocking mode) on standard output upon exiting. This causes the next shell scripts commands to potentially fail in EAGAIN.
The reproducer is:

#!/bin/sh
(
ssh localhost true
cat /dev/zero
) | sleep 30
 
Restoring the blocking modes happens with the duped file descriptors and too late.

The changes causing this problem was introduced in 4d5456c
The previous patches was provided in #244

This is an alternate approach to fix this problem provided by Masaki Kimura


djmdjm:
Thanks - a similar fix was committed in similar fix was

djmdjm:
(err, truncated)

... committed in 7be4ac813

PR #247, opened 2021-05-06
fix debug msg in client during publickey cert auth

alex-sherwin:
the `sign_id->filename` should be shown when it's present during publickey
certificate based authentication exchange


djmdjm:
Committed as d45b9c63f94  - thanks

PR #248, opened 2021-05-07
mitigate CVE-2020-14145 - allways use default algorithms

manfred-kaiser:
CVE-2020-14145 describes a method, which allows to figure out, if a client connects for the first time or has already stored a known host key.

By removing the preferred host keys and always using default hostkeys, a mitm attacker is not able to exploit CVE-2020-14145.

### DIFF
--- a/sshconnect2.c
+++ b/sshconnect2.c
@@ -253,16 +253,10 @@ ssh_kex2(struct ssh *ssh, char *host, struct sockaddr *hostaddr, u_short port,
 	    (char *)compression_alg_list(options.compression);
 	myproposal[PROPOSAL_MAC_ALGS_CTOS] =
 	    myproposal[PROPOSAL_MAC_ALGS_STOC] = options.macs;
-	if (use_known_hosts_order) {
-		/* Query known_hosts and prefer algorithms that appear there */
-		myproposal[PROPOSAL_SERVER_HOST_KEY_ALGS] =
-		    compat_pkalg_proposal(ssh,
-		    order_hostkeyalgs(host, hostaddr, port, cinfo));
-	} else {
-		/* Use specified HostkeyAlgorithms exactly */
-		myproposal[PROPOSAL_SERVER_HOST_KEY_ALGS] =
-		    compat_pkalg_proposal(ssh, options.hostkeyalgorithms);
-	}
+	
+	/* Use specified HostkeyAlgorithms exactly */
+	myproposal[PROPOSAL_SERVER_HOST_KEY_ALGS] =
+	    compat_pkalg_proposal(ssh, options.hostkeyalgorithms);
 
 	if (options.rekey_limit || options.rekey_interval)
 		ssh_packet_set_rekey_limits(ssh, options.rekey_limit,


trymine:
Hi,
https://github.com/openssh/openssh-portable/commit/b3855ff053f5078ec3d3c653cdaedefaa5fc362d
The above modification is also to solve this problem. Is there any difference between the current modification and the above modification?

manfred-kaiser:
This is the fix, which was introduced with OpenSSH 8.4.

When the server uses a `ecdsa-sha2` host key, the ordering will not be altered. If the server uses another algorithm, the client does not use the default order. So the problem still exists.

You can test it with my ssh-mitm tool https://github.com/ssh-mitm/ssh-mitm
The check against CVE-2020-14145 is enabled by default,

**Steps to reproduce:**

Installation:

- download OpenSSH 8.5 and/or 8.6 and compile it
- install the latest verion of SSH-MITM:  `pip install ssh-mitm` 
- start the ssh-mitm server with: `ssh-mitm`
- connect the compiled ssh clients to the server: `ssh -p 10022 localhost`

If you do net have a known host key, your client should ask you, if you accept it. You should see following log message from ssh-mitm:

`[INFO]  CVE-2020-14145: Client connecting for the FIRST time!`

Acept the host key and close the client. We will need a known host key for the next step.

Connect again to the server: `ssh -p 10022 localhost`

The client should not ask you for the host key and you should see an login promt. 
ssh-mitm will show following log message:

`[INFO]  CVE-2020-14145: Client has a locally cached remote fingerprint!`

Note: If you got a waring on changed host keys, you have restarted ssh-mitm, which will generate new host keys. If you have already a known host key, this does not affect the test. because the check runs, before the client disconnects.

You can read more in the documentation of ssh-mitm: https://docs.ssh-mitm.at/CVE-2020-14145.html

**Mitigation**

To mitigate CVE-2020-14145, the client must use default ordering.

trymine:
Thanks for your answer.

djmdjm:
No addition option is needed to change the ordering. If you want to hard-code the default then use the HostKeyAlgorithms option.

Between b3855ff and the more recent change to enable UpdateHostKeys, most users should soon receive the default ordering anyway.

manfred-kaiser:
> No additional option is needed to change the ordering. If you want to hard-code the default then use the existing HostKeyAlgorithms option.

Most users do not change the `HostKeyAlgorithms` option. If you are a company with a central deployment, this is possible, but if you install a fresh environment and forget to set this option, you are vulnerable.

> Between b3855ff and the more recent change to enable UpdateHostKeys, most users should soon receive the default ordering anyway.

Does this mean, that current versions of OpenSSH server always send the default order? Can you explain the changes in more detail or do you have a commit hash?

If this is the case, one problem is, that most (linux) distributions does not have the latest version of openssh server installed and those changes need to be backported.

**Note:** I have tested the clients in Version 8.5 and 8.6. The current master was not tested. 

**Please explain, how do you mitigate CVE-2020-14145 in current versions.** At the moment, I think the best mitigation is to remove the option for preffered host keys algorithms from the client.

manfred-kaiser:
Please not, that a functional exploit for CVE-2020-14145 is available in [SSH-MITM](https://github.com/ssh-mitm/ssh-mitm)

ssh-mitm integrates only a check, if a client is vulnerable, biut an attacker can write a plugin, which only intercepts clients, when the server host key is not known. 

This allows an attacker to stay hidden for a longer time and increases the chance, that a user accepts the unknown fingerprint without verification to a trusted source.

manfred-kaiser:
I would prefer to discuss this vulnerability on another channel. I have sent you a mail to you mailing list, but I didn't received any answers :-(

**If you need help to test some changes, I can do it for you.** 

The goal of ssh-mitm is not to weaken and exploit ssh connections. It is intended as an audit tool and should help to avoid or fix vulnerabilities.

manfred-kaiser:
I have search through the latest commits and did not found a fix for CVE-2020-14145. Do you have a commit hash?

@djmdjm  How do you verify, that this vulnerability is fixed? Can you describe your tests?

manfred-kaiser:
The vulnerability was also verified by Nist and CPEs for version 8.5 and 8.6 were added to CVE-2020-14145.

Source: https://nvd.nist.gov/vuln/detail/CVE-2020-14145#vulnConfigurationsArea

The versions in the description should be updated in the next days by Mitre.

PR #249, opened 2021-05-10
Bypass File Permissions

florian-lnx:
Bypass File Permission Check


florian-lnx:
Oh sorry, wrong Repo, this was for my Version of sshd for Archlinux with proot on Android.

PR #252, opened 2021-05-20
Treat DNS authoritive answer as validated

huww98:
This is useful when we set up an authoritative DNS server and reuse it to serve
recursive query from clients.

### DIFF
--- a/openbsd-compat/getrrsetbyname.c
+++ b/openbsd-compat/getrrsetbyname.c
@@ -273,7 +273,7 @@ getrrsetbyname(const char *hostname, unsigned int rdclass,
 
 #ifdef HAVE_HEADER_AD
 	/* check for authenticated data */
-	if (response->header.ad == 1)
+	if (response->header.ad == 1 || response->header.aa == 1)
 		rrset->rri_flags |= RRSET_VALIDATED;
 #endif
 


jaceppo (approved):

DemiMarie:
Does this create a security hole in some setups?  For instance, does `systemd-resolved` strip the `aa` bit from responses?

huww98:
> Does this create a security hole in some setups? For instance, does `systemd-resolved` strip the `aa` bit from responses?

Yes, `systemd-resolved` strip the `aa` bit. But why does this create a security hole? If `aa` bit is not present, then ssh just behaves as before.

DemiMarie:
`systemd-resolved` stripping the `aa` bit is safe, but consider an untrusted DNS server on public Wi-Fi.  My understanding is that many libraries strip the `ad` bit unless the resolver is trusted (such as by being on `localhost`), but they may not do the same for the `aa` bit.  See the `trust-ad` option in `resolv.conf` on some systems.

huww98:
Thanks for that info. It makes sense. So maybe the best solution now is setting up `systemd-resolved` on every client to do the DNSSEC validation locally then set the `ad` bit.

PR #253, opened 2021-05-29
Allow import Edwards keys from OpenSSL

beldmit:
Modern OpenSSL versions support ed25519 keys. These patches add the detection of the necessary functions and uses them for import of ed25519 keys to OpenSSH


djmdjm:
support for reading EDDSA PEM keys was added a while ago

PR #254, opened 2021-06-02
Support Unix socket forwarding with -W

steffen-kiess:
Currently ssh -W will parse Unix socket arguments correctly, but will
then fail to set up a tunnel because parse_forward() sets listen_path
instead of listen_host.

Choose the correct value for stdio_forward_host and use the correct
rtype in channel_connect_stdio_fwd().


kalvdans:
Implemented by 94842bfe9b09fc93189c6ed0dc9bbebc1d44a426, I think we can close this PR.

steffen-kiess:
Yes, this PR can be closed.

PR #255, opened 2021-06-04
Standardize the string functions

gAlfonso-bit:
Current standards say strlcpy and strlcat must have restrict qualifiers in their function definitions, also mode_t is now the type of the first argument in strmode.


daztucker:
> Current standards

which standards?  OpenSSH requires nothing beyond C89 and POSIX.

strlcat and strlcpy implementations originate from OpenBSD (https://github.com/openbsd/src/blob/master/lib/libc/string/strlcpy.c, https://github.com/openbsd/src/blob/master/lib/libc/string/strlcat.c) and those implementations do not have "restrict".   Likewise strmode comes from OpenBSD where it takes an int not a mode_t.

PR #256, opened 2021-06-09
docs: Use more appropriate language when referring to the user

l1ving:
This pull request changes the language used in the `ssh.1` manpage and `sshconnect.c` comments in order to be more modern and correct, when referring to the user. The [singular they](https://www.merriam-webster.com/words-at-play/singular-nonbinary-they) pronoun is more appropriate to use here, and is less awkward than saying `his/her`.

### DIFF
--- a/ssh.1
+++ b/ssh.1
@@ -88,7 +88,7 @@ or a URI of the form
 .No ssh:// Oo user @ Oc hostname Op : port .
 .Sm on
 The user must prove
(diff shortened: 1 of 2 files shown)


daztucker:
Applied upstream, it'll be in the next major release.  Thanks.

Motardinho:
This "modern" world beats records of stupidity every single day...

PR #258, opened 2021-06-17
CVE-2021-36368 - added option to disable trivial auth methods

manfred-kaiser:
I have added an option ``-o DisableTrivialAuth=yes`` to disable trivial authentications as mentioned in my mails


manfred-kaiser:
i will also support for ssh1, because ssh1 is also affected

manfred-kaiser:
Would you merge this pull request, because we are planning to create a CVE and I think this should be mitigated before we publish a full disclosure.

manfred-kaiser:
commit 9e1882ef6489a7dd16b6d7794af96629cae61a53 introduced a change in verbose logging, which can be used to detect trivial success authentication. This is an improvement, but is to complicated and inconvenient to mitigate it.
This is the reason, why I have updated my pull request to the latest master.

manfred-kaiser:
AUT-milCERT want's to release information about "trivial success authentication" in the next weeks. 

Do you want to merge our patch or are you planning to implement some other mitigation approaches in the next release?

djmdjm:
No, we do not plan to merge this patch.

PR #259, opened 2021-06-18
Basic support of RFC 6594

beldmit:
Set SHA256 as a default digest for RSA/DSA SSHFP records

### DIFF
--- a/dns.c
+++ b/dns.c
@@ -87,12 +87,12 @@ dns_read_key(u_int8_t *algorithm, u_int8_t *digest_type,
 	case KEY_RSA:
 		*algorithm = SSHFP_KEY_RSA;
 		if (!*digest_type)
-			*digest_type = SSHFP_HASH_SHA1;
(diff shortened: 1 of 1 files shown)


daztucker (commented):
~~The problem is that doing this will mean that any currently-working configurations relying on SHA1 fingerprints will stop working unless the DNS also contains the corresponding SHA256 records.~~

edit: actually there is a check hostkey_digest_type != dnskey_digest_type which will override the default type.  If I'm reading this right the check inside the loop means it'll always end up using the hash algos returned by DNS and your change won't have any effect.  (Whether or not its current behaviour is desirable is a separate question).

daztucker:
The behaviour was changed in b75a80fa8369864916d4c93a50576155cad4df03 such that it'll verify every fingerprint available to it and bail if any fail to match.

Additional discussion available in https://bugzilla.mindrot.org/show_bug.cgi?id=3322

PR #260, opened 2021-06-20
thx

SiraphobS:
thx


djmdjm:
this doesn't look like an intentional PR to the OpenSSH project

PR #262, opened 2021-06-28
Class-imposed login restrictions (simplified)

yuichiro-naito:
If the following functions are available, add an additional check if users are allowed to login imposed by login class.

* auth_hostok(3)
* auth_timeok(3)

These functions are implemented on FreeBSD.

After I got the advice in the https://github.com/openssh/openssh-portable/pull/261,
I changed my patch not to leak any informations includes behaviors when authentication is failed.

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -566,6 +566,9 @@ getpwnamallow(struct ssh *ssh, const char *user)
 {
 #ifdef HAVE_LOGIN_CAP
 	extern login_cap_t *lc;
+#ifdef HAVE_AUTH_HOSTOK
+	const char *from_host, *from_ip;
+#endif
 #ifdef BSD_AUTH
 	auth_session_t *as;
 #endif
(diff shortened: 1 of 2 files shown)


yuichiro-naito:
I updated my patch not to log a message when authentication is failed.
I re-considered not to leak any informations about authentication failure.
If syslog is sent over the network, it is possible to detect the reason by watching the packets.
So I changed log message to debug.

Please review my code for merging.

emaste:
I'd suggest squashing the two commits into one.

I've mentioned this pull request on the openssh mailing list, and am preparing to update FreeBSD's base system OpenSSH w/ this patch.

emaste:
Review to apply this change to (the older) OpenSSH in the FreeBSD base system: https://reviews.freebsd.org/D31760

yuichiro-naito:
Thanks for the comment.
I squashed my commits into one and force pushed.

emaste:
Now committed to FreeBSD base system: https://reviews.freebsd.org/R10:27ceebbc2402e4c98203c7eef9696f4bd3d326f8

djmdjm:
Merged as c276672fc0e99f0

PR #264, opened 2021-07-21
OpenSSH's ssh-keygen can't parse encrypted PKCS#8 private keys being …

beldmit:
…built against openssl 3.0

See bz#3330

### DIFF
--- a/sshkey.c
+++ b/sshkey.c
@@ -4504,6 +4504,16 @@ translate_libcrypto_error(unsigned long pem_err)
 		default:
 			return SSH_ERR_INVALID_FORMAT;
 		}
+#ifdef ERR_R_INTERRUPTED_OR_CANCELLED
+	case ERR_LIB_CRYPTO:
+		switch (pem_reason) {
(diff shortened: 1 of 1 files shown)


daztucker (commented):
It seems that if  ERR_R_INTERRUPTED_OR_CANCELLED is not defined, ERR_LIB_CRYPTO should still return SSH_ERR_INVALID_FORMAT (ie the scope of the ifdef should be smaller)?

beldmit:
Yes, we could narrow the scope, but we shouldn't get a ERR_LIB_CRYPTO in such a situation.

beldmit:
I reduced the #ifdef scope.

beldmit:
`make tests` passed successfully, both with and without this patch.

daztucker:
I added a comment to the bugzilla bug with more detail, but I was not able to reproduce the problem built against the OpenSSL 3.0.0 dev branch head.

I also added a regression test in 8b7af02dcf9d2b738787efd27da7ffda9859bed2 which I think should catch the problem you're describing (but since I can't reproduce it, I'm not sure).

PR #266, opened 2021-08-11
Fix memory leak bug

Shchelk:
the pointer to the allocated memory is overwritten and lost on line 889


daztucker:
Applied in b6aac22fd0426bc38e47bc4f9c2a2f915cb37c70, thanks.

PR #267, opened 2021-08-25
Bug 3210 - Confusing errors when pam_acct_mgmt() fails

darrenmoffat:
Alternate proposal to preserve pam_acct_mgmt() return value

I'm adding a slightly different proposed patch, that I believe is both more generic, in that all error values from pam_acct_mgmt() can be passed through.  It is also slightly smaller a change.
Unlike the prior patch it intentionally overrides a PAM_SUCCESS sshpam_err value with the one from pam_acct_mgmt(). This is so that an account that has successfully authenticate but for some other reason is not allowed access "just now" has an appropriate error returned (likely PAM_PERM_DENIED).

### DIFF
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -522,10 +522,10 @@ sshpam_thread(void *ctxtp)
 	if (sshpam_err != PAM_SUCCESS)
 		goto auth_fail;
 
-	if (!do_pam_account()) {
-		sshpam_err = PAM_ACCT_EXPIRED;
+	sshpam_err = do_pam_account();
+	if (sshpam_err != PAM_SUCCESS && sshpam_err != PAM_NEW_AUTHTOK_REQD)
 		goto auth_fail;
-	}
+
 	if (sshpam_authctxt->force_pwchange) {
 		sshpam_err = pam_chauthtok(sshpam_handle,
 		    PAM_CHANGE_EXPIRED_AUTHTOK);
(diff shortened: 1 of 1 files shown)


wmgl5593:
> Alternate proposal to preserve pam_acct_mgmt() return value
> 
> I'm adding a slightly different proposed patch, that I believe is both more generic, in that all error values from pam_acct_mgmt() can be passed through. It is also slightly smaller a change.
> Unlike the prior patch it intentionally overrides a PAM_SUCCESS sshpam_err value with the one from pam_acct_mgmt(). This is so that an account that has successfully authenticate but for some other reason is not allowed access "just now" has an appropriate error returned (likely PAM_PERM_DENIED).

[.patch]()

auth-pam.c
```diff
@@ -576,16 +576,15 @@ sshpam_thread(void *ctxtp)
...
 		fatal("%s: buffer error: %s", __func__, ssh_err(r));
 	/* XXX - can't do much about an error here */
-	if (sshpam_err == PAM_ACCT_EXPIRED)
-		ssh_msg_send(ctxt->pam_csock, PAM_ACCT_EXPIRED, buffer);
-	else if (sshpam_maxtries_reached)
+	if (sshpam_maxtries_reached)
 		ssh_msg_send(ctxt->pam_csock, PAM_MAXTRIES, buffer);
 	else
-		ssh_msg_send(ctxt->pam_csock, PAM_AUTH_ERR, buffer);
+		ssh_msg_send(ctxt->pam_csock, sshpam_err, buffer);
```
djmdjm, line 587: I don't understand this part of the change - AFAIK the `sshpam_err` value gets read back in `sshpam_query()` via `ssh_msg_recv()`, but that function would handle arbitrary error codes via its `default:` switch case - does this do what we want in all circumstances?

Maybe I'm not following (thanks for having largely swapped out the PAM code from my brain), but the material part of this change seems to be plumbing the correct error code back to the `sshbuf_put_cstring()` call above, and not this. Am I off-base here?

auth-pam.c
```diff
@@ -576,16 +576,15 @@ sshpam_thread(void *ctxtp)
 	pthread_exit(NULL);
 
  auth_fail:
+	debug3("PAM: auth_fail sshpam_err=%d %s", sshpam_err, pam_strerror(sshpam_handle, sshpam_err));
```
djmdjm, line 579: nits: `s/debug3("PAM: /debug3_f("/` and wrap at 80 columns

djmdjm (commented):
Sorry for slow response. A couple of comments/questions

PR #270, opened 2021-09-15
CVE-2016-20012 - Publickey Information leak - Only allow SSH_MSG_USERAUTH_REQUEST with signature

manfred-kaiser:
CVE: https://nvd.nist.gov/vuln/detail/CVE-2016-20012

This pull request adds a new configuration option to the server to disable user enumeration when using publickey authentication.

The user enumeration was first mentioned on 6. june 2002:
https://github.com/openssh/openssh-portable/blob/d0fffc88c8fe90c1815c6f4097bc8cbcabc0f3dd/auth2-pubkey.c#L261-L265

### Publickey authentication:

Publickey authentication is defined in [RFC-4252](https://datatracker.ietf.org/doc/html/rfc4252#page-8) (released January 2006)

A client is allowed to ask the server, if a specific public key is known by the server. This allows the client to generate a signature only if it's necessary. If the publickey is not known, no signature must be created.

> Private keys are often stored in an encrypted form at the client
   host, and the user must supply a passphrase before the signature can
   be generated.  Even if they are not, the signing operation involves
   some expensive computation.  To avoid unnecessary processing and user
   interaction, the following message is provided for querying whether
   authentication using the "publickey" method would be acceptable.
>
>      byte      SSH_MSG_USERAUTH_REQUEST

**This behavior allows an attacker to check if a user is allowed to login.**

RFC-4252 also defines, that it's allowed to send a `SSH_MSG_USERAUTH_REQUEST` with a signature directly:

>   To perform actual authentication, the client MAY then send a
   signature generated using the private key.  The client MAY send the
   signature directly without first verifying whether the key is
   acceptable.  The signature is sent using the following packet:
>
>      byte      SSH_MSG_USERAUTH_REQUEST

### The patch

This patch adds a new configuration option to the OpenSSH server: `PubkeyDisablePKCheck no`

If this configuration option is enbabled, the client is not able to send a `SSH_MSG_USERAUTH_REQUEST` without signature. In cases, when a client asks the server if a publickey is known, the server denies the login.

A patch was also applied to the client, because the client allways asks the server if a publickey is known. Even, when the configuration option `IdentitiesOnly` is enabled.

If the client was configured with `IdentitiesOnly`, `SSH_MSG_USERAUTH_REQUEST` with a signature is directly sent to the server. This makes the OpenSSH Client compatible with the server.

### Other clients

I have created a patch for PuTTY, to make PuTTY compatible with the new setting: https://github.com/manfred-kaiser/putty/commit/5e33243c57aef8aa12a9c8aa0d198c66834aaaa1

This patch was sent to Simon Tatham (PuTTY developer), with a request to merge it in the next release.

WinSCP, Filezille and other clients, which are based on PuTTY can use the new configuration option. So they should be compatible in the near future.

### DIFF
--- a/auth2-pubkey.c
+++ b/auth2-pubkey.c
@@ -264,9 +264,10 @@ userauth_pubkey(struct ssh *ssh)
 		 * if a user is not allowed to login. is this an
 		 * issue? -markus
 		 */
-		debug_f("options.pubkey_disable_pk_check %d", options.pubkey_disable_pk_check);
-		if (options.pubkey_disable_pk_check)
+		if (options.pubkey_disable_pk_check) {
+			debug_f("SSH_MSG_USERAUTH_REQUEST without signature are forbidden");
 			goto done;
+		}
 		if (PRIVSEP(user_key_allowed(ssh, pw, key, 0, NULL))) {
 			if ((r = sshpkt_start(ssh, SSH2_MSG_USERAUTH_PK_OK))
 			    != 0 ||
--- a/auth2-pubkey.c
+++ b/auth2-pubkey.c
@@ -264,6 +264,9 @@ userauth_pubkey(struct ssh *ssh)
 		 * if a user is not allowed to login. is this an
 		 * issue? -markus
 		 */
+		debug_f("options.pubkey_disable_pk_check %d", options.pubkey_disable_pk_check);
+		if (options.pubkey_disable_pk_check)
+			goto done;
 		if (PRIVSEP(user_key_allowed(ssh, pw, key, 0, NULL))) {
 			if ((r = sshpkt_start(ssh, SSH2_MSG_USERAUTH_PK_OK))
 			    != 0 ||
--- a/readconf.c
+++ b/readconf.c
@@ -154,7 +154,7 @@ typedef enum {
 	oBatchMode, oCheckHostIP, oStrictHostKeyChecking, oCompression,
 	oTCPKeepAlive, oNumberOfPasswordPrompts,
 	oLogFacility, oLogLevel, oLogVerbose, oCiphers, oMacs,
-	oPubkeyAuthentication,
+	oPubkeyAuthentication, oPubkeyDisablePKCheck,
 	oKbdInteractiveAuthentication, oKbdInteractiveDevices, oHostKeyAlias,
 	oDynamicForward, oPreferredAuthentications, oHostbasedAuthentication,
 	oHostKeyAlgorithms, oBindAddress, oBindInterface, oPKCS11Provider,
@@ -233,6 +233,7 @@ static struct {
 	{ "skeyauthentication", oKbdInteractiveAuthentication }, /* alias */
 	{ "tisauthentication", oKbdInteractiveAuthentication },  /* alias */
 	{ "pubkeyauthentication", oPubkeyAuthentication },
+	{ "pubkeydisablepkcheck", oPubkeyDisablePKCheck},
 	{ "dsaauthentication", oPubkeyAuthentication },		    /* alias */
 	{ "hostbasedauthentication", oHostbasedAuthentication },
 	{ "identityfile", oIdentityFile },
@@ -1103,6 +1104,10 @@ parse_time:
 
 	case oPubkeyAuthentication:
 		intptr = &options->pubkey_authentication;
+		goto parse_flag;
+
+	case oPubkeyDisablePKCheck:
+		intptr = &options->pubkey_disable_pk_check;
 		goto parse_flag;
 
 	case oHostbasedAuthentication:
@@ -2332,6 +2337,7 @@ initialize_options(Options * options)
 	options->fwd_opts.streamlocal_bind_mask = (mode_t)-1;
 	options->fwd_opts.streamlocal_bind_unlink = -1;
 	options->pubkey_authentication = -1;
+	options->pubkey_disable_pk_check = -1;
 	options->gss_authentication = -1;
 	options->gss_deleg_creds = -1;
 	options->password_authentication = -1;
@@ -2488,6 +2494,8 @@ fill_default_options(Options * options)
 		options->fwd_opts.streamlocal_bind_unlink = 0;
 	if (options->pubkey_authentication == -1)
 		options->pubkey_authentication = 1;
+	if (options->pubkey_disable_pk_check == -1)
+		options->pubkey_disable_pk_check = 0;
 	if (options->gss_authentication == -1)
 		options->gss_authentication = 0;
 	if (options->gss_deleg_creds == -1)
@@ -3285,7 +3293,7 @@ dump_client_config(Options *o, const char *host)
 	dump_cfg_fmtint(oGssDelegateCreds, o->gss_deleg_creds);
 #endif /* GSSAPI */
 	dump_cfg_fmtint(oHashKnownHosts, o->hash_known_hosts);
-	dump_cfg_fmtint(oHostbasedAuthentication, o->hostbased_authentication);
+	dump_cfg_fmtint(oHostbasedAuthentication, o->pubkey_disable_pk_check);
 	dump_cfg_fmtint(oIdentitiesOnly, o->identities_only);
 	dump_cfg_fmtint(oKbdInteractiveAuthentication, o->kbd_interactive_authentication);
 	dump_cfg_fmtint(oNoHostAuthenticationForLocalhost, o->no_host_authentication_for_localhost);
@@ -3293,6 +3301,7 @@ dump_client_config(Options *o, const char *host)
 	dump_cfg_fmtint(oPermitLocalCommand, o->permit_local_command);
 	dump_cfg_fmtint(oProxyUseFdpass, o->proxy_use_fdpass);
 	dump_cfg_fmtint(oPubkeyAuthentication, o->pubkey_authentication);
+	dump_cfg_fmtint(oPubkeyDisablePKCheck, o->pubkey_disable_pk_check);
 	dump_cfg_fmtint(oRequestTTY, o->request_tty);
 	dump_cfg_fmtint(oSessionType, o->session_type);
 	dump_cfg_fmtint(oStdinNull, o->stdin_null);
--- a/readconf.c
+++ b/readconf.c
@@ -3293,7 +3293,7 @@ dump_client_config(Options *o, const char *host)
 	dump_cfg_fmtint(oGssDelegateCreds, o->gss_deleg_creds);
 #endif /* GSSAPI */
 	dump_cfg_fmtint(oHashKnownHosts, o->hash_known_hosts);
-	dump_cfg_fmtint(oHostbasedAuthentication, o->pubkey_disable_pk_check);
+	dump_cfg_fmtint(oHostbasedAuthentication, o->hostbased_authentication);
 	dump_cfg_fmtint(oIdentitiesOnly, o->identities_only);
 	dump_cfg_fmtint(oKbdInteractiveAuthentication, o->kbd_interactive_authentication);
 	dump_cfg_fmtint(oNoHostAuthenticationForLocalhost, o->no_host_authentication_for_localhost);
--- a/readconf.h
+++ b/readconf.h
@@ -37,6 +37,7 @@ typedef struct {
 	char   *xauth_location;	/* Location for xauth program */
 	struct ForwardOptions fwd_opts;	/* forwarding options */
 	int     pubkey_authentication;	/* Try ssh2 pubkey authentication. */
+	int		pubkey_disable_pk_check; /* If true, disables ssh2 pubkey check. */
 	int     hostbased_authentication;	/* ssh2's rhosts_rsa */
 	int     gss_authentication;	/* Try GSS authentication */
 	int     gss_deleg_creds;	/* Delegate GSS credentials */
--- a/scp.1
+++ b/scp.1
@@ -211,6 +211,7 @@ For full details of the options listed below, and their possible values, see
 .It ProxyJump
 .It PubkeyAcceptedAlgorithms
 .It PubkeyAuthentication
+.It PubkeyDisablePKCheck
 .It RekeyLimit
 .It SendEnv
 .It ServerAliveInterval
--- a/servconf.c
+++ b/servconf.c
@@ -129,6 +129,7 @@ initialize_server_options(ServerOptions *options)
 	options->hostbased_accepted_algos = NULL;
 	options->hostkeyalgorithms = NULL;
 	options->pubkey_authentication = -1;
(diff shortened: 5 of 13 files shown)


manfred-kaiser:
I made this information public, because this vulnerability is **public known at least since 2016!**

https://utcc.utoronto.ca/~cks/space/blog/tech/SSHKeysAreInfoLeak

I also found a script, which was created in 2019, which is able to do the user enumeration:
https://rushter.com/blog/public-ssh-keys/

djmdjm:
There is no "user enumeration" here. This feature allows testing whether a particular key may be accepted by a particular account. There is no general way to query which accounts are valid without valid public keys for each one, much less the ability to enumerate all the accounts on a system. At most, it is an account existence _oracle_ and a limited one at that.

OpenSSH has long had a policy of not treating account existence oracles as security bugs. E.g. https://www.openwall.com/lists/oss-security/2018/08/24/1

The motivation for this feature is to avoid unnecessary unwrapping and exposure of private keys. Removing or disabling it would likely result in users exposing more keys via ssh-agent or by password-less files. We have judged the consequences to outweigh the benefits.

manfred-kaiser:
### Server side:

It's an optional feature and disabled by default. So this will not change the behavior in default installations!
An administrator has to enable this feature on the server.

### Client side:

A connection has to be configured with `IdentitiesOnly yes` and `IdentityFile /home/user/.ssh/id_rsa`.

The important config option is `IdentitiesOnly yes`, because when this option is enabled, only the configured keys are used. **So there is no exposing of keys or the need of unencrypted keys!** 

This configuration option already exists and is not added by the patch. I only changed the bahavior. If this option is configured, the client send `SSH_MSG_USERAUTH_REQUEST` with a signature, instead of asking, if a publickey is usable. 

The user has to configure the keys, which should be used to login to a server. This does not expose any unused keys or bother the user to enter a password for a key, which is not usable.

If you think, that the `IdentitiesOnly` should not be used, a new option can be added. But my intention was not to make it more complicated as necessary.

### Further steps:

I think, the users/administrators should decide, if they want to enable the new setting or not. 
But it's important, that the administrators know about the advantages and disadvantages about the new config option.

As a developer, we should not decide how the software should be used. We should give the users options and let them choose how to use it.


### Disclosure

I'm working on an audit tool for ssh. This tool is capable of man in the middle attacks. In the past it was not able to spoof the login process. The reason was, that the a mitm server didn't know, how the client wants to login to the server.

With this information leak, my tool is able to check, if the users wants to login with password or with publickey authentication. If publickey authentication is possible, my tool accepts the same key as the destination server, which allows other spoofing attacks (we already talked about that).

I know, that the fingerprint mitigates man in the middle attacks, but how many users really compare the fingerprint to a trusted source? But this is another problem...

manfred-kaiser:
fixed the initial value for `options->pubkey_disable_pk_check = 0;` in `servconf.c`, which enabled the new feature `PubkeyDisablePKCheck` if it was not set in the config file.

Now the tests should run without errors and the new option is disabled.

manfred-kaiser:
I'm getting an error for the agent tests :disappointed: 

```
ok connect after login grace timeout
run test agent.sh ...

WARNING: Unsafe (group or world writable) directory permissions found:
 /home/user/openssh-portable/regress /home/user/openssh-portable /home/user

These could be abused to locally escalate privileges.  If you are
sure that this is not a risk (eg there are no other users), you can
bypass this check by setting TEST_SSH_UNSAFE_PERMISSIONS=1

ssh connect with failed (exit code 255)
...
ssh connect with failed (exit code 255)
failed simple agent test
make[1]: *** [Makefile:219: t-exec] Fehler 1
```

The reason is, that I'm reusing the `IdentitiesOnly` config option to send the `SSH_MSG_USERAUTH_REQUEST` directly, which breaks the tests :disappointed:

manfred-kaiser:
I have added a new configuration option `PubkeyDisablePKCheck` to the client. 

It was not intended to introduce a new one, but by reusing `IdentitiesOnly` the tests failed :disappointed: 

The behavior of the client is only changed when **`PubkeyDisablePKCheck` and `IdentitiesOnly`** is set!

If one setting is not enabled, it uses the default behavior and asks if a public key is allowed.

Example config for a client, which sends a `SSH_MSG_USERAUTH_REQUEST` with a signature directly:

```
Host testhost
     HostName localhost
     User user
     Port 10022
     PubkeyDisablePKCheck yes
     IdentitiesOnly yes
     IdentityFile /home/user/.ssh/id_rsa
```

manfred-kaiser:
I have fixed a copy&paste error :disappointed: and the tests are working now.

**Please review the code!**

If I should change some things, please write back.

djmdjm:
It's not a problem of incompatibility. This can AFAIK be avoided by having
the server return success for all PK_OK trials.

I think you're missing why this protocol feature exists and are not
thinking through the consequences of disabling it. It exists to avoid the
client unwrapping private keys unnecessarily. If this is changed, then many
clients will have to unwrap private keys for authentication attempts where
they will likely be unsuccessful.

This has an immediate and direct effect that private keys will be exposed
more frequently on client hosts. You must also consider user reaction to
being required to enter their passphrases, touch their security keys,
confirm their agents, etc unnecessarily and more often. What will users do
in this case? I can predict two effects: 1) some users will react by trying
to remove the usability bur, e.g. by removing passphrases from private
keys, disabling touch requirements for FIDO tokens, etc. 2) the users who
do not do this will become more careless to unwrapping their private keys
in situations where they aren't expected.

I believe this optimises for a case where security has already been forgone
(an attacker has successfully impersonated a trusted host) in a way that
has negative consequences for every other user of the protocol. IMO this is
a bad tradeoff

mahesh621 (commented):
-

mahesh621 (commented):
.

mahesh621 (commented):
@manfred-kaiser 
Has this proposed patch been accepted by the openssh maintainers and integrated latest OPNESSH version?

cjwangd:
could you fix it as soon as possible?

ymecho:
Is this issue fixed now?

daztucker:
The opinion of the OpenSSH team is that this change causes more problems than it solves and we do not intend to apply it.  (See Damien's earlier comment for more details.)  If you disagree you are welcome to use Manfred's fork.

manfred-kaiser:
> Is this issue fixed now?

This "issue" won't get fixed in the near future. One reason is, most (all?) clients rely on the publickey lookup. It also makes ssh more difficult to use and more error prone - you have to configure each connection.

In my opinion, if you want to get rid of the publickey lookup, you have to patch the clients first to be compatible with such a big change. 

But keep in mind this will introduce more problems -> there are applications which rely on unmaintained libraries or will get abandoned in the near future. For example the maintainer of [sshfs](https://github.com/libfuse/sshfs) has stopped development and looks for a new maintainer. With such big changes I fear more developers will stop maintaining their projects.

For example the deprecation of the sha1 algorithms introduced serious problems because a lot of projects where not able provide a compatible version in time. Some of those applications are still not fully compatible or has some errors in handling the introduced features.

One reason for opening this pull request was to discuss the publickey lookup and show the security implications but @daztucker and @djmdjm and even Simon Tatham (PuTTY developer) convinced me that it's a bad idea at the moment to disable the publickey lookup.

@ymecho At the moment I recommend using my fork only for testing purposes but if you really want this feature you are free to use my fork. But keep in mind, I'm only updating the source code - I don't provide ready to use packages.

897221242:
May I ask if the building CVE-2016-20012 has been fixed in the openssh-9.3p2.tar.gz version? Thank you

daztucker:
> May I ask if the building [CVE-2016-20012](https://github.com/advisories/GHSA-84j4-ccmw-hwpg) has been fixed in the openssh-9.3p2.tar.gz version? Thank you

The OpenSSH team does not intend to change anything regarding this behaviour as in our opinion the existing proposals to change it cause more problems than they solve.  It's arguable that whether or not there should even be a CVE since the behaviour is part of the spec.

manfred-kaiser:
> May I ask if the building [CVE-2016-20012](https://github.com/advisories/GHSA-84j4-ccmw-hwpg) has been fixed in the openssh-9.3p2.tar.gz version? Thank you

The behavior is defined in [RFC 4252](https://datatracker.ietf.org/doc/html/rfc4252#section-7). 

> Private keys are often stored in an encrypted form at the client
> host, and the user must supply a passphrase before the signature can
> be generated.  ... To avoid unnecessary processing and user
> interaction, the following message is provided for querying whether
> authentication using the "publickey" method would be acceptable.


According to the RFC this step is optional but **at the moment ALL ssh client implementations require this step**. 

So the proposed change will slightly increase security but the result is, that you don't have any compatible ssh clients, which is a much bigger problem.

@897221242 if you really want to get this "fixed" you should propose a change to RFC 4252 and implement compatible authentication procedures in all major ssh clients and libraries.

ilyagr:
I am far from an expert, but this is quite confusing to me, and I thought this might not be the worst place to ask a question (feel free to redirect me if you have time for that but no time for a longer response)

OpenSSH seems to take quite a bit of effort to prevent similar information being exposed via timing, by making sure that public key verification for a given username takes the same amount of time regardless of whether that user is present on the system:

https://github.com/openssh/openssh-portable/blob/df2b28163ac75e023837de445d6492dc57359105/auth2.c#L348-L350

However, because of the issue this PR discusses, OpenSSH also happily gives the same information for the low, low price of a public key. I can understand how the trade-off might be worth it, but there does seem to be a minor contradiction here. Is this just a historical weirdness; was the timing attack protection implemented before a decision was made that this kind of information gathering is not something OpenSSH should prevent? Or is the goal of that snippet of code something different?

> There is no "user enumeration" here. This feature allows testing whether a particular key may be accepted by a particular account. There is no general way to query which accounts are valid without valid public keys for each one, much less the ability to enumerate all the accounts on a system.

It's quite easy to get a likely public key for any GitHub user. For example, <https://github.com/djmdjm.keys> and <https://github.com/ilyagr.keys> work even if you aren't logged in to GitHub. While these are *GitHub* public keys, many people reuse the same key on most machines they have access to (should they not do that?). My understanding is that OpenSSH is meant to work securely in a world where everyone has a habit of shouting all of their public keys from a rooftop every evening, but feel free to correct me.

(This was also mentioned by the article linked from https://github.com/openssh/openssh-portable/pull/270#issuecomment-920053952)

------

As an aside, the reason I ended up here has to do with thrussh and russh having really draconic defaults of taking 1s to reject any public key to avoid the figuring-out-if-a-user-exists timing attack (so, rejecting 10 keys takes 10s), but russh also allows properly responding to SSH_MSG_USERAUTH_REQUEST (not by default; the default is to always respond PK_OK; I'm guessing you don't look favorably at this). Compared to those projects, OpenSSH pays a much smaller cost to prevent the timing attack IIUC, but there is still a cost. If you have any comment on whether paying a higher price is justified to prevent the timing attack, that would be valuable to me, but this is an OpenSSH project page, so feel free to ignore this paragraph.


----

**Update:** I followed https://www.openwall.com/lists/oss-security/2018/08/24/1, and does actually address most of my questions, assuming the timing protection cost is considered "low".

Most relevant quote from that link:

> We have and will continue to fix bugs like this when we are made aware
> of them and when the costs of doing so aren't too high, but we aren't
> going to get excited about them enough to apply for CVEs or do security
> releases to fix them. The following explains our reasoning.

**Remaining question:** It still seems mildly contradictory to me to implement the timing protections causing others to believe they are useful if there is an easy way to get the same info, but perhaps the timing protections are more useful with other kinds of auth than public key?

manfred-kaiser:
@ilyagr As discussed in this PR, the public key lookup step is defined in [RFC 4252, Section 7](https://datatracker.ietf.org/doc/html/rfc4252#section-7) and virtually all SSH clients depend on it. Disabling it would mean clients have to sign with every key on every attempt, which causes unnecessary passphrase prompts or FIDO2 touch requests, and users will likely react by removing those protections.

After discussing this with @djmdjm, @mfriedl, @daztucker, and Simon Tatham (PuTTY developer), I agree with the OpenSSH team that fixing this would cause more problems than it solves. In my experience, the SHA-1 deprecation already showed how much breakage even a well-announced protocol change can cause, and every SSH client I tested relies on the publickey lookup step.

Regarding the 1-second timeout approach mentioned by thrussh/russh: this doesn't really solve the problem. Users often have multiple keys, and by default up to 6 keys are sent to the server one after another, resulting in up to 6 seconds of delay for a legitimate user. An attacker who already knows the specific public key only needs to test that one key once, so the 1-second penalty hurts legitimate users far more than the attacker. 

On top of that, OpenSSH already has its own penalty mechanisms for failed authentication attempts. Since the attacker never completes a full authentication, these penalties are triggered anyway and slow down or even block the attacker far more effectively than the fixed per-key delay used by thrussh/russh.

The user validity oracle in this case is also a smaller risk compared to the timing attacks, because the attacker needs to know both the username *and* the public key. The timing attacks discussed earlier were about probing whether a username is valid with no prior knowledge at all, which opens up much more dangerous follow-up attacks like targeted password brute-forcing against only confirmed accounts.

Best practice, even if not always followed, is to use separate keys for different purposes: one for GitHub, one for work servers, one for private use, etc. That at least makes it harder to take a known public key and test it against other servers.

manfred-kaiser:
@ilyagr You mentioned the different approaches of thrussh and russh:

> russh also allows properly responding to SSH_MSG_USERAUTH_REQUEST (not by default; the default is to always respond PK_OK)

To clarify: thrussh and russh are Rust SSH libraries used by developers to add SSH functionality to their applications (e.g. russh is used by the Warp terminal), not standalone SSH servers like OpenSSH or Dropbear.

These two libraries handle this differently.

**thrussh** responds to every publickey probe with a constant-time delay (`auth_rejection_time`, default 1s) regardless of whether the key is valid or not. This prevents the timing oracle but means that with multiple configured keys, a legitimate user waits 1 second per key.

**russh** takes a different approach: the default implementation of `auth_publickey_offered` returns `Auth::Accept` unconditionally, as I [verified in the source code](https://github.com/Eugeny/russh/blob/main/russh/src/server/mod.rs), meaning it always responds with PK_OK without checking whether the key is actually valid for the user. Since russh is a library, this permissive default is likely intended as a starting point for developers to override, but any deployment that does not override it ends up always requiring clients to sign with every key.

With russh's default implementation, `auth_publickey` (the actual verification after the signature) returns `Auth::reject()`, meaning authentication will always fail. The client still signs with every configured key and the user may be prompted to unlock passphrase-protected or FIDO-secured keys, only to get rejected each time.

If a developer implements `auth_publickey_offered` correctly but leaves `auth_publickey` at its default, the result is even more confusing: the server signals that the key is acceptable, the user unlocks it and signs, and then gets rejected anyway. This could lead users to assume their keys are broken or to remove protections in an attempt to troubleshoot.

PR #272, opened 2021-09-19
Fix "Allocated port" message for unix sockets

pstray:
A small patch to fix the Allocated port message to be in line of the debug-message a handful of lines earlier in the code.

Currently ssh prints (null) insted of the path of the unix socket, in the message to the user, this PR fixes that


daztucker:
Applied upstream and synced as 8c07170135dde82a26886b600a8bf6fb290b633d.

PR #273, opened 2021-09-21
Fix agent handling of certificates

finlaywashere:
Hello, I encountered an issue with OpenSSH's handling of ssh certificates and storing them in the ssh agent where the certificates would not be stored in the agent even if the key used to authenticate with them is which causes ssh agent forwarding to fail on some systems using ssh certs. Here is a fairly minimal patch that fixes it. Any feedback would be greatly appreciated.

### DIFF
--- a/sshconnect2.c
+++ b/sshconnect2.c
@@ -1620,11 +1620,15 @@ load_identity_file(Identity *id)
 				sshkey_free(private);
 				return NULL;
 			}
-			old_type = private->type;
-			private->type = id->key->type;
+			if(sshkey_type_is_cert(id->key->type) > 0){
+				old_type = private->type;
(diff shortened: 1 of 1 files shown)


finlaywashere:
I'm not really sure why the tests are failing now, the ed25519 cert test is failing even though I can log in successfully to my server using ed25519 w/ certificate. Does anyone know why this could be happening?

finlaywashere:
Any update on if this is going to be merged or needs changes?

djmdjm:
Is this for AddKeysToAgent?

finlaywashere:
Hello, yes this pr was to fix a bug that would cause addkeystoagent to only add the actual key and not the certificate to the agent

PR #276, opened 2021-10-06
update openssh man

Ashwin-droid:
update openssh man


daztucker:
this looks like it was sent in error.  if you do intend sending a pull request, please send one with a small number of commits against the openssh/openssh-portable tree.

PR #278, opened 2021-10-10
Be strict about masklength

khaleesicodes:
strtoul is too lax about characters which are used for numbers.
Manually check that masklength starts with digit.
PoC
Add following match block to your sshd_config
 Match address "127.0.0.1/  +0032"
        ForceCommand ls
If you access your server with `ssh 127.0.0.1` then ls is executed.
This behaviour is unexpected as masklength should not be valid.

Shoutout to [@c3h2_ctf](
https://twitter.com/c3h2_ctf
)


djmdjm:
committed upstream and synced back as fda1eaaec - thanks!

PR #282, opened 2021-10-17
Reduced visibility of addr functions

stoeckmann:
Completely removed addr_sa_pton since it was not called anywhere.


djmdjm:
Merged as 74ff6382f - thanks!

PR #284, opened 2021-10-21
Only free allocated memory on UnixWare

stoeckmann:
If libaif is compiled in together with shadow support, then a
misconfigured system could trigger invalid free calls:

The "spw" variable can be NULL even though a password entry
exists. This can be reproduced on systems with shadow support if
/etc/shadow lacks a user entry which exists in /etc/passwd: Thus
it requires a misconfigured system to trigger this bug.

If "spw" is NULL then get_iaf_password is never called, which
means that "passwd" still points to memory within the pw struct.

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -150,7 +150,8 @@ allowed_user(struct ssh *ssh, struct passwd * pw)
 			locked = 1;
 #endif
 #ifdef USE_LIBIAF
-		free((void *) passwd);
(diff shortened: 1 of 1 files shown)


daztucker:
I've committed the proposed patch as 2923d026e55998133c0f6e5186dca2a3c0fa5ff5.  Thanks for the report.

PR #286, opened 2021-10-22
Adjust error code paths.

stoeckmann:
- Always clear password in RAM, even in error path
- Do not call kill with -1 argument by mistake


daztucker:
Applied upstream, thanks.

PR #287, opened 2021-10-22
Improve sshbuf defensive measures

stoeckmann:
- Keep track of reference count even if parent is set multiple times
- Gracefully handle failed re-allocation in sshbuf_reset

### DIFF
--- a/sshbuf.c
+++ b/sshbuf.c
@@ -109,6 +109,8 @@ sshbuf_set_parent(struct sshbuf *child, struct sshbuf *parent)
 	if ((r = sshbuf_check_sanity(child)) != 0 ||
 	    (r = sshbuf_check_sanity(parent)) != 0)
 		return r;
+	if (child->parent != NULL)
+		child->parent->refcount--;
 	child->parent = parent;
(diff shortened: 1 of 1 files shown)


sshbuf.c
```diff
@@ -109,6 +109,8 @@ sshbuf_set_parent(struct sshbuf *child, struct sshbuf *parent)
 	if ((r = sshbuf_check_sanity(child)) != 0 ||
 	    (r = sshbuf_check_sanity(parent)) != 0)
 		return r;
+	if (child->parent != NULL)
+		child->parent->refcount--;
```
djmdjm, line 113: I don't think it makes sense for a child buffer to be "re-parented".

IMO we it would be better to explicitly disallow this, e.g.

```
        if (child->parent != NULL && child->parent != parent)
                return SSH_ERR_INTERNAL_ERROR;
```

stoeckmann: Yes, in that case it makes sense to treat it as internal error.

djmdjm:
Thanks - these have been applied and will be in OpenSSH 9.1

PR #289, opened 2021-11-12
Fix two signal races in sshd pre-auth 

HerrSpace:
grace_alarm_handler is used as a SIGALRM handler in sshd. It will kill
pmonitor->m_pid, which is meant to be a child of a fork done by sshd,
and then exit.

The  first signal race could have caused the SIGALRM handler to be
called before we save the pid from the fork in pmonitor->m_pid. In this
case we would not have killed our child before exiting.

The second signal race could have caused the SIGALRM handler to be
called while waiting for a blocking waitpid or right after that, before
clearing pmonitor->m_pid. In this case we would potentially have killed
a foreign process, that reused this pid.

Shoutout to @c3h2-ctf and @stoeckmann

### DIFF
--- a/sshd.c
+++ b/sshd.c
@@ -478,6 +478,7 @@ privsep_preauth(struct ssh *ssh)
 {
 	int status, r;
 	pid_t pid;
+	sigset_t oldset, ourset;
 	struct ssh_sandbox *box = NULL;
 
 	/* Set up unprivileged child process to deal with network data */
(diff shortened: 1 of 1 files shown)


HerrSpace:
Gentle reminder regarding this PR. Would somebody be up to review this? :) CC @daztucker as you have looked at PRs in the past here.

daztucker:
I spent some time looking at these, and while I agree your patches fix the problem, I think a better way to fix it is to remove the monitor pid reference from the handler entirely.  Its function is covered by the kill(0, ...), which was added later, since it's in the same process group, and now that privsep is mandatory we no longer have to handle the !privsep case.

daztucker:
The race should be fixed by commit 4e62c13ab419b4b224c8bc6a761e91fcf048012d.  Thanks for the analysis.

PR #291, opened 2021-11-30
Add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which Openssh already runs (https://github.com/google/oss-fuzz/tree/master/projects/openssh)

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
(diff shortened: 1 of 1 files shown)


daztucker:
I don't think this will work since we currently don't allow any third party workflows on the OpenSSH project.

DavidKorczynski:
Okay - I am happy to leave this open in case your policy changes and feel free to close the issue if you prefer

daztucker:
We now allowlist a Cygwin workflow for testing purposes, we could probably extend the allowlist and enable this now.

daztucker:
Applied, thanks.  I'm still fiddling with the allowlist settings.

PR #292, opened 2021-12-21
Use SHA2 when client supports it

beldmit:
Intended to fix #3375


djmdjm:
this was fixed in openssh-8.9

PR #294, opened 2022-01-05
add missing *-sk types to ssh-keyscan manpage

skazi0:


daztucker:
Applied upstream, thanks.

PR #295, opened 2022-01-06
load dsa key last

ProBackup-nl:
The order in which keys are loaded into ssh-agent defines the order in which keys are used in ssh. Thus it might be wiser to load the relative less secure dsa last.


fyfe:
@daztucker @djmdjm 
This change has been applied incorrectly in commit ad16a84e64a8cf1c69c63de3fb9008320a37009c. It's moved RSA to the bottom of the list not DSA; [readconf.c line 2535](https://github.com/openssh/openssh-portable/blob/master/readconf.c#L2534).

daztucker:
Thanks, fixed in 336685d223a59f893faeedf0a562e053fd84058e

PR #296, opened 2022-01-20
ssh-agent: remove all keys upon SIGUSR1..

sdaoden:
With the advent of per-user temporary directories it became
hard for an administrator to remove all keys from all running
ssh-agent instances; what formerly could be done like so

   if command -v ssh-add >/dev/null 2>&1; then
      for a in /tmp/ssh-*/agent.*; do
         [ -e "$a" ] || continue
         act "SSH_AUTH_SOCK=\"$a\" ssh-add -D </dev/null >/dev/null 2>&1 &"
         inc
      done
   fi

has become a major undertaking, especially with even more
containerization.  Being able to remove all keys from all agents
with a single command seems so desirable that it is available in
other agents in the software world.

### DIFF
--- a/ssh-agent.c
+++ b/ssh-agent.c
@@ -189,6 +189,9 @@ static int fingerprint_hash = SSH_FP_HASH_DEFAULT;
 /* Refuse signing of non-SSH messages for web-origin FIDO keys */
 static int restrict_websafe = 1;
 
+/* Request to remove_all_identities() pending */
+static volatile sig_atomic_t received_sigusr1 = 0;
+
 static void
 close_socket(SocketEntry *e)
 {
@@ -915,11 +918,10 @@ process_remove_identity(SocketEntry *e)
 }
 
 static void
-process_remove_all_identities(SocketEntry *e)
+remove_all_identities(void)
 {
 	Identity *id;
 
-	debug2_f("entering");
 	/* Loop over all identities and clear the keys. */
 	for (id = TAILQ_FIRST(&idtab->idlist); id;
 	    id = TAILQ_FIRST(&idtab->idlist)) {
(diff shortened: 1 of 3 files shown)


ssh-agent.c
```diff
@@ -1981,6 +1992,16 @@ cleanup_handler(int sig)
 	_exit(2);
 }
 
+/*ARGSUSED*/
+static void
+remove_all_handler(int sig)
+{
+	if (run_state == RS_POLL)
+		remove_all_identities();
```
daztucker, line 2000: doing work like this in a signal handler is not safe.  In this case, it should be sufficient to write to the sig_atomic_t, then use ppoll instead of poll to ensure it wakes up on receipt of the signal.

sdaoden:
But .. i do not think this is how it will happen the way ssh-agent works?  It is a single threaded program and if it sits on poll and the signal arrives the signal handler executes exclusively as ssh_signal() has ensured. So lets just let it do its job and when the signal handler returns the system call restarts?
And if we are not sitting on poll then just set the flag.  Isn't it safe like this?

sdaoden:
Hm. I now see that ppoll() is always available in openssh.
You -- what you meant was "just use ppoll()", so that SA_RESTART does not come into play for SIGUSR2 at all in ssh-agent. Then just pick up the state in the main loop. Right?
Ok, will do.

sdaoden:
Yes, better.
So, i have rewritten it to use ppoll(2), following style found in sshd.c! Did not know even pselect(2) is gracefully available!
It is more code now.. I have taken the freedom to drop the second reaper() call, as reaper() is called from within prepare_poll() anyway?
I have also added yet a second round of USR1 clearance to the test.
I personally would maybe/likely join the ssh_signal() and the sigaddset() block to one, like this there would be only one (d_flag | D_flag) condition test.

PR #297, opened 2022-01-22
ssh-agent: remove all keys upon SIGUSR1..

sdaoden:
With the advent of per-user temporary directories it became
hard for an administrator to remove all keys from all running
ssh-agent instances; what formerly could be done like so

   if command -v ssh-add >/dev/null 2>&1; then
      for a in /tmp/ssh-*/agent.*; do
         [ -e "$a" ] || continue
         act "SSH_AUTH_SOCK=\"$a\" ssh-add -D </dev/null >/dev/null 2>&1 &"
         inc
      done
   fi

has become a major undertaking, especially with even more
containerization.  Being able to remove all keys from all agents
with a single command seems so desirable that it is available in
other agents in the software world.

### DIFF
--- a/regress/agent.sh
+++ b/regress/agent.sh
@@ -157,30 +157,49 @@ done
 
 ## Deletion tests.
 
+delete_cycle() {
+	# make sure they're gone
+	${SSHADD} -l > /dev/null 2>&1
+	r=$?
+	if [ $r -ne 1 ]; then
+		fail "ssh-add -l returned unexpected exit code: $r"
+	fi
+	trace "readd keys"
+	# re-add keys/certs to agent
+	for t in ${SSH_KEYTYPES}; do
+		${SSHADD} $OBJ/$t-agent-private >/dev/null 2>&1 || \
+			fail "ssh-add failed exit code $?"
+	done
+	# make sure they are there
+	${SSHADD} -l > /dev/null 2>&1
+	r=$?
+	if [ $r -ne 0 ]; then
+		fail "ssh-add -l failed: exit code $r"
+	fi
+}
+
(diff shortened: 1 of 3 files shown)


sdaoden:
Ah. Merde! Sorry, i thought "mergeable" meant the pull request!

sdaoden:
..and one more sorry, this was not meant to be closed. I might become a github expert.

djmdjm:
This looks good to me and the better signal handling could allow us to make `cleanup_handler()` a regular call in future (i.e. no longer calling it in signal context).

Would appreciate @daztucker taking a look at this too

sdaoden:
(It is all taken from sshd.c of course.  Only to mention it.  I .. do not implement this cleanup like it is done in sshd.c in this context here now!?!)
Thank you for looking into this.

sdaoden:
But surely not because of this changeset, no?
Never used CI as it is here; but there is no log at all for the failure, not even a build one?

sdaoden:
(i do not understand this CI-fuzz github thing which fails in one out of dozens of things.  this should have nothing to do with the little patch, i would think.)

sdaoden:
Interestingly "the ppoll part has just recently been committed" almost as stated here.
Please let me allow one sentence: what a crappy attitude to leave that truly desired piece of code laying around for two years.  git allows selective staging via "git add -p", and you would have had it 22 months ago.
Anyhow, this little pull request is now solely the SIGUSR1 handling, and all tests ran.

sdaoden:
P.S.: "you would have had it 22 months ago": within less than 30 seconds.

djmdjm:
a similar change has been committed de644b1831b

PR #298, opened 2022-01-31
sandbox-seccomp-filter: allow gettid

al3xtjames:
Some allocators (such as Scudo) use gettid [while tracing allocations][1].
Allow gettid in preauth to prevent sshd from crashing with Scudo.
[This NixOS issue][2] has some background info.

[1]: https://github.com/llvm/llvm-project/blob/llvmorg-13.0.0/compiler-rt/lib/gwp_asan/common.cpp#L46
[2]: https://github.com/NixOS/nixpkgs/issues/157451


daztucker:
Applied, thanks.  It will be in the next major release.  (I moved it down a few entries to maintain alphabetical order.)

al3xtjames:
Thank you!

PR #299, opened 2022-02-04
Better emulation of old scp behavior

beldmit:
When we use scp in SFTP mode, we try to create a non-existent directory
to better emulate the old SCP behavior

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -1311,7 +1311,7 @@ source_sftp(int argc, char *src, char *targ, struct sftp_conn *conn)
 
 	if (src_is_dir && iamrecursive) {
 		if (upload_dir(conn, src, abs_dst, pflag,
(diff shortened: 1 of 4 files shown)


beldmit:
Red Hat Enterprise Linux 8 series is based on OpenSSH 8.0, which is rather old.

Soon after the OpenSSH 8.0 was published, the upstream introduced a server-side fix for sftp, which changes the use of POSIX realpath (on LInux) to the OpenBSD one with various quirks:
https://github.com/openssh/openssh-portable/commit/569b650f93b561c09c655f83f128e1dfffe74101

If the OpenBSD realpath is in use on the server, the current master successfully copies directories using scp in SFTP mode. Otherwise problems occur.

As RHEL 8 series will definitely be in wide use for several more years, I kindly ask the OpenSSH upstream to consider this patch, which is trying to workaround the server side on the client side.

PR #301, opened 2022-02-14
[WIP] HPE Nonstop (Tandem) port changes for 8.8 code base

rsbeckerca:
This change includes all requirements for a port to HPE NonStop for client functions. A key component of the port is the difference in the ROOT user on the platform, 65536 instead of 0, and ROOT group of 255 vs. 0. This is a high impact change. Other changes are primarily header file arrangements and compiler issues with const to not-const casting.


sftp-client.c
```diff
@@ -1708,7 +1708,12 @@ do_download(struct sftp_conn *conn, const char *remote_path,
 			tv[0].tv_sec = a->atime;
 			tv[1].tv_sec = a->mtime;
 			tv[0].tv_usec = tv[1].tv_usec = 0;
+#ifndef __TANDEM
 			if (utimes(local_path, tv) == -1)
+#else
+			/* NonStop declares utils without const */
```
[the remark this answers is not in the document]
rsbeckerca: Updated thanks. I caught the next one also, which didn't have a comment.

includes.h
```diff
@@ -16,14 +16,41 @@
 #ifndef INCLUDES_H
 #define INCLUDES_H
 
+#ifdef __TANDEM
+/*
+ * This must be done here instead of on the configure command line to
+ * avoid missing the socklen_t declaration in header files.
+ */
+#define _XOPEN_SOURCE_EXTENDED 1
```
daztucker, line 24: Can you do this in configure.ac with a -D flag?  That's how it's done for, eg, hpux:
CPPFLAGS="$CPPFLAGS -D_HPUX_SOURCE -D_XOPEN_SOURCE -D_XOPEN_SOURCE_EXTENDED=1"

rsbeckerca: I would love to oblige. The comment above this line explains that the openssh-portable configure script misidentifies socklen_t and fails to complete. I can post the config.log file that shows this. It seems like a header conflict, which is annoying. I forgot to wrap this in an `#ifdef`, though. Will fix this ASAP.

daztucker: if it's a single test you might be able to copy the socklen_t test into the configure block for Tandem immediately before this.  configure will then cache the result.

rsbeckerca: I have to admit that I do not really know how to pull that off but it is a better solution, I agree

daztucker: If you can post the configure output and config.log someplace I can review and see if I can figure out what's going on.  (If you don't have any better place, maybe also open a bug at bugzilla.mindrot.org and point to this PR).

openbsd-compat/getrrsetbyname.c
```diff
@@ -123,6 +123,14 @@ _getshort(msgp)
 	GETSHORT(u, msgp);
 	return (u);
 }
+#elif HAVE__GETSHORT == 2
```
daztucker, line 126: I think you could do away with the magic value with:

#if defined(HAVE_DECL__GETSHORT) && HAVE_DECL__GETSHORT == 1 && !defined(HAVE__GETSHORT)

rsbeckerca: Don't get me started on this one. I am opening a case with the HPE compiler group because there is something seriously wrong here. The issue is that we have `_getshort()` declared in `/usr/include/resolv.h`, but it appears not to be exported anywhere (I have searched every DLL on both platform variants. If I don't configure HAVE_GETSHORT, I get into a conflicting declaration.

openbsd-compat/getrrsetbyname.c
```diff
@@ -170,11 +186,20 @@ struct dns_response {
...
 };
 
+#ifndef __TANDEM
 static struct dns_response *parse_dns_response(const u_char *, int);
 static struct dns_query *parse_dns_qsection(const u_char *, int,
     const u_char **, int);
 static struct dns_rr *parse_dns_rrsection(const u_char *, int, const u_char **,
     int);
+#else
+/* NonStop compilers fail if the const is present. */
```
daztucker, line 196: can you solve this with
 #define const
at the top of the file?  we do something similar in auth-pam.c to work around API incompatibilities in PAM implementations.

includes.h
```diff
@@ -50,6 +77,11 @@
 # include <paths.h>
 #endif
 
+#ifdef __TANDEM
+/* Declaration missing from header files. */
+int ftruncate(int filedes, off_t length);
```
daztucker, line 82: This one should be checked by configure (AC_CHECK_DECLS or similar) rather than by platform and probably belongs in openbsd-compat/openbsd-compat.h.  There's a couple of other examples in there (eg memmem).

README.platform
```diff
@@ -85,6 +85,24 @@ There is no official registry of 3rd party event numbers, so if this
...
+OpenSSH is primarily useful on the NonStop platform primarily as a client.
+There is already an implementation with the operating system supporting most
+sshd functions. To build on the platform, standard configure rules apply with
+the follwing requirements:
+
+_RLD_LIB_PATH, includes, and --withssl-dir needs to specify the path to
+OpenSSL on the platform. c99 and c11 are recommended. The CPPFLAGS variable
+should include -Wextensions and -DNSIG=32. --disable-utmpx is required. Once
+configured, the following changes are required in config.h because the configure
+script needs help:
```
daztucker, line 99: what does ./config.guess report on this platform?

You can add that to the giant 'case "$host" in' block (alphabetical order please) and start putting your platform-specific stuff in it.

openbsd-compat/getrrsetbyname.c
daztucker: it conflicts because the function is const-ified and the declaration is not?  The #define const hack might solve that?

openbsd-compat/getrrsetbyname.c
rsbeckerca: Technically, yes, but type-safety

rsbeckerca: Technically, yes, but the OCD type-safety of mine has an issue with that.

includes.h
rsbeckerca: The method exists in the common runtime library, but someone forgot to put the external declaration in any header file. Sloppy, I know.

README.platform
rsbeckerca: `nse-tandem-nskJ06` and `nse-tandem-nskL21` depending on the variant. There are other variations as well.

rsbeckerca: That would be good for the required items. I will gather those up and supply what `config.log` shows (as it fails).

includes.h
daztucker: I get that the problem is that the declaration is missing.  My point is that the ifdef should be "is the definition missing" not "is this a Tandem".  To do this you'd add a AC_CHECK_DECLS(ftruncate, ...) and use the slightly odd ifdef structure outlined here: https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Generic-Declarations.html,  That way if any other platform makes the same mistake it'll be covered, and if your stops making it that'll also be covered.

README.platform
daztucker: sounds like a "nse-tandem-*) case would be a good starting point.  It's not going to conflict with anything else and you can always make it more specific if necessary.

channels.c
```diff
@@ -2315,7 +2315,7 @@ channel_post_mux_listener(struct ssh *ssh, Channel *c)
 		close(newsock);
 		return;
 	}
-	if ((euid != 0) && (getuid() != euid)) {
+	if ((euid != ROOT_UID) && (getuid() != euid)) {
```
daztucker, line 2318: this and similar changes will be a headache since they'll constantly conflict with changes as we bring them in from openbsd.  After the 8.9 release I'll see if openbsd would be willing to turn these into functions representing finer grained permissions (can_change_uid(), can_bind_low_ports() or similar).  This would also help for, eg, POSIX capabilities.

audit-linux.c
```diff
@@ -61,7 +61,7 @@ linux_audit_record_event(int uid, const char *username, const char *hostname,
 	 * Do not report error if the error is EPERM and sshd is run as non
 	 * root user.
 	 */
-	if ((rc == -EPERM) && (geteuid() != 0))
+	if ((rc == -EPERM) && (geteuid() != ROOT_UID))
```
daztucker, line 64: this one is in the linux specific auditing code, I don't think it's applicable to you?

openbsd-compat/getrrsetbyname.c
daztucker: I meant inside an ifdef __TANDEM.  In that case the system functions that are not declared const you're not getting any type safety anyway.  This is one of the files that is synced with upstream, so extensive changes to it cause ongoing headaches.  Because of this we prefer to keep changes to such files to a minimum.

So why does the compiler choke on the const anyway?

audit-linux.c
rsbeckerca: It is, the effective UID and initiating UID is definitely a factor on the platform. In older ports, we have had people complaining (and getting caught) for violating this. It is a real and frequent thing on NonStop.

openbsd-compat/getrrsetbyname.c
rsbeckerca: I'm going to be working with the compiler team on this one. It is definitely a bug - a very old one spanning at least 10 years. My hypothesis is that the symbol table has overflowed, but I cannot confirm without access to the code owned by other companies.

audit-linux.c
daztucker: yeah but this specific instance is in Linux specific code and the whole file is inside #if defined(USE_LINUX_AUDIT.  Are you setting that?

includes.h
```diff
@@ -16,14 +16,41 @@
...
 #define _GNU_SOURCE /* activate extra prototypes for glibc */
 #endif
 
+#ifndef __TANDEM
 #include <sys/types.h>
+#endif
 #include <sys/socket.h> /* For CMSG_* */
+#ifdef __TANDEM
+/*
+ * NonStop needs types to come first from socket.h for platform historial
```
daztucker, line 48: historial -> historical.

openbsd-compat/bsd-misc.c
```diff
@@ -143,7 +143,11 @@ utimensat(int fd, const char *path, const struct timespec times[2],
 		return -1;
 	}
 # ifndef HAVE_FUTIMES
+#ifndef __TANDEM
 	return utimes(path, tv);
```
daztucker, line 147: what's the reason for this change?  const again?

Jojo-Schmitz: Yes

README.platform
Jojo-Schmitz: Actually `nsx-tandem-nskL21`. So "ns[ex]-tandem-nsk*" Or even, to cover older and meanwhile outdated systems "ns[rex]-tandem-nsk*"

Jojo-Schmitz: All that is in newer versions of these configure tools. (and since a few years now)

Jojo-Schmitz: Yet we find this issue time and time again.

channels.c
Jojo-Schmitz: Code like this found its way into several open source ports, but indeed tends to break again now and then, as the 0 for root is way too deep in UNIXers genetics

openbsd-compat/getrrsetbyname.c
Jojo-Schmitz: More like 20 years,,,

openbsd-compat/bsd-misc.c
rsbeckerca: Specifically const misinterpretation by the compiler.

audit-linux.c
rsbeckerca: Will revert this one, although would it not make sense to be consistent about the UID across the entire code?

channels.c
rsbeckerca: From a code-review perspective, 0 is a magic number in this case instead of a constant meaning nothing bad happened.

openbsd-compat/getrrsetbyname.c
rsbeckerca: I will look into this particular change. It is limited in scope, so could be done.

openbsd-compat/getrrsetbyname.c
rsbeckerca: It conflicts, but there is no actual implementation for `_getshort` or `_getlong` in the operating system even though it is declared. `/usr/include/resolv.h` is `#included` through a platform header file so these declarations interfere with declaring our own. The existing implementation in `openbsd-compat/getrrsetbyname.c` conflicts with the compiler's POV of what is in `resolv.h`, so we have the DECL. I'm thinking we force #undef HAVE_GETSHORT in config.h and do it the way @daztucker suggests. If would be nice to tell configure to do that for us (suggestions?).

README.platform
rsbeckerca: I don't think we need `nsr` anymore. Those systems are way off support, but I'll keep them in case someone wants to play on a dusty system. I'm not going to maintain `nsr` builds. Nope. :)

rsbeckerca: We need to keep some of the knobs, like `--with-ssl-dir` because we don't know which version one will be used. 1.0.2, 1.1.1, and 3.0.x are all active (although I don't think OpenSSH is on 3.0.x yet).

Jojo-Schmitz: The nsr part most probably is in config.guess and config.sub already, no need to remove it ;-)

openbsd-compat/getrrsetbyname.c
daztucker: actually I think the first stop would be to see if we can remove the "register" upstream since it's obsolescent anyway.

daztucker: and in fact it looks like these definitions came from a different file and were de-registered a long time ago:
https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libc/net/res_comp.c#rev1.12

rsbeckerca: > and in fact it looks like these definitions came from a different file and were de-registered a long time ago: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libc/net/res_comp.c#rev1.12

Should I leave things alone for now? But how did they get here in that case?

daztucker: If it makes things better for you I'd delete them in your tree.  I'll remove them from our tree after the 8.9 release (it's *probably* a low-impact change, but it doesn't fix anything that previously worked, so it can wait).

They aren't in the revision of gerrsetbyname.c that was imported, so it looks like Damien added them to make the port work:
https://github.com/openssh/openssh-portable/commit/d9ec370ac3a27fc9374c4182b7477ec0fc8473f4

rsbeckerca: > If it makes things better for you I'd delete them in your tree. I'll remove them from our tree after the 8.9 release (it's _probably_ a low-impact change, but it doesn't fix anything that previously worked, so it can wait).
> 
> They aren't in the revision of gerrsetbyname.c that was imported, so it looks like Damien added them to make the port work: [d9ec370](https://github.com/openssh/openssh-portable/commit/d9ec370ac3a27fc9374c4182b7477ec0fc8473f4)

Will do... wait... doesn't d9ec370 that violate rule 4? ;)

daztucker: (for context, that was "make the minimum changes to files that we keep in sync with OpenBSD").

Maybe?  On one hand putting _getshort and friends in their own file would make it both files easier to sync, but on the other had you then can't make them static so have potential symbol pollution and higher chance of header collisions similar to yours.

Anyway that commit was nearly 20 years ago, it's possible we've gained some experience about what works for us since then.

includes.h
rsbeckerca: It looks like putting `-D_XOPEN_SOURCE_EXTENDED=1` into `configure.ac` gets past this one. I will commit `configure.ac` separately.

openbsd-compat/getrrsetbyname.c
rsbeckerca: I don't know how `#if defined(HAVE_DECL__GETSHORT) && HAVE_DECL__GETSHORT == 1 && !defined(HAVE__GETSHORT)` works, because the prior block defines the incompatible `_getshort` before this `#if` gets invoked. Putting it first changes the semantics of the if. I'm a bit stuck on this one.

includes.h
rsbeckerca: I'm close to getting this working. I'm not sure I need the AC_CHECK_DECLS but will use it anyway. Somehow, ftruncate is being recognized as being there when it's not, but I'll make it work. Commit coming later.

rsbeckerca: This one is done. Comments?

README.platform
rsbeckerca: This file should now be up to speed.

includes.h
rsbeckerca: Fixed

rsbeckerca:
I have rebased to remove all of the mixed reverts. Requesting another look at this.

rsbeckerca:
@daztucker the code should be in better shape now. I added a separate control in configure.ac to handle the `_getshort()` and `_getlong()` issue. If there is a better place to put the code, expecting so, I can do that too.

rsbeckerca:
@Jojo-Schmitz Found a problem with the build that came up. The `-fPIC` option is not supported on this platform. The platform default is PIC, so there is no option required. As in:
```
c99 -g   -I. -I. -I/usr/local-ssl1.1/include -I/usr/local-ssl1.1/include -I/usr/local/include -Wextensions -WIEEE_float -D_XOPEN_SOURCE -D_XOPEN_SOURCE_EXTENDED=1 -Wnowarn=262,1252 -DNSIG=32 -DSSHDIR=\"/usr/local-ssl1.1/etc\" -D_PATH_SSH_PROGRAM=\"/usr/local-ssl1.1/bin/ssh\" -D_PATH_SSH_ASKPASS_DEFAULT=\"/usr/local-ssl1.1/libexec/ssh-askpass\" -D_PATH_SFTP_SERVER=\"/usr/local-ssl1.1/libexec/sftp-server\" -D_PATH_SSH_KEY_SIGN=\"/usr/local-ssl1.1/libexec/ssh-keysign\" -D_PATH_SSH_PKCS11_HELPER=\"/usr/local-ssl1.1/libexec/ssh-pkcs11-helper\" -D_PATH_SSH_SK_HELPER=\"/usr/local-ssl1.1/libexec/ssh-sk-helper\" -D_PATH_SSH_PIDDIR=\"/usr/local-ssl1.1/etc\" -D_PATH_PRIVSEP_CHROOT_DIR=\"/var/empty\" -DHAVE_CONFIG_H -fPIC -shared -o regress/misc/sk-dummy/sk-dummy.so regress/misc/sk-dummy/sk-dummy.lo regress/misc/sk-dummy/fatal.lo ed25519.lo hash.lo ge25519.lo fe25519.lo sc25519.lo verify.lo \
    -L. -Lopenbsd-compat -lopenbsd-compat -L. -Lopenbsd-compat/ -L/usr/local-ssl1.1/lib -L/usr/local-ssl1.1/lib -L/usr/local/lib  -lcrypto -lrld -lz -lrld
c99: illegal option -- f
```

rsbeckerca:
> @Jojo-Schmitz Found a problem with the build that came up. The `-fPIC` option is not supported on this platform. The platform default is PIC, so there is no option required. As in:
> 
> ```
> c99 -g   -I. -I. -I/usr/local-ssl1.1/include -I/usr/local-ssl1.1/include -I/usr/local/include -Wextensions -WIEEE_float -D_XOPEN_SOURCE -D_XOPEN_SOURCE_EXTENDED=1 -Wnowarn=262,1252 -DNSIG=32 -DSSHDIR=\"/usr/local-ssl1.1/etc\" -D_PATH_SSH_PROGRAM=\"/usr/local-ssl1.1/bin/ssh\" -D_PATH_SSH_ASKPASS_DEFAULT=\"/usr/local-ssl1.1/libexec/ssh-askpass\" -D_PATH_SFTP_SERVER=\"/usr/local-ssl1.1/libexec/sftp-server\" -D_PATH_SSH_KEY_SIGN=\"/usr/local-ssl1.1/libexec/ssh-keysign\" -D_PATH_SSH_PKCS11_HELPER=\"/usr/local-ssl1.1/libexec/ssh-pkcs11-helper\" -D_PATH_SSH_SK_HELPER=\"/usr/local-ssl1.1/libexec/ssh-sk-helper\" -D_PATH_SSH_PIDDIR=\"/usr/local-ssl1.1/etc\" -D_PATH_PRIVSEP_CHROOT_DIR=\"/var/empty\" -DHAVE_CONFIG_H -fPIC -shared -o regress/misc/sk-dummy/sk-dummy.so regress/misc/sk-dummy/sk-dummy.lo regress/misc/sk-dummy/fatal.lo ed25519.lo hash.lo ge25519.lo fe25519.lo sc25519.lo verify.lo \
>     -L. -Lopenbsd-compat -lopenbsd-compat -L. -Lopenbsd-compat/ -L/usr/local-ssl1.1/lib -L/usr/local-ssl1.1/lib -L/usr/local/lib  -lcrypto -lrld -lz -lrld
> c99: illegal option -- f
> ```

@daztucker Any idea how to get past this one? It is hard-coded in `Makefile.in` We just have to remove `-fPIC`. When configure runs, `checking whether -fPIC is accepted... no` is reported.

daztucker:
> @Jojo-Schmitz Found a problem with the build that came up. The `-fPIC` option is not supported on this platform. The platform default is PIC, so there is no option required.

configure sets PICFLAG:
AC_SUBST([PICFLAG])

I think the literal -fPIC in Makefile.in should be $(PICFLAG)

rsbeckerca:
> > @Jojo-Schmitz Found a problem with the build that came up. The `-fPIC` option is not supported on this platform. The platform default is PIC, so there is no option required.
> 
> configure sets PICFLAG: AC_SUBST([PICFLAG])
> 
> I think the literal -fPIC in Makefile.in should be $(PICFLAG)

I'm having more issues with PIC code. The .lo extension is not supported on the platform. That seems to be hard-coded in Makefile.in. Maybe I should just drop the DLL thing for now.

rsbeckerca:
> > > @Jojo-Schmitz Found a problem with the build that came up. The `-fPIC` option is not supported on this platform. The platform default is PIC, so there is no option required.
> > 
> > 
> > configure sets PICFLAG: AC_SUBST([PICFLAG])
> > I think the literal -fPIC in Makefile.in should be $(PICFLAG)
> 
> I'm having more issues with PIC code. The .lo extension is not supported on the platform. That seems to be hard-coded in Makefile.in. Maybe I should just drop the DLL thing for now.

Actually, I raised this one in 2020 as Bug 3205. I don't think this may be solvable at this point.

rsbeckerca:
> > > > @Jojo-Schmitz Found a problem with the build that came up. The `-fPIC` option is not supported on this platform. The platform default is PIC, so there is no option required.
> > > 
> > > 
> > > configure sets PICFLAG: AC_SUBST([PICFLAG])
> > > I think the literal -fPIC in Makefile.in should be $(PICFLAG)
> > 
> > 
> > I'm having more issues with PIC code. The .lo extension is not supported on the platform. That seems to be hard-coded in Makefile.in. Maybe I should just drop the DLL thing for now.
> 
> Actually, I raised this one in 2020 as Bug 3205. I don't think this may be solvable at this point.

I think handling changes to `Makefile.in` with `PICFLAG` and any extension handling should be in a different pull request. While running with DLLs is a nice-to-have, it's not strictly required for the port, so I'd rather do that later. Work for you, @daztucker @Jojo-Schmitz?

daztucker:
> I think handling changes to `Makefile.in` with `PICFLAG` and any extension handling should be in a different pull request. 

All the context is here, I'd just make it a separate commit on this PR.  After the 8.9 release I'll come back and pick up the obvious ones (eg this, the ftruncate one and the _getshort ones).

rsbeckerca:
> > I think handling changes to `Makefile.in` with `PICFLAG` and any extension handling should be in a different pull request.
> 
> All the context is here, I'd just make it a separate commit on this PR. After the 8.9 release I'll come back and pick up the obvious ones (eg this, the ftruncate one and the _getshort ones).

As you wish. Any thought on the library object (`lo`) extension thing? I'd like to set a default in `configure.ac` and somehow reference it in `Makefile.in`. That could close bug 3205 too :)

Jojo-Schmitz:
> [Edit: Nevermind below. The depth of issues associated with using DLLs with this code is a lot of work.]
> 
> @Jojo-Schmitz Found a problem with the build that came up. The `-fPIC` option is not supported on this platform. The platform default is PIC, so there is no option required. As in:

I believe that -fPIC is a "gcc"ìsm

rsbeckerca:
I posted a "not working" change to deal with `-fPIC` and `-shared`. The removal of `-fPIC` works. I think I did something wrong with `SHAREDFLAG` as the value is not being picked up. Could use help on this and file extensions. I think we may have to leave the `.lo` extensions as is and then conditionally change them to `.o` in `Makefile.in` if some other flag is set.

rsbeckerca:
I posted a change to deal with `-fPIC` and `-shared`. The removal of `-fPIC` works. The SHAREDFLAG value is not being picked up from my "tandem" section - don't know why because it should. Thinking about the `.lo` problem, I think we need a separate variable for linkable objects, like `LINK_OBJECT_EXT=lo`, and then override that, then in `Makefile.in`, put individual recipes for each `.lo` objects instead of the implied rule, but until I get the `SHAREDFLAG` sorted out, I'm hesitant to try this.

openbsd-compat/getrrsetbyname.c
daztucker: I've just committed the register removal, might be worth rebasing.

configure.ac
```diff
@@ -1034,9 +1034,10 @@ mips-sony-bsd|mips-sony-newsos4)
 	CPPFLAGS="$CPPFLAGS -Wextensions -D_XOPEN_SOURCE -D_XOPEN_SOURCE_EXTENDED=1 -Wnowarn=262,1252 -DNSIG=32"
 	AC_DEFINE([HAVE_STRUCT_IN6_ADDR], [1], [struct in6_addr actually defined])
 	AC_DEFINE([HAVE_STRUCT_SOCKADDR_IN6], [1], [struct sockaddr_in6 actually defined])
-	# Todo: Fix this and getrsetbyname.c
-	AC_DEFINE([HAVE__GETSHORT], [2], [incorrect hack to be fixed])
-	AC_DEFINE([HAVE__GETLONG], [2], [incorrect hack to be fixed])
+	AC_DEFINE([HAVE__GETSHORT], [1], [decl exists but impl is missing])
```
daztucker, line 1038: this doesn't look right to me.   There are existing checks in configure.ac:
                AC_CHECK_FUNCS([_getshort _getlong])
                AC_CHECK_DECLS([_getshort, _getlong], ...

Based on the comments here, I'd without your changes I'd expect it to set HAVE_DECL__GETSHORT=1, HAVE_DECL__GETLONG=1 and not set HAVE__GETSHORT or HAVE__GETLONG.  Is that what it does?  If so the existing #ifndef HAVE__GETLONG and similar should do the right thing.

rsbeckerca:
| this doesn't look right to me. There are existing checks in configure.ac:
| AC_CHECK_FUNCS([_getshort _getlong])
| AC_CHECK_DECLS([_getshort, _getlong], ...

| Based on the comments here, I'd without your changes I'd expect it to set HAVE_DECL__GETSHORT=1,
| HAVE_DECL__GETLONG=1 and not set HAVE__GETSHORT or HAVE__GETLONG. Is that what it does? If so the existing #ifndef
| HAVE__GETLONG and similar should do the right thing.

What's wrong here is that configure sets these to 1, but the declaration is incompatible with what is in the code, and there is no actual implementation in the operating system - that's a defect, but fixing it will take a very long time - more than a year assuming I can convince the developers - and outside my control. So we need a way to specify a compatible implementation. The existing define structure does not allow that.

Jojo-Schmitz:
There are merge conflicts, so a rebase seems needed

daztucker:
> What's wrong here is that configure sets these to 1, but the declaration in header files is incompatible with what is in the body of code above, and there is no actual implementation in the operating system

Ah, ok.  We might end up doing something like this:
https://github.com/daztucker/openssh-portable/commit/0132056efabc5edb85c3c7105d2fb6dee41843c6

which I think would cover the case you describe here.

rsbeckerca:
> > What's wrong here is that configure sets these to 1, but the declaration in header files is incompatible with what is in the body of code above, and there is no actual implementation in the operating system
> 
> Ah, ok. We might end up doing something like this: [daztucker@0132056](https://github.com/daztucker/openssh-portable/commit/0132056efabc5edb85c3c7105d2fb6dee41843c6)
> 
> which I think would cover the case you describe here.

Will rebase the PR and try to fix this. May be slightly delayed - OOO.

rsbeckerca:
Rebase just done. I'm running a regression cycle and will look at anything broken by the rebase and force-push the branch when done.

rsbeckerca:
> > What's wrong here is that configure sets these to 1, but the declaration in header files is incompatible with what is in the body of code above, and there is no actual implementation in the operating system
> 
> Ah, ok. We might end up doing something like this: [daztucker@0132056](https://github.com/daztucker/openssh-portable/commit/0132056efabc5edb85c3c7105d2fb6dee41843c6)
> 
> which I think would cover the case you describe here.

@daztucker I'm going to give your suggested fix a try. The remaining code is rebased and passing.

rsbeckerca:
Should be ready.

openbsd-compat/getrrsetbyname.c
```diff
@@ -142,6 +141,7 @@ _getlong(msgp)
 	GETLONG(u, msgp);
 	return (u);
 }
+#endif
```
daztucker, line 144: upstream already has this?

openbsd-compat/getrrsetbyname.c
```diff
@@ -110,10 +110,19 @@ struct __res_state _res;
...
 /*
- * Routines to insert/extract short/long's.
+ * If the system doesn't have _getshort/_getlong or that are not exactly what
+ * we need then use local replacements, avoiding name collisions.
  */
+#if !defined(HAVE__GETSHORT) || !defined(HAVE__GETLONG) || \
+    !defined(HAVE_DECL__GETSHORT) || HAVE_DECL__GETSHORT == 0 || \
+    !defined(HAVE_DECL__GETLONG) || HAVE_DECL__GETLONG == 0
+#define _getshort(x) (_ssh_compat_getshort(x))
+#define _getlong(x) (_ssh_compat_getlong(x))
```
daztucker, line 120: this is also in upstream

daztucker (commented):
this doesn't look like it was rebased on master?

rsbeckerca:
@daztucker I'm very certain the rebase was done and this reflects it. I do not know how the above was included in master already - not my doing, but it was, and I'm fine with ignoring that particular file's changes (my updates had a few more comments based on your proposed changes). Either way is fine.

daztucker:
"This branch is 12 commits ahead, 18 commits behind openssh:master."   The last upstream commit it has is b30698662b from 11 Feb, almost a month ago.   It doesn't, for example, have 1b2920e3  or 0132056 (both 25 Feb) which improve getrrsetbyname.

Jojo-Schmitz:
There are even conflicts being reported here

rsbeckerca:
> "This branch is 12 commits ahead, 18 commits behind openssh:master." The last upstream commit it has is [b306986](https://github.com/openssh/openssh-portable/commit/b30698662b862f5397116d23688aac0764e0886e) from 11 Feb, almost a month ago. It doesn't, for example, have [1b2920e](https://github.com/openssh/openssh-portable/commit/1b2920e3b63db2eddebeec7330ffe8b723055573) or [0132056](https://github.com/openssh/openssh-portable/commit/0132056efabc5edb85c3c7105d2fb6dee41843c6) (both 25 Feb) which improve getrrsetbyname.

What happened was that I rebased on top of `8.9` so we could release it. I will rebase or merge again on top of `master` later today. It might be easier on the conflicts to squash part of this down - the `getshort`/`getlong` bits.

daztucker:
> What happened was that I rebased on top of `8.9` [...] which is what I thought was what was expected in the PR.

Changes always go in to Portable master or OpenBSD HEAD.  In some cases we'll cherrypick changes into the release branch (V_8_9 in this case) in case we end up doing a p2 release and for the benefit of vendors putting together patched packages. 

In most cases the differences are minor and it doesn't matter which one a patch is against.  In this case we've already addressed some of the things in this PR in master, so having commits in here that conflict just makes thing harder.  I could rework them to apply, but given this is a platform I can't test on I'd rather have clean, tested patches.

rsbeckerca:
> > What happened was that I rebased on top of `8.9` [...] which is what I thought was what was expected in the PR.
> 
> Changes always go in to Portable master or OpenBSD HEAD. In some cases we'll cherrypick changes into the release branch (V_8_9 in this case) in case we end up doing a p2 release and for the benefit of vendors putting together patched packages.
> 
> In most cases the differences are minor and it doesn't matter which one a patch is against. In this case we've already addressed some of the things in this PR in master, so having commits in here that conflict just makes thing harder. I could rework them to apply, but given this is a platform I can't test on I'd rather have clean, tested patches.

I will make this clean even if I have to rebase manually. Need about 3 hours.

rsbeckerca:
Rebase seems to have worked. I am running a regression test now.

rsbeckerca:
@daztucker Usual set of tests all passed with my OpenSSL 3.0 port on platform. Should be good to go now.

openbsd-compat/bsd-misc.c
daztucker: I'm not going to carry a diff like this to paper over a a compiler bug, especially for such an esoteric platform.  If you want to put the equivalent '-Dconst' in CFLAGS in the tandem block that's fine.

openbsd-compat/getrrsetbyname.c
```diff
@@ -169,12 +169,21 @@ struct dns_response {
 	struct dns_rr		*additional;
 };
 
+#ifdef __TANDEM
+/* Bypass NonStop compiler defect */
+#define const
```
daztucker, line 174: ditto

openbsd-compat/getrrsetbyname.c
```diff
@@ -375,6 +384,11 @@ freerrset(struct rrsetinfo *rrset)
 	free(rrset);
 }
 
+#ifdef __TANDEM
+/* Bypass NonStop compiler defect */
+#define const
```
daztucker, line 389: ditto

openbsd-compat/getrrsetbyname.c
```diff
@@ -553,6 +567,10 @@ parse_dns_rrsection(const u_char *answer, int size, const u_char **cp,
 	return (head);
 }
 
+#ifdef __TANDEM
+#undef const
```
daztucker, line 571: ditto

configure.ac
```diff
@@ -1034,6 +1034,18 @@ mips-sony-bsd|mips-sony-newsos4)
...
 	TEST_MALLOC_OPTIONS="AFGJPRX"
 	;;
+*-tandem-*)
+	CPPFLAGS="$CPPFLAGS -Wextensions -D_XOPEN_SOURCE -D_XOPEN_SOURCE_EXTENDED=1 -Wnowarn=262,1252 -DNSIG=32"
+	AC_DEFINE([HAVE_STRUCT_IN6_ADDR], [1], [struct in6_addr actually defined])
+	AC_DEFINE([HAVE_STRUCT_SOCKADDR_IN6], [1], [struct sockaddr_in6 actually defined])
+	# Todo: Fix this and getrsetbyname.c
+	AC_DEFINE([HAVE__GETSHORT], [2], [incorrect hack to be fixed])
+	AC_DEFINE([HAVE__GETLONG], [2], [incorrect hack to be fixed])
+	AC_CHECK_DECL([ftruncate],
```
daztucker, line 1053: you can remove the ftruncate test, I've added it upstream.

openbsd-compat/openbsd-compat.h
```diff
@@ -65,6 +65,10 @@ int bindresvport_sa(int sd, struct sockaddr *sa);
 void closefrom(int);
 #endif
 
+#if defined(HAVE_DECL_FTRUNCATE) && HAVE_DECL_FTRUNCATE == 0
+int ftruncate(int filedes, off_t length);
```
daztucker, line 69: applied, thanks.

session.c
```diff
@@ -2158,7 +2158,12 @@ session_signal_req(struct ssh *ssh, Session *s)
 
 	debug_f("signal %s, killpg(%ld, %d)", signame, (long)s->pid, sig);
 	temporarily_use_uid(s->pw);
+#ifndef __TANDEM
 	r = killpg(s->pid, sig);
+#else
+	/* NonStop does not implement killpg() */
+	r = kill(s->pid, sig);
```
daztucker, line 2165: this should be done by providing a killpg implementation instead of modifying the main code.  I've just added one, please let me know if it doesn't work.

ssh-agent.c
```diff
@@ -2146,10 +2146,15 @@ main(int ac, char **av)
 	 * a few spare for libc / stack protectors / sanitisers, etc.
 	 */
 #define SSH_AGENT_MIN_FDS (3+1+1+1+4)
+#ifndef __TANDEM
 	if (rlim.rlim_cur < SSH_AGENT_MIN_FDS)
 		fatal("%s: file descriptor rlimit %lld too low (minimum %u)",
 		    __progname, (long long)rlim.rlim_cur, SSH_AGENT_MIN_FDS);
 	maxfds = rlim.rlim_cur - SSH_AGENT_MIN_FDS;
+#else
+	/* NonStop does not implement rlim but there is no configure knob */
```
daztucker, line 2155: can this be done using HAVE_GETRLIMIT or RLIMIT_NOFILE?

sftp-client.c
```diff
@@ -1708,7 +1708,12 @@ do_download(struct sftp_conn *conn, const char *remote_path,
 			tv[0].tv_sec = a->atime;
 			tv[1].tv_sec = a->mtime;
 			tv[0].tv_usec = tv[1].tv_usec = 0;
+#ifndef __TANDEM
 			if (utimes(local_path, tv) == -1)
+#else
+			/* NonStop declares utimes without const */
```
daztucker, line 1714: I think that would also be fixed with CFLAGS="-Dconst=" ?

daztucker (changes requested):

daztucker:
>  I tried -Dconst and it blew up

On most compilers that'll define it as "1".  You need to define it to the null string.

For comparison,  "CFLAGS=-Dconst= ./configure" builds OK on Linux while  "CFLAGS=-Dconst ./configure" can't even compile its first test program.

rsbeckerca:
> > I tried -Dconst and it blew up
> 
> On most compilers that'll define it as "1". You need to define it to the null string.
> 
> For comparison, "CFLAGS=-Dconst= ./configure" builds OK on Linux while "CFLAGS=-Dconst ./configure" can't even compile its first test program.

Sorry, was not clear. I did define `-Dconst=` as empty. That wasn't the issue. It was the mismatch from header files including `stdio.h` that have OS methods aliased and cannot be changed to remove the `const`. It's not a viable workaround. As I said, the position of where to use `#define const` and `#undef const` in the code is critical to making the compile work, and cannot be done across the board as an argument to the compiler. The `#define const` must follow most of the system header files or the compile will fail.

rsbeckerca:
@daztucker I was trying to rebase onto 9.0 and hit a snag. Is there an upgraded dependency on `zlib`? We can configure with 8.9 but not 9.0 because of that.

daztucker:
The only zlib change recently is this one, but it's trivial.  Other than that zlib has not changed for many releases.

commit d1051c0f11a6b749027e26bbeb61b07df4b67e15
Author: djm@openbsd.org <djm@openbsd.org>
Date:   Mon Jan 17 22:56:04 2022 +0000

    upstream: when decompressing zlib compressed packets, use
    
    Z_SYNC_FLUSH instead of Z_PARTIAL_FLUSH as the latter is not actually
    specified as a valid mode for inflate(). There should be no practical change
    in behaviour as the compression side ensures a flush that should make all
    data available to the receiver in all cases.
    
    repoted by lamm AT ibm.com via bz3372; ok markus

Are you using the same version of autoconf as you were in the past?

rsbeckerca:
We are on autoconf 2.69. I think zlib needs something more recent for the latest changes. We're on 1.2.8 of zlibh and stuck there..

daztucker:
zlib 1.2.8 and autoconf 2.69 should both be fine.  From your description it sounds like it's failing while running ./configure?  What's the failure?

rsbeckerca:
Well, more like an ID10T problem. I rearranged in the `-I` sequence and was able to get it to configure. Let's see how this goes. We have to disable `IP_TOS` for some of the older hardware configurations. Will advise.

rsbeckerca:
The rebase at 9.0 was a real mess. I have to reconstruct what is in `getrrsetbyname.c`. This may take some time and I will try to reintegrate the advice above.

PR #302, opened 2022-02-14
Add WinHello support

github-cygwin:
This is a first cut at adding WinHello support to OpenSSH-portable

Right now this depends on libfido 1.10.0 and the fact that libfido2 returns the "clientPin" capability, but not the "uv" capability for WinHello yet.  In theory this should change in libfido2 to the opposite capability setting (no "clientPin", but "uv" = TRUE),
as outlined in https://github.com/Yubico/libfido2/issues/521.


github-cygwin:
Just rebased to latest master

github-cygwin:
Just rebased to latest master

github-cygwin:
Rebased to latest master

sshconnect2.c
```diff
@@ -1262,10 +1262,20 @@ identity_sign(struct identity *id, u_char **sigp, size_t *lenp,
 			goto out;
 		}
 		sign_key = prv;
-		if (sshkey_is_sk(sign_key)) {
+		if (sshkey_is_sk(sign_key) &&
+		    (sign_key->sk_flags & SSH_SK_USER_PRESENCE_REQD)) {
+			fmprintf(stdout, "You may need to touch your "
+				 "authenticator.\n");
```
martelletto, line 1268: Maybe move the `notify_start(..., "Confirm user presence...")` code here?

sshconnect2.c
```diff
@@ -1262,10 +1262,20 @@ identity_sign(struct identity *id, u_char **sigp, size_t *lenp,
...
+		}
+	}
+ retry_pin:
+	if ((r = sshkey_sign(sign_key, sigp, lenp, data, datalen,
+	    alg, options.sk_provider, pin, compat)) != 0) {
+		debug_fr(r, "sshkey_sign");
+		if (pin == NULL && !retried && sshkey_is_sk(sign_key) &&
+		    r == SSH_ERR_KEY_WRONG_PASSPHRASE) {
```
martelletto, line 1276: Changing the PIN prompting logic from "inspect key flags, prompt, try to sign" to "try to sign, look at error, prompt for PIN if needed, retry" is a step in the right direction, IMHO. It will also allow us to kill SSH_SK_USER_VERIFICATION_REQD.

martelletto (changes requested):
Hi! I have left some comments. If you prefer to resume the conversation via e-mail, I am available at pedro@yubico.com. I would also be happy to arrange UV/non-UV keys for testing, if needed. Thank you.

sk-usbhid.c
```diff
@@ -487,6 +504,13 @@ sk_select_by_cred(const fido_dev_info_t *devlist, size_t ndevs,
 		skv[0] = NULL;
 		goto out;
 	}
+	for (i = 0; i < skvcnt; i++) {
+		if (fido_dev_is_winhello (skv[i]->dev)) {
+			sk = skv[i];
+			skv[i] = NULL;
+			goto out;
+		}
+	}
```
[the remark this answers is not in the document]
github-cygwin: I see your point. When I created this patch, the idea was to prefer WinHello for the admin user as well. The reason is this: On Windows since Vista, the admin user is usually not a special account anymore. We have users with administrative privileges, which are by default not present in the user token. If the user wants to perform admin tasks, the user starts a process with admin privileges explicitely, typically in the same session.

What I'm up to here is, without my change, calling ssh in an admin session and calling ssh in a non-admin session, even on the same desktop, would use different UIs.  The admin session would prefer direct device access, the non-admin session is restricted to WinHello.

From a UI perspective, this is a bad interface.

My question here is, how important is ssh-keygen -K? In theory, the user wants to use the token, that's what FIDO2 is about. I'm aware of the potential deficits, but I think a consistent UI is preferrable over easy usage of seldomly used  functionality.

sshconnect2.c
github-cygwin: Thanks, I think so too.  And thanks for reviewing the patches!

sshconnect2.c
github-cygwin: Hmm. I think I did that during hacking this stuff and somehow it didn't quite work as desired... but ATM I don't remember.
Let me check.

github-cygwin: Ok, the problem is that the notification prompt is only required if we're not using WinHello, so we may only emit the notification if the key is not accepted by the first call of sshkey_sign with SSH_ERR_KEY_WRONG_PASSPHRASE, which automatically implies we're not using WinHello. The printout of "You may need to touch your authenticator." is an uninformed hint to the user at this point and not the start of actually requesting user presence.

sk-usbhid.c
```diff
@@ -450,6 +450,15 @@ check_sk_options(fido_dev_t *dev, const char *opt, int *ret)
 		skdebug(__func__, "device is not fido2");
 		return 0;
 	}
+	/*
+	 * Workaround required up to libfido2 1.10.0.  As soon as 1.11.0
+	 * is released and updated in the Cygwin release, we can drop this.
+	 */
+	if (fido_dev_is_winhello(dev) && strcmp (opt, "uv") == 0) {
```
djmdjm, line 457: does this hurt for libfido2 >= 1.11.0? would we need to pull it out later?

sk-usbhid.c
```diff
@@ -896,13 +927,6 @@ sk_enroll(uint32_t alg, const uint8_t *challenge, size_t challenge_len,
 		goto out;
 	}
 	response->flags = flags;
-	if ((flags & SSH_SK_USER_VERIFICATION_REQD)) {
-		if (check_sk_options(sk->dev, "uv", &internal_uv) == 0 &&
-		    internal_uv != -1) {
-			/* user verification handled by token */
-			response->flags &= ~SSH_SK_USER_VERIFICATION_REQD;
-		}
-	}
```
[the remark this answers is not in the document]
djmdjm: IIRC @martelletto added this to reduce friction for biometric keys, what effect will this have for bio keys on non-cygwin platforms? Could it be wrapped in a is_winhello() test?

sshconnect2.c
```diff
@@ -1262,10 +1262,20 @@ identity_sign(struct identity *id, u_char **sigp, size_t *lenp,
 			goto out;
 		}
 		sign_key = prv;
-		if (sshkey_is_sk(sign_key)) {
+		if (sshkey_is_sk(sign_key) &&
+		    (sign_key->sk_flags & SSH_SK_USER_PRESENCE_REQD)) {
```
djmdjm, line 1266: this block goes on to check for SSH_SK_USER_VERIFICATION_REQD keys. Is it possible to have a key that wants UV but not UP? If so, I think it will skip prompting for PIN

sshconnect2.c
djmdjm: Maybe wrap it in `#ifdef CYGWIN`? AFAIK it will produce notifications that are redundant to notify_start() on other platforms. It would be nicer if we could plumb the is_winhello() thing through, but I think that would require an API change to the middleware interface...

sk-usbhid.c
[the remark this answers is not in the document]
github-cygwin: Adding an is_winhello() test is no problem. Will do so in the next iteration of the patchset.

sshconnect2.c
github-cygwin: Yeah, the missing info about the device type on the upper level is a bit of a problem, but nothing we can't cope with.

I can wrap that in a #if, but I was trying to avoid conditional code. Can you check this on another OS to see how this looks like on non-Windows platforms? Maybe it's not that bad...?

sk-usbhid.c
github-cygwin: It wouldn't hurt. But it should go away at one point, IMHO.

sshconnect2.c
github-cygwin: Sorry, but I don't understand your comment. This block does not check for "uv" at all. It just notices the user that the user *may* have to touch the authenticator, because at this point we don't know, and the presence check may be required right in the first call to sshkey_sign, if "uv" isn't required.

github-cygwin:
@djmdjm ping?

sk-usbhid.c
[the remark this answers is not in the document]
djmdjm: I don't think Pedro's proposal yields inconsistent UI. For the default case (no `-Odevice="..."` specified), it will use the WinHello device. A physical token could only be selected if it's an admin **and** they explicitly request a device

sshconnect2.c
djmdjm: If it's non-conditional, then I'd prefer the notification be done using `notify_start()`, as that is the way that all the other presence notifications are handled. In particular it respects the user's `$SSH_ASKPASS_REQUIRE` setting

djmdjm:
I have merged most of this, but added a couple of comments to the remaining part

github-cygwin:
> I have merged most of this, but added a couple of comments to the remaining part

Hi Damien @djmdjm,

I changed the pending two patches per your and @martelletto's input, built and tested it under Cygwin.  Pleae have a look.

Thanks,
Corinna

sk-usbhid.c
```diff
@@ -880,7 +880,7 @@ sk_enroll(uint32_t alg, const uint8_t *challenge, size_t challenge_len,
 	}
 	if (device != NULL)
 		sk = sk_open(device);
-	else
+	else if ((sk = sk_open("windows://hello")) == NULL)
```
djmdjm, line 883: Thinking about this a bit more, maybe the `sk_open("windows://hello")` call should move inside `sk_probe()`? It looks like we only want to open WinHello when we're about to fall back to probing anyway

martelletto: We don't want to `sk_open("windows://hello")` from sk_load_resident_keys(), as the underlying functionality is not yet supported by Microsoft's webauthn.dll. Adding a `prefer_winhello` parameter to sk_probe() would give the function enough context to decide whether to prefer `windows://hello` should `sk_open("windows://hello")` succeed.

djmdjm: What would WinHello's behaviour be in this case? Would it just fail with it as an unsupported operation?

If so, that's IMO okay because the behaviour on Cygwin would be consistent for all uses of FIDO keys: OpenSSH would use WinHello unless explicitly passed a -Odevice to use.

Of course, if WinHello behaves badly when asked to return resident keys then this would be a non-starter :)

sshconnect2.c
```diff
@@ -1266,10 +1266,22 @@ identity_sign(struct identity *id, u_char **sigp, size_t *lenp,
...
+			notifier = notify_start(options.batch_mode,
+			    "You may need to touch your authenticator.");
+			notify_complete(notifier, NULL);
+			notifier = NULL;
```
djmdjm, line 1274: This section will behave badly with SSH_ASKPASS - the dialog will executed and then be killed instantly. The `notify_complete()` call has to happen after `sshkey_sign()` completes

sk-usbhid.c
djmdjm: To be explicit: if this is workable, then `sk_probe()` wouldn't need the `prefer_winhello` option

martelletto: WinHello would cause fido_credman_get_dev_metadata(), fido_credman_get_dev_rp(), and fido_credman_get_dev_rp() (used by read_rks() in sk-usbhid.c) to reliably fail with FIDO_ERR_INTERNAL. While workable, defaulting to WinHello in this case seems incongruous with the way `ssh-keygen -K` works on every other platform.

Users will already need to issue `ssh-keygen -K` from an administrator prompt, but that is a circumstance beyond our control. Asking them to provide a device path (which effectively makes `ssh-keygen -K` dependent on `fido2-token -L` or a similar tool) when we could handle the operation in harmony with the way it is implemented in other platforms, seems like an unnecessary hindrance to me.

To be clear: I have no strong feelings about this. :) Just offering a different perspective. On a side note, it might be worth looking at https://github.com/openssh/openssh-portable/pull/329 for further PIN-prompting improvements (I will ask @LDVG to mark the PR as ready when they return from vacation).

djmdjm:
I just merged something equivalent to the `windows://hello` change in cd06a76b7

djmdjm:
I think that just leaves the `Defer FIDO token PIN prompt when signing the credentials` change to go; please see my last comments on that.

github-cygwin:
> I think that just leaves the `Defer FIDO token PIN prompt when signing the credentials` change to go; please see my last comments on that.

Hi Damien,

thanks for merging this stuff.

In terms of deferring the PIN prompt, after looking at my patch for
quite a while today, I realized it was buggy from the start: I dropped
the notify_complete() call from the retry loop for no apparent reason!
By reverting this, the pieces fall into place by themselves.

Tested on Linux and Cygwin with "up"-only keys as well as "uv/up"-keys.

Please note that the "You may need to touch your authenticator." message
is not targeting WinHello.  It's targeting all other platforms.  Without
it, if the key is a "up"-only key, the first call to sshkey_sign(pin ==
NULL) would require to touch the authenticator without prompting for it.

Please see the latest patch I just pushed into this MR.  I'm pretty sure
it now works as desired.

As an extension, we could replace the "You may need to touch your 
authenticator." message with a "You may need to confirm user presence
for key %s %s" message instead, but that might look a bit over the top
with "uv/up"-keys:

$ ./ssh server
You may need to confirm user presence for key ED25519-SK SHA256:DHNZMpmDM7HQLUgdn6JUgUf6wwuC4DHsnrmXubxfs98
Enter PIN for ED25519-SK key /home/corinna/.ssh/id_ed25519_sk:
Confirm user presence for key ED25519-SK SHA256:DHNZMpmDM7HQLUgdn6JUgUf6wwuC4DHsnrmXubxfs98
User presence confirmed

so I think that's not feasible.  What do you think?


Thanks,
Corinna

github-cygwin:
Hmm, what are those failures on Ubuntu 22.04? I can't see the error log, unfortunately.
I built and tested this on Fedora 36, so I'm pretty sure it works on latest Linux...

djmdjm:
I've pushed a slightly-tweaked version of that last commit as f9648090, so I think that is everything. Please take a look and close the PR if you're happy

daztucker:
> Hmm, what are those failures on Ubuntu 22.04?

I replace the deprecated 18.04 builders with 22.04, but 22.04 made a change to the home directory permissions which tripped up the agent-getpeereid test when it tried to run ssh-add as "nobody".  It was fixed by 5a5c580b48fc6006bdfa731fc2f6d4945c2c0e4e.

> I can't see the error log, unfortunately.

There's nothing sensitive in the logs, but I don't know if it's possible to make the logs world readable.  I had a quick look at the settings and didn't see anything obvious.

Possibly relevant to your interests: I got a Cygwin build working on the Github hosted runners:
https://github.com/openssh/openssh-portable/runs/7913583568?check_suite_focus=true

I'll be adding the "cygwin-release" config shortly, if there's anything else that should be added please let me know.

daztucker:
> I can't see the error log, unfortunately.

If I'm reading https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/using-workflow-run-logs correctly you should be able to see the logs for the public repo as long as you're logged into github itself.  "You must be logged in to a GitHub account to view workflow run information, including for public repositories."

If that's not the case I'll see what we can do.

github-cygwin:
> I've pushed a slightly-tweaked version of that last commit as [f964809](https://github.com/openssh/openssh-portable/commit/f96480906893ed93665df8cdf9065865c51c1475), so I think that is everything. Please take a look and close the PR if you're happy

Hi Damien, I added a comment to
https://github.com/openssh/openssh-portable/commit/f964809

I'm honestly sure that's not exactly how it should work, but please have a look.

Thanks,
Corinna

github-cygwin:
> > I can't see the error log, unfortunately.
> 
> There's nothing sensitive in the logs, but I don't know if it's possible to make the logs world readable. I had a quick look at the settings and didn't see anything obvious.

Actually, I can see the logs.  I'm just not used to working on github, so I expected an error output in the "Details" view of a CI check.  However, this only shows a red line with the text "Error:" and nothing else.  Your reply here got me thinking and I searched for other ways to see the logs and, lo and behold, I finally found the "View raw logs" menu entry in the cogwheel menu on the upper right.  D'oh.  Thanks for nudging me.
 
> Possibly relevant to your interests: I got a Cygwin build working on the Github hosted runners: https://github.com/openssh/openssh-portable/runs/7913583568?check_suite_focus=true
> 
> I'll be adding the "cygwin-release" config shortly, if there's anything else that should be added please let me know.

This is great news, thanks!  I just noticed that the required devel packages are missing to build a full distro OpenSSH package.  It needs all the Kerberos stuff, libedit-devel, libfido2-devel and all their dependencies.  The config options used when building the
distro package with cygport are

CYGCONF_ARGS="--libexecdir=/usr/sbin \
              --with-kerberos5=/usr \
              --with-libedit \
              --with-xauth=/usr/bin/xauth \
              --disable-strip \
              --with-security-key-builtin"


Thanks,
Corinna

github-cygwin:
One of my patches from the start of this endeavor got lost on the way, apparently.
I pushed it again and the commit message tries to explain the problem it solves as
detailed as possible.

I tested this scenario like this:

With master branch:

1. On Linux: ssh-keygen -t ed25519-sk
2. On Linux: ssh-keygen -t ed25519-sk -O verify-required
3. On Cygwin: ssh-keygen -t ed25519-sk
4. On Cygwin: ssh-keygen -t ed25519-sk -O verify-required
  
1: works on Cygwin, works on Linux
2: works on Cygwin, works on Linux
3: works on Cygwin, works on Linux
4: fails on Cygwin, fails on Linux

==> There's no way to use a "verify-required" key enrolled on Cygwin at present

With my patch c842d3568c81:

1. On Linux: ssh-keygen -t ed25519-sk
2. On Linux: ssh-keygen -t ed25519-sk -O verify-required
3. On Cygwin: ssh-keygen -t ed25519-sk
4. On Cygwin: ssh-keygen -t ed25519-sk -O verify-required

1: works on Cygwin, works on Linux
2: works on Cygwin, works on Linux
3: works on Cygwin, works on Linux
4: works on Cygwin, works on Linux

==> The user can now use the same "verify-required" key on Cygwin as well
    as on Linux, independent of the system is has been enrolled on!

martelletto:
Won't c842d3568c811bd5cb29404606c2456f1a680542 cause ssh(1) to prompt for the PIN on tokens capable of built-in UV, like the Yubikey Bio? That was the reason the flag was dropped.

github-cygwin:
> Won't [c842d35](https://github.com/openssh/openssh-portable/commit/c842d3568c811bd5cb29404606c2456f1a680542) cause ssh(1) to prompt for the PIN on tokens capable of built-in UV, like the Yubikey Bio? That was the reason the flag was dropped.

I don't have such a device for testing. I only have a device called
"Yubico.com Yubikey 4/5 OTP+U2F+CCID" in lsusb output.

I only see that "verify-required" keys for this token enrolled under
WinHello works neither under WinHello, nor under Linux. That's very
certainly not as desired.  I'm not even clear on **why** the WinHello-enrolled
key doesn't work under WinHello if the SSH_SK_USER_VERIFICATION_REQD flag
gets removed from the key response.

A user enrolling a key on this token would very much like to use the same
key on different systems, and enrolling a verify-required key should
create identical keys independent of the OS the device has been enrolled
under, or the method used for entering the PIN.

martelletto:
> I don't have such a device for testing. I only have a device called "Yubico.com Yubikey 4/5 OTP+U2F+CCID" in lsusb output.

If you would like one, please ping me at pedro at yubico.com.

> Actually, would it? I mean, if the token is "uv" capable, then the first call to
> sshkey_sign (with NULL pin) would request the PIN entry on the device. As such, the
> loop is as much short-circuited as when using WinHello, isn't it?

I believe ssh-agent(1) would, since the check for `SSH_SK_USER_VERIFICATION_REQD` in `process_sign_request2()` happens before the first call to `sshkey_sign()`: https://github.com/openssh/openssh-portable/blob/master/ssh-agent.c#L810-L834.

github-cygwin:
> > I don't have such a device for testing. I only have a device called "Yubico.com Yubikey 4/5 OTP+U2F+CCID" in lsusb output.
> 
> If you would like one, please ping me at pedro at yubico.com.

Oh, that might come in handy for testing "uv" under WinHello w/ Cygwin.
Thanks for the offer!

> > Actually, would it? I mean, if the token is "uv" capable, then the first call to
> > sshkey_sign (with NULL pin) would request the PIN entry on the device. As such, the
> > loop is as much short-circuited as when using WinHello, isn't it?
> 
> I believe ssh-agent(1) would, since the check for `SSH_SK_USER_VERIFICATION_REQD` in `process_sign_request2()` happens before the first call to `sshkey_sign()`: https://github.com/openssh/openssh-portable/blob/master/ssh-agent.c#L810-L834.

Oh, darn, I completely lost track of ssh-agent.  I always tested by just calling
ssh on the command line without running ssh-agent.

Does anything speak against tweaking ssh-agent accordingly?

github-cygwin:
I force-pushed a less intrusive version of the previous patch.  This patch
now only drops the SSH_SK_USER_VERIFICATION_REQD flag if the device used for
enrolling the key is WinHello.  The downside is that it might not work
entirely correctly with keys on real "uv" devices, only for "clientPin"
scenarios.

@martelletto, your input on this would be highly appreciated, especially
whether or not the commit message makes sense.

I'm already looking into ssh-agent, but this might take a few days.

github-cygwin:
Actually, the patch to ssh-agent was surprisingly simple.  I tested it on
Linux and Cygwin under WinHello, and it looks good on both systems.

github-cygwin:
> I force-pushed a less intrusive version of the previous patch. This patch now only drops the SSH_SK_USER_VERIFICATION_REQD flag if the device used for enrolling the key is [NOT] WinHello.
> [...]
> The downside is that it might not work entirely correctly with keys on real "uv" devices, only for "clientPin" scenarios.

And so it is. I got a Yubikey Bio from Yubikey for testing and it turns out that we
now have the following situation (using Linux as a placeholder for "any non-WinHello scenario"):

```
- If we enroll a verify-required key under WinHello on a standard
  "pinEntry" YubiKey, and if we drop the SSH_SK_USER_VERIFICATION_REQD flag,
  the key is not recognized as a valid key, neither under WinHello, nor
  under Linux.

  Thus we must not drop the SSH_SK_USER_VERIFICATION_REQD flag.
  The resulting key works under WinHello and Linux alike.

- If we enroll a verify-required key under Linux on a "uv" YubiKey Bio
  with fingerprint recognition active, and if we don't drop the
  SSH_SK_USER_VERIFICATION_REQD, the key is not recognized as a valid key
  under WinHello, it only works under Linux.

  Again, not dropping the SSH_SK_USER_VERIFICATION_REQD flag makes the key
  work under both systems.

```
So I rolled back my sk_enroll patch to the previous version, which never drops
the SSH_SK_USER_VERIFICATION_REQD.  I tested this again on Linux and Cygwin, and
the keys work fine on both systems with active "uv" on the Yubikey Bio, too.

> Actually, the patch to ssh-agent was surprisingly simple. I tested it on Linux and Cygwin under WinHello, and it looks good on both systems.

It still looked ok on Cygwin under WinHello, but it has an unfortunate behaviour under Linux: No prompt while waiting for the "uv" input.  So, not good.

The first two patches in the new patchset constitute the bare minimum to make it work with and without ssh-agent, on WInHello and other systems.

I tested them with keys created on Windows and Linux, both in the "clientPin" entry
as well as in the biometric "uv" variation:

```
                Cygwin  w/ Agent  Linux   w/ Agent

Linux Bio       OK      OK        OK      OK

Cygwin Bio      OK      OK        OK      OK

Linux PIN       OK(1)   OK        OK(1)   OK(2)

Cygwin PIN      OK(1)   OK        OK(1)   OK(2)
```

(1) Preceeding "Confirm user presence ..." message. not acted upon
(2) Preceeding "Confirm user presence ..." dialog, shown for just
    a fraction of a second, followed by the real "Enter PIN" dialog.

To get rid of the (1) and (2) downsides, we actually need more patches
with a function call asking the middleware if we're talking to a "clientPin"
or "uv" device.  I added matching patches to allow just that.

@martelletto, any chance you can check if the new sk_test_option functionality
makes sense as is, or if it requires addition stuff?

With this new feature, I also added a fallback mechanism from "uv" to "clientPin"
if "uv" doesn't work for some reason (e. g. biometric reader fails to recognize
user).  This is default in WinHello and I think it's a good idea to support
this in OpenSSH in general.  

Thanks,
Corinna

github-cygwin:
I forgot to add a sk_test_option function to the regress sk-dummy, sorry about that.
Now addded.

martelletto:
Thank you, @github-cygwin. I hope to give the changes a look on Friday or over the weekend. Meanwhile:

> ```
> - If we enroll a verify-required key under WinHello on a standard
>   "pinEntry" YubiKey, and if we drop the SSH_SK_USER_VERIFICATION_REQD flag,
>   the key is not recognized as a valid key, neither under WinHello, nor
>   under Linux.

That sounds a bit strange. Could you try to capture debug output from ssh on Linux?

github-cygwin:
> > - If we enroll a verify-required key under WinHello on a standard
> >   "pinEntry" YubiKey, and if we drop the SSH_SK_USER_VERIFICATION_REQD flag,
> >   the key is not recognized as a valid key, neither under WinHello, nor
> >   under Linux.
> 
> That sounds a bit strange. Could you try to capture debug output from ssh on Linux?

What kind of debug output?  strace or running ssh with SK_DEBUG/DEBUG_SK flags?

martelletto:
Apologies for being unclear. The output of `FIDO_DEBUG=1 ssh -vvv <host>` should suffice.

github-cygwin:
> Apologies for being unclear. The output of `FIDO_DEBUG=1 ssh -vvv <host>` should suffice.

I collected the debug logs.  Two files attached, one of which is the log for
a key enrolled under WinHello with the SSH_SK_USER_VERIFICATION_REQD removed,
as is the current default in the master branch:
[ssh-with-key-enrolled-under-winhello-with-SSH_SK_USER_VERIFICATION_REQD-flag-removed.log](https://github.com/openssh/openssh-portable/files/9463126/ssh-with-key-enrolled-under-winhello-with-SSH_SK_USER_VERIFICATION_REQD-flag-removed.log)
This key fails to work on WinHello as well as on Linux.

For comparison I attached the debug log for a key enrolled under WinHello, but
with the SSH_SK_USER_VERIFICATION_REQD preserved, as in my patchset:
[ssh-with-key-enrolled-under-winhello-with-SSH_SK_USER_VERIFICATION_REQD-flag-preserved.log](https://github.com/openssh/openssh-portable/files/9463167/ssh-with-key-enrolled-under-winhello-with-SSH_SK_USER_VERIFICATION_REQD-flag-preserved.log)
This key works as desired.

HTH,
Corinna

github-cygwin:
Just for reference, this is what WinHello has to say about the key
with removed SSH_SK_USER_VERIFICATION_REQD flag:
image

djmdjm:
I'm going to wait and not do anything until Pedro has had a chance to look at these changes - he's much better at this stuff then I am :)

martelletto:
Thank you @github-cygwin and @djmdjm. I had a look at the patches and feedback.

A problem we will have in accommodating Windows Hello is that, contrary to FIDO's CTAP protocol, [Microsoft's webauthn.h](https://github.com/microsoft/webauthn) does not provide a way for the application to observe the actual capabilities of the token. All we can do is to perform enrollment/authentication operations and deduce the token's capabilities based on the result. That, however, does not allow us to distinguish between PIN and UV, since they look the same (by design) as far as results are concerned.

We have a good example of this difficulty here:

> ```
> - If we enroll a verify-required key under WinHello on a standard
>   "pinEntry" YubiKey, and if we drop the SSH_SK_USER_VERIFICATION_REQD flag,
>   the key is not recognized as a valid key, neither under WinHello, nor
>   under Linux.

This is happening because of our [decision to consider Windows Hello a UV-capable token](https://github.com/Yubico/libfido2/commit/1a77c38). Thinking the underlying token capable of built-in UV, we [drop `SSH_SK_USER_VERIFICATION_REQD`](https://github.com/openssh/openssh-portable/blob/master/sk-usbhid.c#L983-L989) when that does not reflect reality. I suggest not dropping `SSH_SK_USER_VERIFICATION_REQD` if `fido_dev_is_winhello()`.

Overall, I suggest we take a step back and reevaluate the bare minimum needed to support Windows Hello. Once the bits for Windows Hello are all in, we could close this PR and look at improvements to existing functionality in a separate PR.

With that in mind, leaving the double "user presence" prompt aside for a second, is there anything else we need to do to support Windows Hello other than not dropping `SSH_SK_USER_VERIFICATION_REQD` if `fido_dev_is_winhello()`?

-p.

github-cygwin:
Hi Pedro,

> This is happening because of our [decision to consider Windows Hello a UV-capable token](https://github.com/Yubico/libfido2/commit/1a77c38). Thinking the underlying token capable of built-in UV, we [drop `SSH_SK_USER_VERIFICATION_REQD`](https://github.com/openssh/openssh-portable/blob/master/sk-usbhid.c#L983-L989) when that does not reflect reality. I suggest not dropping `SSH_SK_USER_VERIFICATION_REQD` if `fido_dev_is_winhello()`.

But that's missing out on the other problem:

```
- If we enroll a verify-required key under Linux on a "uv" YubiKey Bio
  with fingerprint recognition active, and if we drop the
  SSH_SK_USER_VERIFICATION_REQD, the key is not recognized as a valid key
  under WinHello, it only works under Linux.

  Again, not dropping the SSH_SK_USER_VERIFICATION_REQD flag makes the key
  work under both systems.
```

By leaving that out, you're giving up on interoperability.  A user enrolling
a key under any system won't accept that the key is unusable under another
system.

That's exactly the reason I reverted my patch so SSH_SK_USER_VERIFICATION_REQD
is always preserved.

The key itself is a "verify-user" key.  The key should not enforce **how**
the user verification takes place.

> Overall, I suggest we take a step back and reevaluate the bare minimum needed to support Windows Hello. Once the bits for Windows Hello are all in, we could close this PR and look at improvements to existing functionality in a separate PR.

So you would like to move the sk_test_option stuff entirely into a new PR?
That's ok, but it 's also a pity, becasue that was exactly the stuff I was
looking for input on :flushed: 

> With that in mind, leaving the double "user presence" prompt aside for a second, is there anything else we need to do to support Windows Hello other than not dropping `SSH_SK_USER_VERIFICATION_REQD` if `fido_dev_is_winhello()`?

Yes, see above.  The "don't drop SSH_SK_USER_VERIFICATION_REQD only if fido_dev_is_winhello()" was insufficient as the experiments with Yubikey Bio
clearly shows.  I, and probably all other users using both systems, would really
prefer to be able to use a key independent of the system it has been enrolled on.


Thanks,
Corinna

martelletto:
Hi Corinna,

> But that's missing out on the other problem:
> 
> ```
> - If we enroll a verify-required key under Linux on a "uv" YubiKey Bio
>   with fingerprint recognition active, and if we drop the
>   SSH_SK_USER_VERIFICATION_REQD, the key is not recognized as a valid key
>   under WinHello, it only works under Linux.
> 
>   Again, not dropping the SSH_SK_USER_VERIFICATION_REQD flag makes the key
>   work under both systems.
> ```

Do we know why that is failing? I suspect it is because of this snippet in sk_sign():

```c
	/*
	 * WinHello requests the PIN by default.  Make "uv" request explicit
	 * to allow keys with and without -O verify-required to make sense.
	 */
	if (pin == NULL && fido_dev_is_winhello (sk->dev) &&
	    (r = fido_assert_set_uv(assert, FIDO_OPT_FALSE)) != FIDO_OK) {
		skdebug(__func__, "fido_assert_set_uv: %s", fido_strerr(r));
	}
```

which explains why you want/need `SSH_SK_USER_VERIFICATION_REQD`. Without the flag, this snippet causes Windows Hello to hang with a "this security key does not look familiar" error message on UV-required keys.

My preference would be to drop this snippet, even if it causes unexpected PIN prompts from Windows Hello, and eventually retire `SSH_SK_USER_VERIFICATION_REQD`. It would be unfortunate to need the flag (and the complexity it entails) just because of Windows Hello.

Perhaps there is something we can do. If possible, can you try to patch libfido2's [set_assert_uv() in winhello.c](https://github.com/Yubico/libfido2/blob/main/src/winhello.c#L275-L296) so that the function sets `*out` to `WEBAUTHN_USER_VERIFICATION_REQUIREMENT_ANY` instead of `WEBAUTHN_USER_VERIFICATION_REQUIREMENT_PREFERRED` if `uv == FIDO_OPT_OMIT`?  

> So you would like to move the sk_test_option stuff entirely into a new PR? That's ok, but it 's also a pity, becasue that was exactly the stuff I was looking for input on 😳

Absolutely! My bad. I just wanted to get Windows Hello out of the picture and look at the functionality as a whole. In a nutshell, I agree with the intention, but believe the execution could be simpler. For example, it would be best if we could avoid having code outside sk-usbhid.c care about FIDO concepts such as PIN and UV and FIDO options in general. Something along the lines of:

```c
 retry_sk:                                                                      
        sk_flags = SSH_SK_CHECK_TOKEN;                                          
        if ((r = sshkey_sign(sign_key, NULL, 0, NULL, 0, alg,                   
            options.sk_provider, pin, compat, &sk_flags)) != 0) {               
                /* error */                                                     
        }                                                                       
        if (sk_flags & SSH_SK_NO_TOKEN_FOUND) {                                 
                /* prompt user to insert token */                               
                goto retry_sk;                                                  
        }                                                                       
        if (pin == NULL && (sk_flags & SSH_SK_PIN_NEEDED)) {                    
                /* read pin */                                                  
        }                                                                       
        if (sk_flags & SSH_SK_USER_PRESENCE_NEEDED) {                           
                /* prompt for user presence */                                  
        }
        sk_flags &= ~SSH_SK_CHECK_TOKEN;                  
        if ((r = sshkey_sign(sign_key, sigp, lenp, data, datalen,               
            alg, options.sk_provider, pin, compat)) != 0) {                     
                /* error */                                                     
        }        
```

Does that make sense?

-p.

github-cygwin:
Hi Pedro,

just the first point for now, I have to check into the other stuff later.

> Hi Corinna,
> 
> > But that's missing out on the other problem:
> > ```
> > - If we enroll a verify-required key under Linux on a "uv" YubiKey Bio
> >   with fingerprint recognition active, and if we drop the
> >   SSH_SK_USER_VERIFICATION_REQD, the key is not recognized as a valid key
> >   under WinHello, it only works under Linux.
> > 
> >   Again, not dropping the SSH_SK_USER_VERIFICATION_REQD flag makes the key
> >   work under both systems.
> > ```
> 
> Do we know why that is failing? I suspect it is because of this snippet in sk_sign():
> 
> ```c
> 	/*
> 	 * WinHello requests the PIN by default.  Make "uv" request explicit
> 	 * to allow keys with and without -O verify-required to make sense.
> 	 */
> 	if (pin == NULL && fido_dev_is_winhello (sk->dev) &&
> 	    (r = fido_assert_set_uv(assert, FIDO_OPT_FALSE)) != FIDO_OK) {
> 		skdebug(__func__, "fido_assert_set_uv: %s", fido_strerr(r));
> 	}
> ```
> 
> which explains why you want/need `SSH_SK_USER_VERIFICATION_REQD`. Without the flag, this snippet causes Windows Hello to hang with a "this security key does not look familiar" error message on UV-required keys.

That's very certainly not the reason for this problem.  I didn't test this yet, but it would be very strange.

WinHello defaults to user verification.  The result is that for OpenSSH keys created without "verify-required", WinHello still requests user verification.  On other systems I can create OpenSSH keys which only ask for user presence, on WinHello this is impossible, because of this default.

So this snippet just sets the **default** for user verification to FALSE.  This allows keys created without "verify-required" to be checked for user presence only on WinHello just as well as on any other system.

However, we're talking about "verify-required" keys here only.  For these keys, the next code snippet will do the right thing, independent of the above code snippet:

```
        if (pin == NULL && (flags & SSH_SK_USER_VERIFICATION_REQD)) {
```
Just what we get, no PIN, but the SSH_SK_USER_VERIFICATION_REQD flag is set
```
                if (check_sk_options(sk->dev, "uv", &internal_uv) < 0 ||
                    internal_uv != 1) {
```
Nope, we actually have a device supporting "uv", so we don't return with SSH_SK_ERR_PIN_REQUIRED, but instead...
```
                        skdebug(__func__, "check_sk_options uv");
                        ret = SSH_SK_ERR_PIN_REQUIRED;
                        goto out;
                }
                if ((r = fido_assert_set_uv(assert,
                    FIDO_OPT_TRUE)) != FIDO_OK) {
```
...we set the "uv" flag explicitely to TRUE.
```
                        skdebug(__func__, "fido_assert_set_uv: %s",
                            fido_strerr(r));
                        ret = fidoerr_to_skerr(r);
                        goto out;
                }
        }

```
I don't see how setting the value to FALSE by default changes anything here, because the subsequent code will set the value for "verify-required" keys to TRUE anyway, isn't it?

Thx,
Corinna

github-cygwin:
Oh, wait.  I confused cause and effect!

If we actually **drop** the SSH_SK_ERR_PIN_REQUIRED flag, the condition
```
  if (pin == NULL && (flags & SSH_SK_USER_VERIFICATION_REQD)) {
```
fails, and thus
```
  fido_assert_set_uv(assert, FIDO_OPT_TRUE)
```
is never called.

So if I change
```
        if (pin == NULL && fido_dev_is_winhello (sk->dev) &&
            (r = fido_assert_set_uv(assert, FIDO_OPT_FALSE)) != FIDO_OK) {
                skdebug(__func__, "fido_assert_set_uv: %s", fido_strerr(r));
        }
```
to
```
        if (pin == NULL && fido_dev_is_winhello (sk->dev) &&
            !(flags & SSH_SK_USER_VERIFICATION_REQD) &&
            (r = fido_assert_set_uv(assert, FIDO_OPT_FALSE)) != FIDO_OK) {
                skdebug(__func__, "fido_assert_set_uv: %s", fido_strerr(r));
        }
```
it should work as desired.  This would allow user presence keys to be used as actual user presence keys on WinHello, too, and user verification keys should work, even if we drop the SSH_SK_USER_VERIFICATION_REQD flag for keys enrolled on "uv" devices.

But still, IMHO, the SSH_SK_USER_VERIFICATION_REQD flag is used wrongly.  Its name is confusing, because it **seems* to include keys enrolled on "uv"-capable devices, but in fact it appears to include only keys enrolled on "clientPin" devices.

While the above might fix the problem, I really wonder if it's the right thing to do.

IMHO, a "verify-required" key should not have different flags, just because the device it has been enrolled on has different capabilities. It's the wrong place to make such a decision.

So, yes, it would be great to get rid of the SSH_SK_USER_VERIFICATION_REQD flag in the long run, but wouldn't it be just as well to clean up how it's used as long as we still have it?

Thanks,
Corinna

martelletto:
Hi Corinna,

Your analysis is correct, but do note that it is predicated on `SSH_SK_USER_VERIFICATION_REQD` being set for tokens advertising built-in UV. That is what I meant when I said this is the reason you want/need the flag for Windows Hello.

On a not completely unrelated note, have you had a chance to try `-O no-touch-required` with Windows Hello? Last I checked, [disabling user presence was not supported by webauthn.dll](https://github.com/Yubico/libfido2/blob/main/src/winhello.c#L581-L585). This is bound to create annoying interoperability inconsistencies as well.

-p.

martelletto:
oops, we replied at the same time. Glad we are on the same page. Let me mull it a bit and get back to you; an adequate reply would take more time than I have at the moment. In the meantime, it would probably make sense to look into `-O no-touch-required` as per preceding comment.

github-cygwin:
> Hi Corinna,
> 
> Your analysis is correct, but do note that it is predicated on `SSH_SK_USER_VERIFICATION_REQD` being set for tokens advertising built-in UV. That is what I meant when I said this is the reason you want/need the flag for Windows Hello.

Yes, we need the flag for WinHello, but the problem is, we need it for keys enrolled on any system, not only for keys on WinHello.  The current code breaks interoperability,
because the enrollment decides over the fate of the key, not the signing.  Some decisions can not be performed at signing time for that reason.

> On a not completely unrelated note, have you had a chance to try `-O no-touch-required` with Windows Hello? Last I checked, [disabling user presence was not supported by webauthn.dll](https://github.com/Yubico/libfido2/blob/main/src/winhello.c#L581-L585). This is bound to create annoying interoperability inconsistencies as well.

Yes, I did this right at the start.  no-touch-required keys don't work on WinHello.
I ignored that so far, but we need a solution for this, too, at one point, even if it's just a message telling the user about this fact.

Corinna

github-cygwin:
> So if I change
> [...]
> to
> [...]
> it should work as desired. This would allow user presence keys to be used as actual user presence keys on WinHello, too, and user verification keys should work, even if we drop the SSH_SK_USER_VERIFICATION_REQD flag for keys enrolled on "uv" devices.

On second thought, no, it wouldn't.

If I enroll a "verify-required" key on a "uv" device, the SSH_SK_USER_VERIFICATION_REQD flag will be removed.  Thus, there's no way anymore to distinguish plain keys which only require user presence from keys requiring user verification.  Thus the idea to allow plain user presence keys will be broken for "uv" devices, if the "uv" device defaults to user verification.

So, if I'm not again missing a crucial detail here, that would be another reason never to drop the SSH_SK_USER_VERIFICATION_REQD flag, as outlined in my previous monologue...


Corinna

github-cygwin:
> For example, it would be best if we could avoid having code outside sk-usbhid.c care about FIDO concepts such as PIN and UV and FIDO options in general. Something along the lines of:
> 
> ```
>  retry_sk:                                                                      
>         sk_flags = SSH_SK_CHECK_TOKEN;                                          
>         if ((r = sshkey_sign(sign_key, NULL, 0, NULL, 0, alg,                   
>             options.sk_provider, pin, compat, &sk_flags)) != 0) {               
>                 /* error */                                                     
>         }                                                                       
>         if (sk_flags & SSH_SK_NO_TOKEN_FOUND) {                                 
>                 /* prompt user to insert token */                               
>                 goto retry_sk;                                                  
>         }                                                                       
>         if (pin == NULL && (sk_flags & SSH_SK_PIN_NEEDED)) {                    
>                 /* read pin */                                                  
>         }                                                                       
>         if (sk_flags & SSH_SK_USER_PRESENCE_NEEDED) {                           
>                 /* prompt for user presence */                                  
>         }
>         sk_flags &= ~SSH_SK_CHECK_TOKEN;                  
>         if ((r = sshkey_sign(sign_key, sigp, lenp, data, datalen,               
>             alg, options.sk_provider, pin, compat)) != 0) {                     
>                 /* error */                                                     
>         }        
> ```
> 
> Does that make sense?

It makes sense, and it would really be nice if we could encapsulate the FIDO concepts inside sk-usbhid.c.

I just see a problem there as soon as user prompting is required.  Right now, the UI is solely in the application layer.  The above code is moving prompting into the middleware.  That means, ssh-sk-helper would have to be able to call all the prompting calls like read_passphrase(), notify_start(), etc.  I think that's not feasible, the helper is not UI-aware.

My stance is this: The application layer is responsible for the UI. If the UI gets changed or improved, the middleware should be unaffected.

To do the UI right in terms of hardware tokens, there's no easy way out. The application layer needs **some** information about the current scenario, and that includes to learn what way of user verification should be performed, because that's UI.

Having said that, the way I solved it is not necessarily the best way.  It requires actual knowledge of the FIDO2 protocol and the meaning of strings used therein ("uv", "clientPin").

Either way, the UI basically needs the information
- is some kind of user verification required?
- does the current situation require entering a PIN?

Maybe for a start we should just dumb down the sk_test_option call I added in https://github.com/openssh/openssh-portable/pull/302/commits/99d7032343a0e210452d7b48c95e2919853c76e3, along the lines of
```
  int sk_test_if_we_need_pin_prompting();  /* true == device doesn't support "uv" */
  int sk_test_if_we_allow_pin_prompting(); /* true == device supports "clientPin" */
```
The function names need some polishing, but you get my drift, I think.

Thanks,
Corinna

martelletto:
Hi Corinna,

Quickly chiming in here to keep the conversation going. Please excuse me for my brevity.

> If I enroll a "verify-required" key on a "uv" device, the SSH_SK_USER_VERIFICATION_REQD flag will be removed. Thus, there's no way anymore to distinguish plain keys which only require user presence from keys requiring user verification. Thus the idea to allow plain user presence keys will be broken for "uv" devices, if the "uv" device defaults to user verification.

There is a way to differentiate between keys requiring only user presence from keys requiring user verification without `SSH_SK_USER_VERIFICATION_REQD`, which is to attempt to use the key and check for `FIDO_ERR_PIN_REQUIRED`. This is how the code currently works.

However, as observed in this PR, that is not possible when using Windows Hello, because Windows Hello is not a FIDO2 token, and therefore does not conform to the FIDO2 client-to-authenticator (CTAP) spec. Instead of returning `FIDO_ERR_PIN_REQUIRED`, Windows Hello hangs with a "this security does not look familiar" message.

I am not fundamentally opposed to the reintroduction of `SSH_SK_USER_VERIFICATION_REQD`, but it is important that we are on the same page about what it does and why it is needed. The only reason to reintroduce `SSH_SK_USER_VERIFICATION_REQD` (even if we end up using it in other code paths) is Windows Hello.

Regarding changes to the notification/prompting process:

> I just see a problem there as soon as user prompting is required. Right now, the UI is solely in the application layer. The above code is moving prompting into the middleware. That means, ssh-sk-helper would have to be able to call all the prompting calls like read_passphrase(), notify_start(), etc. I think that's not feasible, the helper is not UI-aware.

If that is how you understood it, then I wasn't clear. My apologies. I did not mean to suggest moving the prompting to the middleware. What I had in mind was having the middleware try to use the key and respond with a set of flags telling ssh/ssh-agent what to prompt for. The prompting would take place where it currently does, it would just be "instructed" by the middleware.

-p.

github-cygwin:
> There is a way to differentiate between keys requiring only user presence from keys requiring user verification without `SSH_SK_USER_VERIFICATION_REQD`, which is to attempt to use the key and check for `FIDO_ERR_PIN_REQUIRED`. This is how the code currently works.
> 
> However, as observed in this PR, that is not possible when using Windows Hello, because Windows Hello is not a FIDO2 token, and therefore does not conform to the FIDO2 client-to-authenticator (CTAP) spec. Instead of returning `FIDO_ERR_PIN_REQUIRED`, Windows Hello hangs with a "this security does not look familiar" message.
> 
> I am not fundamentally opposed to the reintroduction of `SSH_SK_USER_VERIFICATION_REQD`, but it is important that we are on the same page about what it does and why it is needed. The only reason to reintroduce `SSH_SK_USER_VERIFICATION_REQD` (even if we end up using it in other code paths) is Windows Hello.

Ok, I think I got that now.  Bummer that WinHello is not working conformant in this single area, it could have been so easy, oh well. Unless... there isn't a way in WinHello to enforce that the call returns immediately, by any chance?

So, yeah, I accept that we need the SSH_SK_USER_VERIFICATION_REQD flag for WinHello only, but apparently we need it yet.

How do we proceed?  I guess in the light of the above, we can push the first two patches, https://github.com/openssh/openssh-portable/pull/302/commits/baa2928a67d3e931799e322aa25eb03fb8241b25 and https://github.com/openssh/openssh-portable/pull/302/commits/6b3c62d58645c18318b6ceff8a485ced0feba1f5 now?

Also, https://github.com/openssh/openssh-portable/pull/302/commits/2f2b0b1d758deb9e834d8e0c5a582eab33fbaffa should be obvious, right?  @djmdjm?

> Regarding changes to the notification/prompting process:
> 
> > I just see a problem there as soon as user prompting is required. Right now, the UI is solely in the application layer. The above code is moving prompting into the middleware. That means, ssh-sk-helper would have to be able to call all the prompting calls like read_passphrase(), notify_start(), etc. I think that's not feasible, the helper is not UI-aware.
> 
> If that is how you understood it, then I wasn't clear. My apologies. I did not mean to suggest moving the prompting to the middleware. What I had in mind was having the middleware try to use the key and respond with a set of flags telling ssh/ssh-agent what to prompt for. The prompting would take place where it currently does, it would just be "instructed" by the middleware.

I'm pretty sure I don't understand exactly how to do that.  Look at https://github.com/openssh/openssh-portable/pull/302/commits/7262b1027e6a87e3d6e19d427cb28c6d356b0246, please.  I rearranged prompting again, to be able to generate the correct user information in the UI **before** calling sshkey_sign.

For instance if the device is a "uv" device, how would you be able to do this inside the sshkey_sign call itself?  The first call might already succeed and you're too late to do **any* prompting.

How you would do that without giving the UI at least **some** information prior to the sshkey_sign call, some kind of call allowing the UI to ask the middleware what is supposed to happen next.  That's where my sshsk_test_option() call comes into play.  Apart from stylistic changes, I'm really unsure how to do this differently, let alone, encapsulated inside sshkey_sign.

BTW., I'm more or less offline the next 4 weeks, starting this weekend.


Corinna

github-cygwin:
just rebased...

sk-usbhid.c
```diff
@@ -1149,6 +1152,33 @@ check_sign_load_resident_options(struct sk_option **options, char **devicep)
...
 
+int
+sk_test_option(const char *test_option)
+{
+	struct sk_usbhid *sk = NULL;
+	int ret = SSH_SK_ERR_GENERAL, internal_uv;
+
+	fido_init(SSH_FIDO_INIT_ARG);
+
+	sk = sk_probe(NULL, NULL, 0, 0);
```
djmdjm, line 1163: I don't think this will do the right thing when multiple devices are present, which is a case we've worked fairly hard to support. Given multiple devices IMO "device supports feature X" is only useful the presence of a key handle that refers to that device.

Maybe a better way of supporting this would be to add another per-flag, say `SSH_SK_UV_PROVIDED` that is set when the token provides user verification internally. We could set this for biometric keys, etc. Would this help?

I.e. putting something like this in `sk_enroll()`

```
if (check_sk_options(sk->dev, "uv", &internal_uv) == 0 && internal_uv != -1) {
        /* user verification handled by token */
        response->flags &= ~SSH_SK_UV_PROVIDED;
}
```

(yes, this looks a lot like the code that I just removed via Corinna's change)

github-cygwin:
just rebased onto current master

github-cygwin:
The Fuzzing check failed with a weird DWARF error, followed by
ssh-sk.o: in function `sshsk_open':
ssh-sk.c:(.text.sshsk_open[sshsk_open]+0x354): undefined reference to `ssh_sk_test_option'

I built and ran the testsuite successfully, so I'm puzzeled why this fails in the fuzzing test.  What's missing?

PR #303, opened 2022-02-16
prevent deprecated DSA keys from being generated by ssh-keygen -A

jsegitz:


daztucker (commented):
Since they're not currently used by default, generating them and leaving them unused on disk is not a big deal.  Removing them from the default set means that anyone wanting to use DSA keys on a new install has an extra step, but maybe that's a Feature not a Bug.

The man page says "For each of the key types (rsa, dsa, ecdsa and ed25519) for which
host keys do not exist," so that would also need updating.

jsegitz:
That would be feature to me. 

I'll update the man page, thanks for the comment

daztucker:
Applied upstream, thanks.

PR #304, opened 2022-02-23
Specify "remote port" when timeouts are logged

MatthewCroughan:
When timeouts are logged by sshd, it does not specify that the port
number being logged is in fact a remote port, not a local one. This can
be a source of confusion and ambiguity. This commit removes this
ambiguity by specifying that this port is a remote one.

An example of this confusion occurring in the wild can be found here:
https://unix.stackexchange.com/questions/32090/why-logs-show-a-different-ssh-connection-port-than-the-one-specified

Note that this may break log parsers, if they incorrectly assumed the OpenSSH log format was a stable API of some sort.

### DIFF
--- a/sshd.c
+++ b/sshd.c
@@ -367,7 +367,7 @@ grace_alarm_handler(int sig)
 	}
 
 	/* Log error and exit. */
-	sigdie("Timeout before authentication for %s port %d",
+	sigdie("Timeout before authentication for %s on remote port %d",
 	    ssh_remote_ipaddr(the_active_state),
 	    ssh_remote_port(the_active_state));
 }


MatthewCroughan:
@djmdjm any chance of a review?

daztucker:
The address in the log message is also remote, that together with the port identify the remote connection*.   I don't see your proposal as any clearer since it could be interpreted as local address and remote port.

* strictly speaking you'd also need the local address and port to uniquely identify a connection, but that would only matter in the non-default case where sshd was listening on more than one address or port.

Butsabaphon:
ขออนุญาตเข้าใช้งานด้วยค่ะ

Butsabaphon:
[Uploading Building_a_Trusted_Ecosystem_for_Millions_of_Apps_A_Threat_Analysis_of_Sideloading.pdf…]()

djmdjm:
Logging "[address] port [port]" is consistent with how the server logs other error conditions. IMO it's the most logical way to do it, since the local port is almost always fixed at 22 and, even when it is not, it's usually easily discoverable. Contrast with the remote port, which is usually ephemeral and is likely to be needed if you ever have to correlate sshd logs with those from another source (firewall, netflow, etc). Therefore I don't think we should change this.

PR #305, opened 2022-03-07
Document tokens accepted by ProxyJump keyword

pallxk:
`ProxyJump` translates to `ProxyCommand` under the hood and it accepts the same tokens as `ProxyCommand` if I understand it correctly.


daztucker:
Applied upstream, thanks.

PR #306, opened 2022-03-09
ssh-keygen: fix unsigned integer overflows in `-M generate` subcommand

int-e:
The purpose of this command is to produce candidate safe primes for DH key exchange. They are screened in a second phase, so bugs here are harmless as long as enough viable candidates remain.

Experimentally, while generating 16384 bit safe primes, the overflows cause about 20% of the candidates to be lost, and a small number of extra candidates to be included.

**Note** that it might be saner to rewrite the code using 64 bit integers instead.

The experiment I did can be found at https://gist.github.com/int-e/d9994fa61f08fcc775db2468de8940a4

### DIFF
--- a/moduli.c
+++ b/moduli.c
@@ -197,42 +197,76 @@ sieve_large(u_int32_t s)
 	else
 		u = s - r; /* largebase+u is first entry divisible by s */
 
-	if (u < largebits * 2) {
+	/*
+	 * Note that s and u use essentially the whole 32 bit range of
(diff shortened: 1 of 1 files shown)


djmdjm:
I'll take a look at this, but I suspect it will be simpler to use 64 bit math here :)

djmdjm:
Fixed upstream and synced back as 0bc6b4c8f04e29257 by conversion to 64-bit math. Thanks for the report!

PR #308, opened 2022-03-29
ssh: accept SSH_CONFIG_FILE as a default for -F

grahamc:
Treat the environment variable SSH_CONFIG_FILE as a default
value for the `-F` argument.

This supports my use case of multiple projects and repositories
which each have significantly different SSH configurations, and
a wide enough variety of projects and repositories that it isn't
necessarily feasible to construct a per-user (~/.ssh/config)
file with Match blocks.

### DIFF
--- a/regress/sshcfgparse.sh
+++ b/regress/sshcfgparse.sh
@@ -34,6 +34,12 @@ verbose "reparse minimal config"
  ${SSH} -G -F $OBJ/ssh_config.1 somehost >$OBJ/ssh_config.2 &&
  diff $OBJ/ssh_config.1 $OBJ/ssh_config.2) || fail "failed to reparse minimal"
 
+verbose 'verify -F is equivalent to $SSH_CONFIG_FILE'
+(${SSH} -G -F $OBJ/ssh_config somehost >$OBJ/ssh_config.1 &&
+ SSH_CONFIG_FILE=$OBJ/ssh_config.1 ${SSH} -G somehost >$OBJ/ssh_config.2 &&
+ diff $OBJ/ssh_config.1 $OBJ/ssh_config.2) \
+    || fail "difference between -F and SSH_CONFIG_FILE"
+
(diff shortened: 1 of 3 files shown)


djmdjm:
I've thought about this a bit and I don't think that we want ssh to load a different config file based on its environment - ability to modify or override ssh_config is tantamount to at least local code execution (via `LocalCommand` and others), so we want to keep the config resolution as simple as possible there.

An additional, lesser problem relates to the variable `$SSH_CONFIG` specifically - this seems very close to a variable name that people might have already used in scripts, etc. Indeed, a search of a large codebase that I have at hand found one such example, though it looks like it would have worked by accident.

Being able to vary configuration by environment variable does seem like a useful feature though. Obviously, it's possible to do this via wrapping `ssh` with a script or alias, but that's not particularly satisfying. So I was thinking that maybe we could expand the existing `Include` directory to accept `${env}` expansions. E.g. putting this at the start of a config file:

```
Include "~/.ssh/ssh_config-${HOSTNAME}"
```

would allow fairly easy variation of configuration across a fleet.

Some other thoughts:

1. What to do if the environment variable is not set? IMO this should cause the `Include` directive to be ignored
2. Currently environment expansion is performed by `misc.c:vdollar_percent_expand()` and @daztucker reminds me that this will fatal if the environment is missing. We'd need to arrange for it to return an error instead.
3. While we're in there, maybe `Include` should accept some %-expansions like the other directives (e.g. `%h` expanding to the remote hostname)? This would be simple if we're using `dollar_percent_expand()`,

Would doing this by `Include` solve your use case?

cole-h:
I talked with Graham yesterday, and we agreed that that would work for us.

I've attempted to implement your suggestions in https://github.com/openssh/openssh-portable/pull/319.

grahamc:
Closing in favor of #319, thanks for taking that forward Cole!

PR #310, opened 2022-04-09
ssh-keygen: check if resident -sk key exists and prompt before overwrite

martelletto:
When enrolling a resident key on a security token, check if a credential with matching application and user ID strings already exists. If so, prompt the user for confirmation before overwriting the credential.

CC @djmdjm @mfriedl


ssh-keygen.c
```diff
@@ -3186,6 +3186,24 @@ save_attestation(struct sshbuf *attest, const char *path)
...
+	printf("A resident key scoped to '%s' with user id '%s' already "
+	    "exists.\n", application == NULL ? "ssh:" : application,
+	    user == NULL ? "null" : user);
+	printf("Overwrite key in token (y/n)? ");
+	fflush(stdout);
+	if (fgets(yesno, sizeof(yesno), stdin) == NULL)
+		return 0;
+	if (yesno[0] != 'y' && yesno[0] != 'Y')
+		return 0;
+	printf("Touch your authenticator to authorize key generation.\n");
```
djmdjm, line 3203: I think this should not be here, but handled below in the for loop

ssh-keygen.c
```diff
@@ -3773,6 +3791,13 @@ main(int argc, char **argv)
 			    &private, attest);
 			if (r == 0)
 				break;
+			if (r == SSH_ERR_KEY_BAD_PERMISSIONS &&
+			    (sk_flags & SSH_SK_RESIDENT_KEY) != 0 &&
+			    (sk_flags & SSH_SK_FORCE_OPERATION) == 0 &&
+			    confirm_sk_overwrite(sk_application, sk_user)) {
+				sk_flags |= SSH_SK_FORCE_OPERATION;
+				continue;
```
djmdjm, line 3799: make remove the continue and make the next statement an `else if` to reuse the below printf reminder?

I notice that this block also messes up the use of `i` as the number of incorrect passphrase prompts - maybe we need to move the `i++` from the `for()` loop to the actual case where the password was incorrect?

ssh-keygen.c
martelletto: fixed in 4baef88; thanks!

martelletto:
rebased on top of master; no functional change

djmdjm:
committed at last :) - thanks again

PR #311, opened 2022-04-09
sk-api: separate flags used for FIDO signature verification from sk-api flags

martelletto:
Use separate flags to reduce the risk of future confusion/conflict between WebAuthn/FIDO and sk-api flags (which are similar in concept, but serve a fundamentally different purpose).


djmdjm:
Good idea, but I think keeping the existing flags with the same values is a recipe for future confusion. IMO we should renumber the `SSH_SK_USER_*_REQD` flags at the same time to make sure we have caught all the cases

PR #313, opened 2022-04-11
new sRouteCookieId option for FreeBSD proposal.

devnexen:
FreeBSD via the SO_USER_COOKIE socket option allows to attribute
an ID for routing purpose.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -247,6 +247,22 @@ set_rdomain(int fd, const char *name)
 }
 
 int
+set_cookieid(int fd, u_int32_t route_cookieid)
+{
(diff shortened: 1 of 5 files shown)


djmdjm:
We don't generally add support for OS-specific sockopts unless they exist to implement something that has wide OS support (e.g. routing domains).

Instead we recommend that users use a ProxyCommand to get at these, which is a bit of a hassle but it means that OpenSSH doesn't have to carry to union of all interesting sockopts from all supported platforms.

devnexen:
3 oses so far ut I get your point.

PR #317, opened 2022-05-09
Remove duplicate bcrypt_pbkdf.o in object file list.

tobhe:
As the title says, found while checking out the Makefiles.


daztucker:
Applied, thanks.

PR #318, opened 2022-05-11
ssh-add: Skip PKCS11 pin prompt with TEE identity

varder:
TEE Identity-based authentication provides functionality to
log in without a pin but using a User or Group identity.
The feature is valuable for embedded devices where there is no
user interaction.

With the TEE Identity authentication, the pin should be empty.

The use case is:
CKTEEC_LOGIN_TYPE=user ssh-add -s /usr/lib/libckteec.so.0

For TEE Identity-based auth pin should be provided as an
empty string. But in the current implementation, if a pin
is empty the message structure will not be populated with
the pin(see sshbuf_put_string). As a result, the error:
"pin required". As a solution add a new line character.

The details about the TEE Identity-based authentication:
https://github.com/OP-TEE/optee_os/pull/4222

Signed-off-by: Valerii Chubar <valerii_chubar@epam.com>

### DIFF
--- a/ssh-add.c
+++ b/ssh-add.c
@@ -449,14 +449,15 @@ update_card(int agent_fd, int add, const char *id, int qflag,
 {
 	char *pin = NULL;
 	int r, ret = -1;
+	int tee_identity = getenv("CKTEEC_LOGIN_TYPE") != NULL;
 
-	if (add) {
+	if (!tee_identity && add) {
 		if ((pin = read_passphrase("Enter passphrase for PKCS#11: ",
 		    RP_ALLOW_STDIN)) == NULL)
 			return -1;
 	}
 
-	if ((r = ssh_update_card(agent_fd, add, id, pin == NULL ? "" : pin,
+	if ((r = ssh_update_card(agent_fd, add, id, tee_identity ? "\n" : pin == NULL ? "" : pin,
 	    lifetime, confirm, dest_constraints, ndest_constraints)) == 0) {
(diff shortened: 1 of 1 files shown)


djmdjm:
A couple of comments - I think it would be better to control this using a command-line flag than an environment variable. There are at least other cases where users are likely to want to use PKCS#11-hosted keys without providing a PIN at `ssh-add` time:

1. Devices that don't require a PIN at all (though I don't know whether/how PKCS#11 supports these generally)
2. Devices that implement `CKF_PROTECTED_AUTHENTICATION_PATH` (e.g. ones that do not require a PIN or have on-device PIN entry)
3. Deferring PIN entry to time of key use, with the user being prompted via `ssh-askpass`. We've done something similar recently for FIDO keys in 39d17e189 and IMO it makes sense to do the same for PKCS#11 keys.

So I think we could add a flag to `ssh-add` that suppresses the PIN prompt when adding PKCS#11 keys. However, this isn't likely to be enough to make TEE keys work, since the `ssh-pkcs11.c` code still requires a PIN for all devices that don't set the `CKF_PROTECTED_AUTHENTICATION_PATH` flag.

I think the correct solution is for the TEE PKCS#11 provider to set this flag, the [specification](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc385057915) seems to indicate that it is the correct way to implement keys that don't require PIN authentication.

PR #321, opened 2022-06-03
ssh-copy-id: replace ssh port argument with sftp port argument when a custom port was used

Anthrazz:


kalvdans (approved):
It's an improvement, and sed is already a requirement in order to run ssh-copy-id.

PR #323, opened 2022-06-10
Don't propose disallowed algorithms during hostkey negotiation

ZoltanFridrich:
I would like to propose a patch from rhel-9.1.0

Context:
In rhel-9.1.0 when I try to connect to host that only presents ssh-rsa host keys, the connection fails with a very cryptic error message:
`ssh_dispatch_run_fatal: Connection to 10.19.208.80 port 22: error in libcrypto`
I believe this happens because ssh-rsa is disallowed by crypto-policies but openssh disregards the configuration when negotiating hostkey algorithm. Therefore the ssh-rsa hostkey alg is negotiated and the connection fails when openssl returns an error somewhere inside sshkey_verify function.

This patch restricts the range of negotiated algorithms to only those that are allowed by the configuration. Effectively making the connection fail during hostkey negotiation and with a much more descriptive error message.

Signed-off-by: Zoltan Fridrich <zfridric@redhat.com>

### DIFF
--- a/sshconnect2.c
+++ b/sshconnect2.c
@@ -218,6 +218,7 @@ ssh_kex2(struct ssh *ssh, char *host, struct sockaddr *hostaddr, u_short port,
 {
 	char *myproposal[PROPOSAL_MAX] = { KEX_CLIENT };
 	char *s, *all_key;
+	char *hostkeyalgs = NULL, *pkalg = NULL;
 	char *prop_kex = NULL, *prop_enc = NULL, *prop_hostkey = NULL;
 	int r, use_known_hosts_order = 0;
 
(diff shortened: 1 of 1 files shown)


sshconnect2.c
```diff
@@ -253,16 +253,21 @@ ssh_kex2(struct ssh *ssh, char *host, struct sockaddr *hostaddr, u_short port,
 	    (char *)compression_alg_list(options.compression);
 	myproposal[PROPOSAL_MAC_ALGS_CTOS] =
 	    myproposal[PROPOSAL_MAC_ALGS_STOC] = options.macs;
+
 	if (use_known_hosts_order) {
 		/* Query known_hosts and prefer algorithms that appear there */
-		myproposal[PROPOSAL_SERVER_HOST_KEY_ALGS] =
-		    compat_pkalg_proposal(ssh,
-		    order_hostkeyalgs(host, hostaddr, port, cinfo));
```
ZoltanFridrich, line 260: This patch also fixes a memory leak where allocated pointer returned from order_hostkeyalgs() was being lost without freeing.

ZoltanFridrich: #324 will need to be resolved first.

sshconnect2.c
```diff
@@ -255,14 +256,19 @@ ssh_kex2(struct ssh *ssh, char *host, struct sockaddr *hostaddr, u_short port,
 	    myproposal[PROPOSAL_MAC_ALGS_STOC] = options.macs;
 	if (use_known_hosts_order) {
 		/* Query known_hosts and prefer algorithms that appear there */
-		myproposal[PROPOSAL_SERVER_HOST_KEY_ALGS] = prop_hostkey =
-		    compat_pkalg_proposal(ssh,
-		    order_hostkeyalgs(host, hostaddr, port, cinfo));
```
ZoltanFridrich, line 260: string returned from order_hostkeyalgs() is lost here

ZoltanFridrich:
The test failures seem expected judging by the output. Basically the negotiation fails because ssh-rsa is not accepted_algo

sshconnect2.c
```diff
@@ -255,14 +256,19 @@ ssh_kex2(struct ssh *ssh, char *host, struct sockaddr *hostaddr, u_short port,
...
+		if ((hostkeyalgs = order_hostkeyalgs(host, hostaddr, port, cinfo)) == NULL)
+			fatal_f("order_hostkeyalgs");
+		pkalg = match_filter_allowlist(hostkeyalgs, options.pubkey_accepted_algos);
```
djmdjm, line 261: This isn't correct: `options.pubkey_accepted_algos` is the `PubkeyAcceptedAlgorithms` configuration option. That is for controlling which signature algorithms are used for user authentication, whereas this is host authentication.

The configuration option that you're looking for is `HostkeyAlgorithms`, but this is already applied inside `order_hostkeyalgs()`

djmdjm:
I added a comment on the change, but this isn't correct. We already have `HostkeyAlgorithms` to control which signature algorithms are used for hostkey negotiation. Your patch instead adds filtering by `PubkeyAcceptedAlgorithms`, which is used for user authentication.

Please just use `HostkeyAlgorithms` for what you're trying to achieve.

Jakuje:
Sorry for being late to the party. To clarify, this issue is related to the previously reported issue https://bugzilla.mindrot.org/show_bug.cgi?id=2924 where we wanted to handle this as part of the `HostkeyAlgorithms` configuration option but it was ignored so this was the second best thing to avoid hard to debug issues, introducing new options.

PR #324, opened 2022-06-10
Fix some memory leaks

ZoltanFridrich:
If I am not mistaken, the functions `order_hostkeyalgs()` and `list_hostkey_types()` both return pointer to dynamically allocated memory. However, the pointers are forever lost as they are just passed into funtion `compat_pkalg_proposal()` where the pointer gets overwritten by newly allocated one.

Signed-off-by: Zoltan Fridrich <zfridric@redhat.com>

### DIFF
--- a/compat.c
+++ b/compat.c
@@ -156,49 +156,64 @@ compat_banner(struct ssh *ssh, const char *version)
 	debug_f("no match: %s", version);
 }
 
+/* Always returns pointer to allocated memory */
 char *
 compat_cipher_proposal(struct ssh *ssh, char *cipher_prop)
 {
+	char *p;
+
 	if (!(ssh->compat & SSH_BUG_BIGENDIANAES))
-		return cipher_prop;
+		return xstrdup(cipher_prop);
 	debug2_f("original cipher proposal: %s", cipher_prop);
-	if ((cipher_prop = match_filter_denylist(cipher_prop, "aes*")) == NULL)
+	if ((p = match_filter_denylist(cipher_prop, "aes*")) == NULL)
 		fatal("match_filter_denylist failed");
-	debug2_f("compat cipher proposal: %s", cipher_prop);
-	if (*cipher_prop == '\0')
+	debug2_f("compat cipher proposal: %s", p);
+	if (*p == '\0')
 		fatal("No supported ciphers found");
-	return cipher_prop;
+	return p;
 }
 
+/* Always returns pointer to allocated memory */
 char *
 compat_pkalg_proposal(struct ssh *ssh, char *pkalg_prop)
 {
+	char *p;
+
 	if (!(ssh->compat & SSH_BUG_RSASIGMD5))
-		return pkalg_prop;
+		return xstrdup(pkalg_prop);
 	debug2_f("original public key proposal: %s", pkalg_prop);
-	if ((pkalg_prop = match_filter_denylist(pkalg_prop, "ssh-rsa")) == NULL)
+	if ((p = match_filter_denylist(pkalg_prop, "ssh-rsa")) == NULL)
 		fatal("match_filter_denylist failed");
-	debug2_f("compat public key proposal: %s", pkalg_prop);
-	if (*pkalg_prop == '\0')
+	debug2_f("compat public key proposal: %s", p);
+	if (*p == '\0')
 		fatal("No supported PK algorithms found");
-	return pkalg_prop;
+	return p;
 }
 
+/* Always returns pointer to allocated memory */
 char *
-compat_kex_proposal(struct ssh *ssh, char *p)
+compat_kex_proposal(struct ssh *ssh, char *prop)
 {
+	char *tmp, *p;
+
+	p = xstrdup(prop);
 	if ((ssh->compat & (SSH_BUG_CURVE25519PAD|SSH_OLD_DHGEX)) == 0)
 		return p;
 	debug2_f("original KEX proposal: %s", p);
-	if ((ssh->compat & SSH_BUG_CURVE25519PAD) != 0)
-		if ((p = match_filter_denylist(p,
-		    "curve25519-sha256@libssh.org")) == NULL)
+	if ((ssh->compat & SSH_BUG_CURVE25519PAD) != 0) {
+		tmp = match_filter_denylist(p, "curve25519-sha256@libssh.org");
+		free(p);
+		if (tmp == NULL)
 			fatal("match_filter_denylist failed");
+		p = tmp;
+	}
 	if ((ssh->compat & SSH_OLD_DHGEX) != 0) {
-		if ((p = match_filter_denylist(p,
-		    "diffie-hellman-group-exchange-sha256,"
-		    "diffie-hellman-group-exchange-sha1")) == NULL)
+		tmp = match_filter_denylist(p, "diffie-hellman-group-exchange-sha256,"
+					       "diffie-hellman-group-exchange-sha1");
+		free(p);
+		if (tmp == NULL)
 			fatal("match_filter_denylist failed");
+		p = tmp;
 	}
 	debug2_f("compat KEX proposal: %s", p);
 	if (*p == '\0')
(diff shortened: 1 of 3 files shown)


ZoltanFridrich:
As it seems, the compat_*_proposal functions sometimes return a newly allocated pointer or they return the original pointer, which makes it difficult to manage the memory. I have changed those functions to always return a pointer to newly allocated memory.

compat.c
```diff
@@ -156,53 +156,80 @@ compat_banner(struct ssh *ssh, const char *version)
...
 
+/* Always returns pointer to allocated memory */
 char *
 compat_cipher_proposal(struct ssh *ssh, char *cipher_prop)
 {
-	if (!(ssh->compat & SSH_BUG_BIGENDIANAES))
-		return cipher_prop;
+	char *p;
+	if (!(ssh->compat & SSH_BUG_BIGENDIANAES)) {
+		p = strdup(cipher_prop);
```
daztucker, line 165: you can simplify this to just

return xstrdup(cipher_prop);

compat.c
```diff
@@ -156,53 +156,80 @@ compat_banner(struct ssh *ssh, const char *version)
...
 
+/* Always returns pointer to allocated memory */
 char *
 compat_pkalg_proposal(struct ssh *ssh, char *pkalg_prop)
 {
-	if (!(ssh->compat & SSH_BUG_RSASIGMD5))
-		return pkalg_prop;
+	char *p;
+	if (!(ssh->compat & SSH_BUG_RSASIGMD5)) {
+		p = strdup(pkalg_prop);
```
daztucker, line 186: return xstrup(pkalg_prop);

compat.c
```diff
@@ -156,53 +156,80 @@ compat_banner(struct ssh *ssh, const char *version)
...
+	return p;
 }
 
+/* Always returns pointer to allocated memory */
 char *
 compat_kex_proposal(struct ssh *ssh, char *p)
 {
+	char *tmp, *prop;
+
+	prop = strdup(p);
```
daztucker, line 207: xstrdup

compat.c
```diff
@@ -156,53 +156,80 @@ compat_banner(struct ssh *ssh, const char *version)
...
+	return p;
 }
 
+/* Always returns pointer to allocated memory */
 char *
 compat_pkalg_proposal(struct ssh *ssh, char *pkalg_prop)
 {
-	if (!(ssh->compat & SSH_BUG_RSASIGMD5))
-		return pkalg_prop;
+	char *p;
```
daztucker, line 184: empty line after variable declarations.

compat.c
```diff
@@ -156,53 +156,80 @@ compat_banner(struct ssh *ssh, const char *version)
...
+	debug2_f("compat public key proposal: %s", p);
+	if (*p == '\0')
 		fatal("No supported PK algorithms found");
-	return pkalg_prop;
+	return p;
 }
 
+/* Always returns pointer to allocated memory */
 char *
 compat_kex_proposal(struct ssh *ssh, char *p)
```
daztucker, line 203: if you change the argument from p to prop and use p as the newly allocated string below, fewer of the lines below will need to change and the diff will be easier to read (plus it'll be more consistent with the functions above).

sshconnect2.c
```diff
@@ -217,7 +217,7 @@ ssh_kex2(struct ssh *ssh, char *host, struct sockaddr *hostaddr, u_short port,
     const struct ssh_conn_info *cinfo)
 {
 	char *myproposal[PROPOSAL_MAX] = { KEX_CLIENT };
-	char *s, *all_key;
+	char *s, *all_key, *hostkeyalgs;
```
daztucker, line 220: plz initialize hostkeyalgs to NULL so that free'ing it is always safe (usage below looks fine, but this might prevent problems in the future).

sshd.c
```diff
@@ -2363,6 +2363,7 @@ static void
 do_ssh2_kex(struct ssh *ssh)
 {
 	char *myproposal[PROPOSAL_MAX] = { KEX_SERVER };
+	char *hostkey_types;
```
daztucker, line 2366: initialize to NULL

compat.c
```diff
@@ -156,53 +156,80 @@ compat_banner(struct ssh *ssh, const char *version)
 	debug_f("no match: %s", version);
 }
 
+/* Always returns pointer to allocated memory */
 char *
 compat_cipher_proposal(struct ssh *ssh, char *cipher_prop)
 {
-	if (!(ssh->compat & SSH_BUG_BIGENDIANAES))
-		return cipher_prop;
+	char *p;
```
daztucker, line 163: empty line after variable declarations.

ZoltanFridrich:
@daztucker thank you for the review, all threads are resolved now.

compat.c
```diff
@@ -156,49 +156,66 @@ compat_banner(struct ssh *ssh, const char *version)
...
 {
+	char *p;
+
 	if (!(ssh->compat & SSH_BUG_RSASIGMD5))
-		return pkalg_prop;
+		return xstrup(pkalg_prop);
 	debug2_f("original public key proposal: %s", pkalg_prop);
-	if ((pkalg_prop = match_filter_denylist(pkalg_prop, "ssh-rsa")) == NULL)
+	p = match_filter_denylist(pkalg_prop, "ssh-rsa");
+	if (p == NULL)
```
daztucker, line 187: I'd prefer to keep the original "if ((foo = bar()) != baz)" form

compat.c
```diff
@@ -156,49 +156,66 @@ compat_banner(struct ssh *ssh, const char *version)
...
+	char *tmp, *p;
+
+	p = xstrdup(p);
 	if ((ssh->compat & (SSH_BUG_CURVE25519PAD|SSH_OLD_DHGEX)) == 0)
 		return p;
 	debug2_f("original KEX proposal: %s", p);
-	if ((ssh->compat & SSH_BUG_CURVE25519PAD) != 0)
-		if ((p = match_filter_denylist(p,
-		    "curve25519-sha256@libssh.org")) == NULL)
+	if (ssh->compat & SSH_BUG_CURVE25519PAD) {
```
daztucker, line 205: please keep the !=0 test so this line remains unchanged.

compat.c
```diff
@@ -156,49 +156,66 @@ compat_banner(struct ssh *ssh, const char *version)
...
+		free(p);
+		if (tmp == NULL)
 			fatal("match_filter_denylist failed");
-	if ((ssh->compat & SSH_OLD_DHGEX) != 0) {
-		if ((p = match_filter_denylist(p,
-		    "diffie-hellman-group-exchange-sha256,"
-		    "diffie-hellman-group-exchange-sha1")) == NULL)
+		p = tmp;
+	}
+	if (ssh->compat & SSH_OLD_DHGEX) {
```
daztucker, line 212: ditto.

sshd.c
```diff
@@ -2384,8 +2385,12 @@ do_ssh2_kex(struct ssh *ssh)
 		ssh_packet_set_rekey_limits(ssh, options.rekey_limit,
 		    options.rekey_interval);
 
+	hostkey_types = list_hostkey_types();
+	if (hostkey_types == NULL)
```
daztucker, line 2389: if ((hostkey_types = list_hostkey_types()) == NULL)

sshconnect2.c
```diff
@@ -255,9 +255,12 @@ ssh_kex2(struct ssh *ssh, char *host, struct sockaddr *hostaddr, u_short port,
 	    myproposal[PROPOSAL_MAC_ALGS_STOC] = options.macs;
 	if (use_known_hosts_order) {
 		/* Query known_hosts and prefer algorithms that appear there */
+		hostkeyalgs = order_hostkeyalgs(host, hostaddr, port, cinfo);
```
daztucker, line 258: if ((hostkeyalgs = order_hostkeyalgs(host, hostaddr, port, cinfo)) == NULL)

compat.c
ZoltanFridrich: I had to add brackets, the line will not be unchanged anyway

ZoltanFridrich:
@daztucker threads resolved again.

ZoltanFridrich:
@daztucker can you please restart the tests? There was a type in strdup.

daztucker:
Done

daztucker (commented):
Looks pretty good.  A couple of other FYI notes (no need to respin the commit, it needs to go upstream to OpenBSD anyway so I'll tweak it on the way).

PR #325, opened 2022-06-10
Providing MinRSABits option to limit RSA key length

beldmit:
There is a need to increase RSA key requirements to make the installations more secure. Just updating the default compiled-in value isn't an option because it may significantly break legacy systems compatibility. This PR introduces a new configuration option to be managed for security's sake.

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -174,6 +174,7 @@ typedef enum {
 	oStreamLocalBindMask, oStreamLocalBindUnlink, oRevokedHostKeys,
 	oFingerprintHash, oUpdateHostkeys, oHostbasedAcceptedAlgorithms,
 	oPubkeyAcceptedAlgorithms, oCASignatureAlgorithms, oProxyJump,
+	oMinRSABits,
 	oSecurityKeyProvider, oKnownHostsCommand,
 	oIgnore, oIgnoredUnknownOption, oDeprecated, oUnsupported
 } OpCodes;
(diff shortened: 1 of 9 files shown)


beldmit:
Next CI fix provided, could you please reapprove the workflow?

beldmit:
Lack of libcbor/libfido2 seem unrelated.

djmdjm:
I don't like the approach of adding a module-global minimum size to ssh-rsa.c partly because we're trying to remove global state, but mostly because it causes enforcement of the size limit to be performed in a situation where there is little context as to the origin of the key and so little possibility of showing the user a useful error message.

https://github.com/djmdjm/openssh-wip/pull/13 has a different approach that explicitly checks the key size at the times they are loaded or received

beldmit:
Well, the low-level approach has the advantage that we don't miss any code path by accident. I agree that providing the context is worth efforts. I also wanted to make the patch as little invasive as possible.

Thank you, I will review your implementation

beldmit:
Our tests found that your patch is incomplete in case when a particular private key is passed via the command line.

I suggest the following fix for it:
```
@@ -1762,6 +1762,12 @@ load_identity_file(Identity *id)
                       private = NULL;
                       quit = 1;
               }
+              if (r = sshkey_check_rsa_length(private, options.rsa_min_size) != 0) {
+                      debug_fr(r, "Skipping key %s", id->filename);
+                      sshkey_free(private);
+                      private = NULL;
+                      quit = 1;
+              }
               if (!quit && private != NULL && id->agent_fd == -1 &&
                   !(id->key && id->isprivate))
                       maybe_add_key_to_agent(id->filename, private, comment,
```

djmdjm:
Just added RequiredRSASize in 1875042c52, 54b333d12e and 07d8771bac

PR #327, opened 2022-06-18
make hint more intuitive

ExposedCat:
Hint looks like you must write only the filename to save the key, in the directory specified in parentheses, which is wrong

### DIFF
--- a/ssh-keygen.c
+++ b/ssh-keygen.c
@@ -302,7 +302,7 @@ ask_filename(struct passwd *pw, const char *prompt)
 	}
 	snprintf(identity_file, sizeof(identity_file),
 	    "%s/%s", pw->pw_dir, name);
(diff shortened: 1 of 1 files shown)


kalvdans (approved):

lijas:
I ran in to this problem and agree that the hint should be better. I only entered the filename and the .ssh folder was not generated. 
It seems to be a rather common problem: https://stackoverflow.com/questions/3076043/creating-a-ssh-key-with-ssh-keygen-does-not-create-the-ssh-folder

ExposedCat:
Given that this was reviewed, can this be merged?

kalvdans:
I'm just a random guy without merge rights, hopefully my reviews helps the real maintainers a bit.

ExposedCat:
Maybe @djmdjm ?

PR #328, opened 2022-06-21
Make it a bit more explicit that RekeyLimit applies to the receiving direction as well.

adelton:


djmdjm:
Thanks - applied upstream and synced back as 455cee8d6c

PR #329, opened 2022-06-29
sk-usbhid: improved PIN error handling

LDVG:
**NOTE:** This PR is a continuation of #310; marked as draft until #310 is merged/closed.

For `verify-required` and `resident` options, `ssh-keygen` always prompts for a PIN. If the authenticator does not have a PIN set, the user is presented with an `"invalid format"` error message. 

Instead of preemptively asking for a PIN, determine whether it's required by examining the authenticator's return value. If a PIN is required but not set, optionally set a new PIN. Otherwise, print a more informative error message to the user.


LDVG:
Rebased since #310 and parts of #302 were merged. The latter dropped preemptive prompts for the authenticator PIN; this PR now primarily aims to help the user enroll a PIN when creating resident or verify-required credentials.

PR #330, opened 2022-07-06
Avoid assuming layout of fd_set

CendioOssman:
POSIX doesn't specify the internal layout of the `fd_set` object, so let's not assume it is just a bit mask. This increases compatibility with systems that have a different layout.

The assumption is also worthless as we already refuse to use file descriptors over `FD_SETSIZE` anyway. Meaning that the default size of `fd_set` is quite sufficient.

### DIFF
--- a/openbsd-compat/bsd-poll.c
+++ b/openbsd-compat/bsd-poll.c
@@ -47,9 +47,8 @@ ppoll(struct pollfd *fds, nfds_t nfds, const struct timespec *tmoutp,
     const sigset_t *sigmask)
 {
 	nfds_t i;
-	int saved_errno, ret, fd, maxfd = 0;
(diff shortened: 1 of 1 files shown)


daztucker:
Change looks reasonable.  On what platforms are fd_sets not bit vectors?

> The failing tests don't look like it comes from this change, so hopefully that's not a blocker?

They were indeed unrelated and is not a blocker. (They can get broken by several dependencies as well as us, and some of the tests are inherently racy.  After some recent infrastructure work they are all green and hopefully should be a bit less flaky.)

daztucker:
Applied, thanks.

CendioOssman:
We needed this for our port to Windows. So not one of the officially supported platforms. I haven't checked if any Unix dialect also needs this.

PR #331, opened 2022-07-11
Add flag to filter default algorithms

CendioOssman:
Allows you to more easily select a subset of algorithms you want to use, without also overriding the priority, or enabling something that is no longer deemed safe to use.

### DIFF
--- a/kex.c
+++ b/kex.c
@@ -264,6 +264,15 @@ kex_assemble_names(char **listp, const char *def, const char *all)
 		}
 		free(list);
 		list = tmp;
+	} else if (*list == '&') {
+		/* Filter default list */
+		if ((*listp = match_filter_allowlist(def, list + 1)) == NULL) {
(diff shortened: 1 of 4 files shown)


djmdjm:
What is the difference between this and the existing `-` flag? E.g. `KexAlgorithms=-diffie-hellman*`

CendioOssman:
If new algorithms are added, then an existing command line using `-` needs to be updated. This is stricter in that only the things mentioned on the command line can be used. So it's closer to the prefix-less behaviour. Only that it doesn't add or reorder.

The goal was to retain some of the trust in OpenSSH's defaults and avoid pushing that responsibility on the caller. In our case, it was because host keys storage is handled by the caller, so we needed a mechanism to inform ssh which ones are available. But we'd like to retain the priority that upstream has deemed best. And not accidentally enable something that is now considered too old.

It's meant to be an intersection operator, rather than a subtraction or addition.

PR #332, opened 2022-07-31
ssh-keysign: Fix free() calls in validate_request

v-rzh:
In `validate_request()`  `key` and `pkalg` are freed before the function returns regardless of whether the function fails. In case the function fails, `key` is freed twice. This patch moves the frees after use.

### DIFF
--- a/ssh-keysign.c
+++ b/ssh-keysign.c
@@ -155,10 +155,7 @@ valid_request(struct passwd *pw, char *host, struct sshkey **ret, char **pkalgp,
 
 	debug3_f("fail %d", fail);
 
-	if (fail) {
-		sshkey_free(key);
-		free(pkalg);
-	} else {
+	if (!fail) {
 		if (ret != NULL) {
 			*ret = key;
 			key = NULL;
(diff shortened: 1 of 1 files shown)


djmdjm:
I think it only needs this

```
-	if (fail)
-		sshkey_free(key);
-	else {
+	if (!fail) {
```

v-rzh:
This would solve the issue with double freeing `key`, however in the case where `validate_request()` is successful we take another reference to `pkalg` via `pkalgp` and it's later used in a call to `ssh_keysign()`. So I think `pkalg` should only be freed if `validate_request()` fails.

djmdjm:
sure, but that can never be double-freed:

```
		if (pkalgp != NULL) {
			*pkalgp = pkalg;
			pkalg = NULL;
		}
```

v-rzh:
Oh right! I'll amend the patch. The free() and sshkey_free() calls I added in `main()` should stay, right?

djmdjm:
I have committed the sshkey_free() fix upstream and synced it back to openssh-portable as 141535b904b - thanks. 

ATM we don't care so much about one-shot leaks in main(), as the process is destined to exit anyway.

v-rzh:
Thanks!

PR #333, opened 2022-08-02
Double free in hash_file function in case of read error

Eusgor:
If an error occurs while reading the file, it is possible to double free ctx


djmdjm:
Committed upstream and synced back as 5062ad48814b - thanks!

PR #336, opened 2022-08-17
Adding `const` to read-only variables

MaxKellermann:


MaxKellermann:
I ... have no idea what this CI failure means, but it doesn't look related to this PR.

daztucker:
> I ... have no idea what this CI failure means, but it doesn't look related to this PR.

Yeah, there's something wrong with sudo on the new ubuntu-22.04 images.  I filed https://github.com/actions/runner-images/issues/6106

daztucker:
5a5c580b48fc6006bdfa731fc2f6d4945c2c0e4e should work around the issue on ubuntu-22.04, so if you rebase it should be happier.

MaxKellermann:
Rebased, let's see.

MaxKellermann:
Is there anybody out there?

PR #338, opened 2022-08-19
sk-usbhid: fix key_lookup() on tokens with built-in UV

martelletto:
`key_lookup()` relied on `pin != NULL`. With ssh-keygen.c r1.458 (git 2a108c0e), that premise is violated. Given we enforce `FIDO_CRED_PROT_UV_{REQUIRED,OPTIONAL_WITH_ID}`, `key_lookup()` will never succeed in finding resident keys if `pin == NULL`.

On tokens without built-in UV, `sk_enroll()` fails, we retry with `pin != NULL`, and `key_lookup()` works. On tokens with built-in UV, `sk_enroll()` does not fail, and any existing resident key is silently overwritten. To fix this, we need to check if `pin == NULL` in `key_lookup()` and, if so, check whether the device supports built-in UV, in which case we enable UV in the assertion used by `key_lookup()`, allowing it to find existing resident keys. This is implemented in the present PR.

To reproduce using a Yubikey Bio:

```                                  
$ ssh-keygen -t ecdsa-sk -O resident -f /tmp/key                                
Generating public/private ecdsa-sk key pair.                                    
You may need to touch your authenticator to authorize key generation.           
Enter passphrase (empty for no passphrase):                                     
Enter same passphrase again:                                                    
Your identification has been saved in /tmp/key                                  
Your public key has been saved in /tmp/key.pub                                  
(...)           
$ ssh-keygen -t ecdsa-sk -O resident -f /tmp/key                                
Generating public/private ecdsa-sk key pair.                                    
You may need to touch your authenticator to authorize key generation.           
Enter passphrase (empty for no passphrase):                                     
Enter same passphrase again:                                                    
Your identification has been saved in /tmp/key                                  
Your public key has been saved in /tmp/key.pub                                  
(...)
```
                                                                        
Expected behaviour:

```                                                                               
$ ssh-keygen -t ecdsa-sk -O resident -f /tmp/key                                
Generating public/private ecdsa-sk key pair.                                    
You may need to touch your authenticator to authorize key generation.           
A resident key scoped to 'ssh:' with user id 'null' already exists.             
Overwrite key in token (y/n)?
```

### DIFF
--- a/sk-usbhid.c
+++ b/sk-usbhid.c
@@ -783,6 +783,7 @@ key_lookup(fido_dev_t *dev, const char *application, const uint8_t *user_id,
 	fido_assert_t *assert = NULL;
 	uint8_t message[32];
 	int r = FIDO_ERR_INTERNAL;
+	int sk_supports_uv, uv;
 	size_t i;
 
 	memset(message, '\0', sizeof(message));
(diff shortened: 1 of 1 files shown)


djmdjm:
Makes sense - thanks again. Merged as 5f45c2395c60865

PR #344, opened 2022-09-22
PROTOCOL: Fix typo

pablomh:
s/to returns/to return/


daztucker:
Applied upstream, thanks.

PR #346, opened 2022-10-04
Permit use of $XDG_RUNTIME_DIR for user files

jcpunk:
Some user files are basically worthless when the connection ends - such as the agent socket.  Other files, like forwarded kerberos tickets, may have value after the session terminates.

GSSAPICleanupCredentials is a good step, but for many systems /tmp is a persistent volume (/tmp).  Having active, unencrypted login authentication credentials on a remote persistent volume is not great.

https://bugzilla.mindrot.org/show_bug.cgi?id=2950
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables


session.c
```diff
@@ -166,6 +166,7 @@ static char *auth_info_file = NULL;
 /* Name and directory of socket for authentication agent forwarding. */
 static char *auth_sock_name = NULL;
 static char *auth_sock_dir = NULL;
+int remove_auth_sock_dir = 1;
```
kalvdans, line 169:
```suggestion
static int remove_auth_sock_dir = 1;
```

ssh-agent.c
```diff
@@ -167,6 +167,7 @@ pid_t cleanup_pid = 0;
 /* pathname and directory for AUTH_SOCKET */
 char socket_name[PATH_MAX];
 char socket_dir[PATH_MAX];
+int remove_socket_dir = 1;
```
kalvdans, line 170:
```suggestion
static int remove_socket_dir = 1;
```

kalvdans (changes requested):
The variables are only used inside one compilation unit.

Diti:
Probably won’t ever be fixed. The OpenSSH project is hardcore close-minded about the issue. See the [archive.org link of the bug cited above](https://web.archive.org/web/20190925004614/https://bugzilla.mindrot.org/show_bug.cgi?id=2050).

auth-krb5.c
```diff
@@ -242,11 +244,14 @@ krb5_cleanup_proc(Authctxt *authctxt)
 krb5_error_code
 ssh_krb5_cc_gen(krb5_context ctx, krb5_ccache *ccache) {
 	int tmpfd, ret, oerrno;
-	char ccname[40];
+	char ccname[PATH_MAX] = {0};
+	char *path;
 	mode_t old_umask;
 
+	path = session_get_runtime_directory();
```
djmdjm: I don't really want the auth code depending on the session code because it interferes with the goal of splitting auth-related things from the sshd process that handles user sessions. Can you adapt this to use the version you have in misc.c? Alternately an auth_get_runtime_directory() might be acceptable.

PR #347, opened 2022-10-04
Give info about ssh-keygen -R to remove offending key

mh4ckt3mh4ckt1c4s:
Giving a `sed` command in the error message about offending server (host) key allowing to just copy and paste the command in order to remove the offending key.

### DIFF
--- a/sshconnect.c
+++ b/sshconnect.c
@@ -1,4 +1,4 @@
-/* $OpenBSD: sshconnect.c,v 1.360 2022/11/03 21:59:20 djm Exp $ */
+/* $OpenBSD: sshconnect.c,v 1.358 2022/08/26 08:16:27 djm Exp $ */
 /*
  * Author: Tatu Ylonen <ylo@cs.hut.fi>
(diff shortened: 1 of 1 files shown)


daztucker:
Firstly, sed is the wrong hammer for this particular nail; ssh-keygen has -R which does this.  Secondly, we have been historically reluctant to provide a way for the user to blindly do this since it enables MITMing of subsequent connections.

There's a bug requesting this: https://bugzilla.mindrot.org/show_bug.cgi?id=2045 however due to the aforementioned reasons it's not seen much love.

mh4ckt3mh4ckt1c4s:
I changed the info to display the use of `ssh-keygen -R` instead of my previous `sed` proposition. I also replaced the display of the host so that the command cannot be completely copied and pasted anymore.

 From my experience not a lot of people know about the `ssh-keygen -R` command, and are using `sed`, or worse, are deleting completely the `known_hosts` file (as the complete path is given in the error message and can be easily copied and paste). Even if we're not giving a complete command to copy and paste blindly, adding the info about `ssh-keygen` seems appropriate to me.

PR #349, opened 2022-10-13
fix configure for clang 15

haampie:
See https://bugzilla.mindrot.org/show_bug.cgi?id=3482

And https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics:

> The -Wimplicit-int warning diagnostic now defaults to an error in C99 and
> later. Prior to C2x, it may be downgraded to a warning with
> -Wno-error=implicit-int, or disabled entirely with -Wno-implicit-int. As of
> C2x, support for implicit int has been removed, and the warning options will
> have no effect. Specifying -Wimplicit-int in C89 mode will now issue warnings
> instead of being a noop.

The configure scripts shouldn't be using `main() {...}` but `int main() {...}`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -713,7 +713,7 @@ case "$host" in
 	AC_RUN_IFELSE([AC_LANG_SOURCE([[
 #include <mach-o/dyld.h>
 #include <stdlib.h>
-main() { if (NSVersionOfRunTimeLibrary("System") >= (60 << 16))
+int main() { if (NSVersionOfRunTimeLibrary("System") >= (60 << 16))
 		exit(0);
 	else
 		exit(1);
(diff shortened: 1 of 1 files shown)


haampie:
@djmdjm the current ci failure seems unrelated, it's on the master branch too?

Would it make sense to add CI for clang 15 to catch other errors?

daztucker:
> @djmdjm the current ci failure seems unrelated, it's on the master branch too?

Yes, the failure on Mac OS X is unrelated (we know what broke, we're just waiting for some more detail on the thing the last commit was intended to fix so we can fix it while also unbreaking OS X.

> Would it make sense to add CI for clang 15 to catch other errors?

Yes, but we are constrained to the versions available on the runners supplied by github (short of building clang as part of each test, but that's not something I would want to do).

daztucker:
Applied, thanks.

PR #352, opened 2022-10-24
doc: Improve IdentityFile documentation

laalsaas:
The IdentityFile option in ssh_config can also be used to specify a public key file, as documented in ssh.1 for the -i option. Document this also for IdentityFile in ssh_config.5, for documentation completeness.


daztucker:
applied, thanks.

PR #354, opened 2022-10-28
Improve openssh to check is getentropy() supported and use compat way…

simsergey:
The version 9.1_p1 is broken for very old kernels. Call getentropy() and goto fatal when error occurred, with no check error for "function is not implemented". The patch make improve to use old compatible method for getting entropy. No any side effects for devices with a fresh kernel.

### DIFF
--- a/openbsd-compat/arc4random.c
+++ b/openbsd-compat/arc4random.c
@@ -49,8 +49,11 @@
  * a system that has a native getentropy OpenSSL cannot call the wrong one.
  */
 #ifndef HAVE_GETENTROPY
-# define getentropy(x, y) (_ssh_compat_getentropy((x), (y)))
+#define _getentropy(x, y) (_ssh_compat_getentropy((x), (y)))
+#else
+#define _getentropy getentropy
 #endif
+
 
 #include "log.h"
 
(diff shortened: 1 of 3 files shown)


z3ntu:
@daztucker Please review.

daztucker:
Can  you give me an example of an affected platform?  (preferably x86 based so I can add it t our CI VM zoo)

This code was lifted from LibreSSL so we should upstream a similar change.

daztucker:
I think I'd rather the fallback be in bsd-getrandom.c, like so: c339e6c4924f70f533c1f1d294a8419bef4661ba
That's where the platform specific code is, and putting it in arc4random will increase the work required to keep it in sync in the future.

Originally _ssh_compat_getentropy was just called getentropy so adding a fallback to the real getentropy would have been confusing at the least.

Does this also solve your problem?

simsergey:
> Can you give me an example of an affected platform? (preferably x86 based so I can add it to our CI VM zoo)
I tested it on Asus Zenfone 2 (ZE551ML) based on Intel Atom Z3590 with kernel 3.10.72 (downstream) and PostmarketOS (based on Alpine).

daztucker:
I think I see what I did wrong.  Could you please try 800d77585a714f93b4532281077792138ef728d1 instead?

simsergey:
> I think I see what I did wrong. Could you please try [800d775](https://github.com/openssh/openssh-portable/commit/800d77585a714f93b4532281077792138ef728d1) instead?

Yes, it is work. Thank you.

daztucker:
Thanks for confirming.  @djmdjm ok for 800d775 ?  I'd probably add an explicit "return 0" inside each ifdef block and a "return -1" at the end (right now it'll fatal() out before it gets to the end, but should that ever change in the future it should return failure).

djmdjm:
ok djm@

daztucker:
Applied 800d77585a714f93b4532281077792138ef728d1, thanks.

PR #355, opened 2022-11-06
configure.ac: Add <pty.h> include for openpty

thesamesam:
Another Clang 16ish fix (which makes -Wimplicit-function-declaration an error by default).

See: 2efd71da49b9cfeab7987058cf5919e473ff466b
See: be197635329feb839865fdc738e34e24afd1fca8


daztucker:
Applied, thanks.

PR #356, opened 2022-11-21
Fixed compatibility issue with getpgrp()

andrewadiletta:
Fixed compatibility issue with getpgrp() to conform to modern linux standards without any formal arguments

### DIFF
--- a/defines.h
+++ b/defines.h
@@ -702,7 +702,7 @@ struct winsize {
 
 #ifndef GETPGRP_VOID
 # include <unistd.h>
-# define getpgrp() getpgrp(0)
+# define getpgrp() getpgrp()
 #endif
 
 #ifdef USE_BSM_AUDIT


daztucker:
The patch is not correct since it changes the define to be a no-op.  If that's really what's needed it should be deleted instead.

That said, there are some platforms that might still need this, and it should be detected correctly by configure.  Under what circumstances is this causing problems?

This was added in 606f880e0fd1a31f5beb3b37ece0c12317a9ed61, and the relevant part of which seems to be "- (djm) [Add Charles Levert <charles@comm.polymtl.ca> getpgrp patch](https://marc.info/?l=openssh-unix-dev&m=96817268515093&w=2)".

I fiddled the configure test in c7b5a47e3b9db9a0f0198f9c90c705f6307afc2b and it's possible I did a worse job than AC_FUNC_GETPGRP does.

thesamesam:
@andrewadiletta What triggered you to report this? It's okay if this test warngs/errors out in config.log fwiw, as that's just a side-effect of the test.

andrewadiletta:
While this error occurred on every machine (hardware and virtual) that we tried, here are some easy steps to reproduce the issue...

Download virtualbox and create a new Ubuntu 64 bit instance (params like memory, storage, etc... don't effect the reproducability of the error)

Download the latest version of Ubuntu (22.04.1) from the official ubuntu download page
https://ubuntu.com/download/desktop


Install Ubuntu in the virtual machine, and with a fresh install of Ubuntu up and running, execute the following:

sudo apt update
sudo apt install autoconf
sudo apt install git
sudo apt install gcc
sudo apt install make
sudo apt install zlib1g-dev
sudo apt install libssl-dev
git clone https://github.com/openssh/openssh-portable.git
cd openssh-portable
autoconf
./Configure

This command will actually fail with the following output:

```
checking if cc supports -Werror... yes
./configure: line 6066: syntax error near unexpected token `-pipe'
./configure: line 6066: `	OSSH_CHECK_CFLAG_COMPILE(-pipe)'
```

The OSSH_CHECK appears to be broken in this configure file. Our team had an old version openssh-portable (which had a configure file that did not contain any OSSH_CHECK lines) which we copied to the machine, then re-executed the ./configure command which completed successfully. However, during the make command we recieved the following error.

```
defines.h:705:20: error: too many arguments to function ‘getpgrp’
  705 | # define getpgrp() getpgrp(0)
```

This error was resolved by my commit. It would actually be preferable to solve the first problem with OSSH_CHECK than the getpgrp issue, as my commit was more of a workaround than a fix. I'm also not sure what all the differences between our version of openssh-portable is, and the most recent version.

andrewadiletta:
I believe our version that works with the getpgrp workaround is commit 8926956f22639132a9f2433fcd25224e01b900f5 - but I'm not totally sure. Let me know if you need any more information

daztucker:
You caused this issue by replacing configure with one prior to c7b5a47e3b9db9a0f0198f9c90c705f6307afc2b which does not have the required tests.  You may have also broken other things in more subtle ways.  You can grab the correct configure from a current snapshot:
https://www.mindrot.org/openssh_snap/

If you want help debugging your configure problem, please email the openssh-unix-devel list or file a bug as per https://www.openssh.com/report.html.

PR #357, opened 2022-11-24
ssh-keygen: New option -G for removing a key passphrase

kixorz:
Hello,
I ran into a problem with `ssh-keygen` while using it from PowerShell. I needed to remove a passphrase from my private key and I determined there's no clean way to do so with the current `ssh-keygen` options and interface.

I fixed the problem by introducing a new `-G` option/switch that simply uses an empty passphrase from internal empty string.

Consider the following example:

```powershell
Start-Process -FilePath "C:\Program Files\OpenSSH\ssh-keygen.exe" -ArgumentList ("-p", "-f", "C:\Users\Administrator\.ssh\id_rsa", "-P", "<current passphrase>", "-N", "") 
```

The last argument, which is the new passphrase is empty.

```
Start-Process : Cannot validate argument on parameter 'ArgumentList'. The argument is null or empty. Provide an argument that is not null or empty, and then try the command
again.
```

With this change applied the code becomes:

```powershell
Start-Process -FilePath "C:\Program Files\OpenSSH\ssh-keygen.exe" -ArgumentList ("-p", "-f", "C:\Users\Administrator\.ssh\id_rsa", "-P", "<current passphrase>", "-G") 
```
and the process to remove a passphrase becomes straightforward even from Powershell.

Please consider suggesting changes or merging this change.

### DIFF
--- a/ssh-keygen.1
+++ b/ssh-keygen.1
@@ -61,6 +61,7 @@
 .Op Fl m Ar format
 .Op Fl N Ar new_passphrase
 .Op Fl P Ar old_passphrase
+.Op Fl G
 .Op Fl Z Ar cipher
 .Nm ssh-keygen
 .Fl i
@@ -468,6 +469,8 @@ when generating or updating a supported private key type will cause the
 key to be stored in the legacy PEM private key format.
 .It Fl N Ar new_passphrase
 Provides the new passphrase.
+.It Fl G
+Use a new empty passphrase.
 .It Fl n Ar principals
 Specify one or more principals (user or host names) to be included in
 a certificate when signing a key.
--- a/ssh-keygen.c
+++ b/ssh-keygen.c
@@ -107,6 +107,9 @@ static char *identity_passphrase = NULL;
 /* This is set to the new passphrase if given on the command line. */
 static char *identity_new_passphrase = NULL;
 
+/* This is the empty passphrase. */
+static char *identity_empty_passphrase = "";
+
 /* Key type when certifying */
 static u_int cert_key_type = SSH2_CERT_TYPE_USER;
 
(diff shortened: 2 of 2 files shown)


daztucker:
>  introducing a new -G option/switch

ssh-keygen option letters are a scarce resource.

> Start-Process : Cannot validate argument on parameter 'ArgumentList'. The argument is null or empty. Provide an argument that is not null or empty, and then try the command
> again.
> ```

That very much sounds like a bug in Powershell.  Borne type shells have been able to do this for roughly half a century.

kixorz:
Thanks for your comment. I agree on the first point. On the second point I think it's some Powershell design decision.

I have considered a change that wouldn't require a new option, but it would change the behavior of the `-N` option and I think that would have been unpopular.

daztucker:
> Thanks for your comment. I agree on the first point. On the second point I think it's some Powershell design decision.

Reading some of the powershell issues eg https://github.com/MicrosoftDocs/PowerShell-Docs/issues/7701 it seems like it might be the API powershell uses (CommandLineToArgvW).  If I'm reading that right, I think that adding a second set of quotes (ie -N "''") will get them passed to 
 CommandLineToArgvW which should do what you want.

> I have considered a change that wouldn't require a new option, but it would change the behavior of the `-N` option and I think that would have been unpopular.

Yeah that's a non-starter.

kixorz:
Thanks for the suggestion and confirmation of my thoughts around changing `-N`. I already went through testing various workarounds, including the one you're suggesting:
```
Start-Process -FilePath "C:\Program Files\OpenSSH\ssh-keygen.exe" -ArgumentList ("-p", "-f", "C:\users\Administrator\.ssh\id_rsa", "-P", "asdf", "-N", "''")
```
It unfortunately just sets the new passphrase to '' (two single quotes).

PR #358, opened 2022-11-24
Fix missing object in LIBOPENSSH_OBJS.

blunderer:
${LIBOPENSSH_OBJS} contains sshkey.o that depends on the `sshsk_sign` symbol exported by ssh-sk-client.o, but doesn't include the object file.

The causes all binaries linking against libssh.a to also have to manually depend on ${SKOBJS}.


blunderer:
I think I see why this object file was not included in the first place: there are 2 implementations, and depending on the binary, on or the other will be used.

PR #360, opened 2022-11-29
Give every printf like function __restrict for the format value

AtariDreams:
The format value has to be a string literal, every time. Otherwise, you are not using these functions correctly. To reinforce this fact, I put __restrict over every non-contrib example of this I could find.


AtariDreams:
Ok, so the CI seemed to get stuck, so I fixed the commit just in case.

kalvdans:
`__restrict` only have effect when two pointer arguments are of the same type, and says that they never can overlap. Your are looking for -Werror=format-nonliteral .

djmdjm:
> The format value has to be a string literal, every time. Otherwise, you are not using these functions correctly

I don't think this is true. The logging code itself uses snprint() with a non-literal format string

PR #361, opened 2022-11-29
Fix some compiler warnings

AtariDreams:
Warnings about using the wrong type, wrong types being passed to printf-like functions, etc.

### DIFF
--- a/auth-krb5.c
+++ b/auth-krb5.c
@@ -77,7 +77,7 @@ auth_krb5_password(Authctxt *authctxt, const char *password)
 #endif
 	krb5_error_code problem;
 	krb5_ccache ccache = NULL;
-	int len;
+	size_t len;
 	char *client, *platform_client;
 	const char *errmsg;
 
--- a/auth-rhosts.c
+++ b/auth-rhosts.c
@@ -154,7 +154,7 @@ check_rhosts_file(const char *filename, const char *hostname,
 			if (!innetgr(host + 1, hostname, NULL, NULL) &&
 			    !innetgr(host + 1, ipaddr, NULL, NULL))
 				continue;
-		} else if (strcasecmp(host, hostname) &&
+		} else if (strcasecmp(host, hostname) != 0 &&
 		    strcmp(host, ipaddr) != 0)
 			continue;	/* Different hostname. */
 
--- a/auth2-hostbased.c
+++ b/auth2-hostbased.c
@@ -171,7 +171,7 @@ hostbased_key_allowed(struct ssh *ssh, struct passwd *pw,
 {
 	const char *resolvedname, *ipaddr, *lookup, *reason;
 	HostStatus host_status;
-	int len;
+	size_t len;
 	char *fp;
 
 	if (auth_key_is_revoked(key))
--- a/auth2-pubkey.c
+++ b/auth2-pubkey.c
@@ -748,7 +748,8 @@ int
 user_key_allowed(struct ssh *ssh, struct passwd *pw, struct sshkey *key,
     int auth_attempt, struct sshauthopt **authoptsp)
 {
-	u_int success = 0, i;
+	int success = 0;
+	u_int i;
 	char *file;
 	struct sshauthopt *opts = NULL;
 	const char *remote_ip = ssh_remote_ipaddr(ssh);
--- a/auth2-pubkeyfile.c
+++ b/auth2-pubkeyfile.c
@@ -225,7 +225,7 @@ auth_process_principals(FILE *f, const char *file,
 	char loc[256], *line = NULL, *cp, *ep;
 	size_t linesize = 0;
 	u_long linenum = 0, nonblank = 0;
-	u_int found_principal = 0;
+	int found_principal = 0;
 
 	if (authoptsp != NULL)
 		*authoptsp = NULL;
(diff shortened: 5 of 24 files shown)


authfile.c
```diff
@@ -442,9 +442,7 @@ sshkey_in_file(struct sshkey *key, const char *filename, int strict_type,
 int
 sshkey_check_revoked(struct sshkey *key, const char *revoked_keys_file)
 {
-	int r;
-
-	r = ssh_krl_file_contains_key(revoked_keys_file, key);
+	int r = ssh_krl_file_contains_key(revoked_keys_file, key);
```
daztucker, line 445: style guide (https://man.openbsd.org/style.9) says not to use function calls as variable initializers.

authfile.c
```diff
@@ -474,10 +472,10 @@ sshkey_check_revoked(struct sshkey *key, const char *revoked_keys_file)
 int
 sshkey_advance_past_options(char **cpp)
 {
-	char *cp = *cpp;
+	char *cp;
 	int quoted = 0;
 
-	for (; *cp && (quoted || (*cp != ' ' && *cp != '\t')); cp++) {
+	for (cp = *cpp; *cp && (quoted || (*cp != ' ' && *cp != '\t')); cp++) {
```
daztucker, line 478: what warning does this fix?

openbsd-compat/base64.c
```diff
@@ -144,8 +144,8 @@ b64_ntop(u_char const *src, size_t srclength, char *target, size_t targsize)
 		srclength -= 3;
 
 		output[0] = input[0] >> 2;
-		output[1] = ((input[0] & 0x03) << 4) + (input[1] >> 4);
```
daztucker, line 147: what's warning is this supposed to fix?

daztucker (commented):
There are unrelated changes in the PR, eg changes to the autoconf files.

AtariDreams:
@daztucker Addressed!

AtariDreams:
> There are unrelated changes in the PR, eg changes to the autoconf files.

The restrict addition is so that __restrict does nothing on compilers that do not support it, while having the restrict keyword on platforms that do.

AtariDreams:
> There are unrelated changes in the PR, eg changes to the autoconf files.

I removed the bzero check since I replaced it with memset and then reran autoreconf

openbsd-compat/base64.c
daztucker: You didn't answer my question.

openbsd-compat/bsd-misc.c
```diff
@@ -387,14 +387,6 @@ llabs(long long j)
...
 #endif
 
-#ifndef HAVE_BZERO
-void
-bzero(void *b, size_t n)
-{
-	(void)memset(b, 0, n);
-}
-#endif
-
```
daztucker, line 397: bzero is still used in some places (eg the compat library, netcat in the tests) so this will break on platforms that don't have bzero.  It's also unrelated to the purported purpose of this commit so doesn't belong in this commit (in addition to being pointless churn).

regress/netcat.c
```diff
@@ -495,7 +495,7 @@ main(int argc, char *argv[])
 					sv = NULL;
 				else {
 					sv = getservbyport(
-					    ntohs(atoi(portlist[i])),
+					    ntohs((int)strtol(portlist[i], NULL, 10)),
```
daztucker, line 498: what warning is this fixing?  atoi already returns an int which is what's needed.

regress/unittests/test_helper/test_helper.c
```diff
@@ -392,7 +392,7 @@ tohex(const void *_s, size_t l)
 
 	assert(r != NULL);
 	for (i = j = 0; i < l; i++) {
-		r[j++] = hex[(s[i] >> 4) & 0xf];
+		r[j++] = hex[s[i] >> 4];
```
daztucker, line 395: what warning is this fixing?

it's also probably wrong on platforms that don't have an integral 8-bit type (Cray is the only example I know of and those are pretty rare these days, but it's allowed by the standards).

sc25519.c
```diff
@@ -166,7 +166,7 @@ int sc25519_lt_vartime(const sc25519 *x, const sc25519 *y)
 
 void sc25519_add(sc25519 *r, const sc25519 *x, const sc25519 *y)
 {
-  int i, carry;
+  crypto_uint32 i, carry;
```
daztucker, line 169: What warning is this fixing?

This code is taken entirely from its upstream (in supercop) and is only ever used for small integer loops so we'd need a really good reason to want to change it.

daztucker (changes requested):
You've still got unrelated changes in this PR (all of the changes to config.guess, config.sub, install-sh).  You're also mixing different classes of changes in a single commit that are  (eg replacing bzero, which is completely unjustified by the stated goal of this PR).

openbsd-compat/base64.c
AtariDreams: > Will roll back. Meant for another PR

AtariDreams:
Done @daztucker

AtariDreams:
Ok, it is all ready now!

AtariDreams:
@daztucker Thoughts on this?

PR #362, opened 2022-11-29
Update autotools

AtariDreams:
Regenerate config files using latest autotools


daztucker:
Applied, thanks (with the exception of the whitespace changes in configure.ac).

PR #363, opened 2022-11-30
Fix leak in sftp-server.c process_read()

jamesr:
process_read() has a static local u_char* buf that can grow using realloc. When growing, the function was allocating a new buffer with realloc(NULL, ...) aka malloc(...) and assigning the result into buf. This leaks if the buffer is allocated more that once. This changes the growth path to call realloc(buf, ...) to reuse or free the old buffer as needed.


djmdjm:
Thanks, this has been committed as 3f9231c2e1f3

PR #369, opened 2023-01-06
Update percent.sh

rmsh1216:
skip tests for '%C' since no openssl in local path


rmsh1216:
tests for '%C' failed since no openssl in local path
image

daztucker:
The percent.sh change should be fixed in d77fc611a62f2dfee0b654c31a50a814b13310dd.  The yes-head.sh one is in 923c3f437f439cfca238fba37e97a7041782f615.

Thanks.

PR #370, opened 2023-01-09
File exists control added before opening remote connection (SCP)

macorapci:
I have such a long password for the remote PC and getting `No such file or directory` after writing the password is so annoying. This is why I add a check before creating a connection. I didn't remove the existing control after entering the password because at that time folder can be removed.

````console
$ ./scp nonExistFolder  ubuntu@192.168.1.11:/home/ubuntu/folder
./scp: stat local "nonExistFolder": No such file or directory
````

Tests are successful. 

````console
test_hostkeys: .................. 18 tests ok
test_match: ...... 6 tests ok
test_misc: ........................................... 43 tests ok
make[1]: Leaving directory '/home/macorapci/Desktop/openssh-portable/regress'
unit tests passed
echo all tests passed
all tests passed
````

https://github.com/macorapci/openssh-portable/blob/ca9bb068015321c2045d02c5ba7fa4616abba4d8/scp.c#L682

I wanna add the condition for `tolocal` but the tests fail so I add it with the comment line. I can delete the line if it breaks something.

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -421,6 +421,7 @@ void run_err(const char *,...)
 int note_err(const char *,...)
     __attribute__((__format__ (printf, 1, 2)));
 void verifydir(char *);
+void verify_file_exists(char *);
 
 struct passwd *pwd;
 uid_t userid;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scp.c [hunk 1]
```diff
@@ -678,6 +679,7 @@ main(int argc, char **argv)
 	else {
 		if (targetshouldbedirectory)
 			verifydir(argv[argc - 1]);
+		// verify_file_exists(argv[argc - 1]);
```
macorapci, line 682: I wanna add the condition for tolocal but the tests fail so I add it with the comment line. I can delete the line if it breaks something.

djmdjm:
merged upstream as 4f0a67648670

PR #319, opened 2022-05-11
Expand environment variables in `Include` directives

cole-h:
This is a follow-up to https://github.com/openssh/openssh-portable/pull/308#issuecomment-1119261184.

In this patch series, I attempt (to the best of my ability) to address points 1. and 2. by changing all `fatal_f` calls in `vdollar_percent_expand` to `error_f`, and then by utilizing the `dollar_expand` function (a light wrapper around `vdollar_percent_expand`) to actually do the expansion for `Include` directives. I didn't seriously attempt point 3. because I don't know exactly what things should be expanded, how to get access to that inside `process_config_line_depth`, and if it is desirable to prevent `percent{,_dollar}_expand` from fatal'ing.

I also moved the "env var has no value" and "invalid environment expansion" messages to the debug channel to tone down the noise.

I may have been a bit overzealous in my `fatal_f` -> `error_f` migration, so please point out any instances which I should revert. It's also possible that some parts of the codebase expect `vdollar_percent_expand` to fatal -- if you know this to be the case, please point me in their directions and I'll attempt to sort it out.

If this is deemed an acceptable approach, I'll write tests.

---

I tested this with the following command line and config files:

```
[nix-shell:~/workspace/vcs/openssh-portable]$ ./ssh root@me -F sshconf -v
OpenSSH_9.0p1, OpenSSL 1.1.1n  15 Mar 2022
debug1: Reading configuration data sshconf
debug1: Reading configuration data /home/vin/workspace/vcs/openssh-portable/inc
debug1: /home/vin/workspace/vcs/openssh-portable/inc line 1: Applying options for me
debug1: vdollar_percent_expand: env var ${DNE} has no value
debug1: sshconf line 2: invalid environment expansion ${DNE}/lol.
debug1: Authenticator provider $SSH_SK_PROVIDER did not resolve; disabling
debug1: Connecting to 127.0.0.1 [127.0.0.1] port 22.
debug1: Connection established.
[...]
The authenticity of host '127.0.0.1 (127.0.0.1)' can't be established.
[...fingerprint...]
This host key is known by the following other names/addresses:
    ~/.ssh/known_hosts:3: localhost
    ~/.ssh/known_hosts:124: scadrial
Are you sure you want to continue connecting (yes/no/[fingerprint])? ^C
```

```
# sshconf
Include "${PWD}/inc"
Include "${DNE}/lol"
```

```
# ${PWD}/inc
Host me
   HostName 127.0.0.1
```

PR #371, opened 2023-01-11
Allow ssh_config location to be set in environment

trice-ardc:
Hi,

I heavily use environment modules (eg lmod) to switch between different sets of environment variables. When in a testing environment, it would be useful to override the default ssh config. I usually do this with the `-F` flag, and I understand it can be put in an alias for the session, but this has a few issues:

* This alias is not inherited by programs that use OpenSSH under the hood, such as Git and Rsync.
* Telling Git and Rsync to add extra arguments to the ssh client is doable but clumsy.
* It's not consistent with the rest of my workflow or with other common tools like Git, which eg offers a `GIT_CONFIG_GLOBAL` environment variable similar to the `SSH_CONFIG` variable proposed in this pull request.

The commit in this pull request passes all tests on my local system, including `make tests` as well as manual tests of the new functionality. It doesn't add any new dependencies because it only uses getenv from stdlib, which is already included.

When the variable is not set, `getenv("SSH_CONFIG")` simply returns NULL, so nothing changes for people not using this variable. If the user also adds a `-F` argument, it overrides the environment variable.

### DIFF
--- a/ssh.c
+++ b/ssh.c
@@ -706,6 +706,8 @@ main(int ac, char **av)
 	logfile = NULL;
 	argv0 = av[0];
 
+    config = getenv("SSH_CONFIG");
+
  again:
 	while ((opt = getopt(ac, av, "1246ab:c:e:fgi:kl:m:no:p:qstvx"
 	    "AB:CD:E:F:GI:J:KL:MNO:PQ:R:S:TVw:W:XYy")) != -1) { /* HUZdhjruz */
--- a/ssh_config.5
+++ b/ssh_config.5
@@ -2148,6 +2148,11 @@ and
 .Cm ProxyJump
 accept the tokens %%, %h, %n, %p, and %r.
 .Sh ENVIRONMENT VARIABLES
+The location of the default ssh configuration file can be overridden by
+setting the
+.Cm SSH_CONFIG
+environment variable.
+.Pp
 Arguments to some keywords can be expanded at runtime from environment
 variables on the client by enclosing them in
 .Ic ${} ,


trice-ardc:
I suspect you'll want some tests added under `regress/`. Would it just be a matter of repeating most of the tests in `sshcfgparse.sh` but adjust them as follows?

```
f=`SSH_CONFIG=$OBJ/ssh_config ${SSH} -G host | awk '/exitonforwardfailure/{print $2}'`
```

trice-ardc:
Most or all of the ci failures seem to be due to a problem with `azure.archive.ubuntu.com`. I assume I can ignore those?

daztucker:
Yeah it looks like github is having connectivity issues at the moment.  I told it to rerun the failed tests, but if they fail again then just ignore it for now and we'll rerun them later.

djmdjm:
I'm not a fan of basic behaviour being dependent on environment variables. Would allowing a variant of the existing ssh_config `Include` directive that allows expansion of environment in paths suit your needs? E.g. something like `Include ${SSH_CONFIG_PATH}`

(We'd still need to solve how to handle an empty environment variable in these situations)

djmdjm:
Something like what I described is implemented in #319

trice-ardc:
Hi @djmdjm ,
Thanks for following up. Yeah, I think I could make that work. I appreciate your explanation in #308 which I now realize is pretty much identical to the current pull request. (Sorry for the noise!) I'll close this off and will look forward to the change from #319 coming down the pipe :)
~ Tim

PR #373, opened 2023-01-23
Fix for Time based Rekeying not getting trigered when no package is t…

sanjeet-karan-singh:
Rekeying was not working until a package is received or sent.
The check to ensure packets are more than 0, before rekey timeout check was preventing rekey to occur, even though it reached timeout.


sanjeet-karan-singh:
Not sure why CI is failing. Can it be something environmental (something with CI)

daztucker:
There were two sets of failures both with the connection-timeout test:
1) the test uses ControlMaster muxing which uses descriptor passing over Unix domain sockets.  That's not supported on (at least) Cygwin.  We now skip that test if configure decides FD passing isn't supported.
2) Valgrind slows things down enough that it messed up the expected timing on the connect-timeout test.  This one is probably fixable, but my quick attempt at it failed, so for now we skip that test when testing with Valgrind.

Both of these things have been committed, so if you rebase you should see those tests pass.

sanjeet-karan-singh:
Thanks @daztucker, just rebased. Can you please review changes.
Sorry, I'm not sure what's procedure for requesting merge of PR.

PR #376, opened 2023-02-03
sftp: Extension request copy-file@openssh.com

mickae1:
add the possibility to copy a file. It is a function that I need. And It can be very usefull for other people.


petrovr:
mickae1 wrote:
> add the possibility to copy a file. ....

copy-file was proposed as standard extension , see draft

Roumen

PR #378, opened 2023-02-24
Actually document the agent channel

AstraLuma:
I can't actually find anyplace anywhere where the agent channel handshake is actually documented. So I'm adding it here.

(Because I had to go back to the RFCs, splunk two code bases, and guess in order to figure it out for https://github.com/paramiko/paramiko/issues/2182)

What's not really documented yet in this PR is that the client sends a `auth-agent-req@openssh.com` basically advertising that an agent is available, and the server opens an `auth-agent@openssh.com` channel to talk to it. But this isn't documented in the RFC about X11 forwarding either.

I also didn't write down any requirements, things like:

> Implementations MUST reject any agent channel open requests if they have not requested agent forwarding.

Since I don't actually know if OpenSSH considers this a requirement.

### DIFF
--- a/PROTOCOL
+++ b/PROTOCOL
@@ -344,6 +344,32 @@ signal to a session attached to a channel. OpenSSH supports one
 extension signal "INFO@openssh.com" that allows sending SIGINFO on
 BSD-derived systems.
 
+2.7. connection: Request agent forward "auth-agent-req@openssh.com"
+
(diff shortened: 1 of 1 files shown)


kalvdans (approved):

djmdjm:
This is documented in https://datatracker.ietf.org/doc/html/draft-miller-ssh-agent#name-forwarding-access-to-an-age

PR #379, opened 2023-02-26
linux,ssh: set IP_BIND_ADDRESS_NO_PORT on connected socket

crrodriguez:
if binding to a local address (ssh -b ...) the source port is taken at bind() time when the kernel does not know if the socket will be connect()ed or listen()ed on. It also does not know the destination host or port so it has to reserve the port until the socket is closed, effectively limiting the number of useful source ports to ~32k. a very small number in the modern era.

Set IP_BIND_ADDRESS_NO_PORT, which delays source port allocation to connect() time allowing a few million connections out from the same -b address.


sshconnect.c
```diff
@@ -392,6 +392,9 @@ ssh_create_socket(struct addrinfo *ai)
 		}
 		memcpy(&bindaddr, res->ai_addr, res->ai_addrlen);
 		bindaddrlen = res->ai_addrlen;
+#ifdef IP_BIND_ADDRESS_NO_PORT
+		setsockopt(sock, SOL_IP, IP_BIND_ADDRESS_NO_PORT, &(int) {1}, sizeof(int));
+#endif
 	} else if (options.bind_interface != NULL) {
```
kalvdans, line 398: The same option is useful if binding to an interface as well, could you rewrite so that both cases are handled?

kalvdans (approved):

sshconnect.c
crrodriguez: let me check that later.

sshconnect.c
```diff
@@ -391,6 +391,9 @@ ssh_create_socket(struct addrinfo *ai)
 		}
 		memcpy(&bindaddr, res->ai_addr, res->ai_addrlen);
 		bindaddrlen = res->ai_addrlen;
+#ifdef IP_BIND_ADDRESS_NO_PORT
```
djmdjm, line 394: Isn't this call unnecessary? Couldn't you have just the single `setsockopt()` call before the final `bind()` call?

kalvdans (approved):
Beautiful, short way to do the argument to setsockopt.

PR #383, opened 2023-03-14
chore: change comment

haochengxia:


daztucker:
Sorry but I don't think these changes add any value (and in the case of the second one, actually makes it less clear).

PR #384, opened 2023-03-15
Fix incomplete error check on getnameinfo()

nielsdos:
getnameinfo() can return a value < 0 on error, but only == -1 was checked. The other callsites of getnameinfo() perform the check correctly by checking != 0. Fix it by changing this check to also check != 0.

Found using a static analysis tool I'm developing.


djmdjm:
Good catch; merged upstream as 009eb4cb48

PR #387, opened 2023-03-17
Correctly match ECDSA curve.

pjd:
Fix similar to commit 5104db7.

https://github.com/openssh/openssh-portable/commit/5104db7cbd6cdd9c5971f4358e74414862fc1022

Obtained from:	Fudo Security


kalvdans:
Could you add a regression test case for this?

djmdjm:
Thanks, I merged a similar fix that handles certificate keys a bit better in 0a1e75499e2c

PR #390, opened 2023-04-04
ssh: suppress warning on identical names

eworm-de:
Running systemd-resolved on a linux host some synthetic records [0] are available, for example _gateway for all current default routing gateway addresses. Connecting to that name gives an error message:

> ignoring bad CNAME "_gateway" for host "_gateway": domain name "_gateway" starts with invalid character

Let's suppress the message if cononical name and cname are identical.

[0] https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html#Synthetic%20Records

### DIFF
--- a/ssh.c
+++ b/ssh.c
@@ -277,7 +277,8 @@ resolve_host(const char *name, int port, int logerr, char *cname, size_t clen)
 		return NULL;
 	}
 	if (cname != NULL && res->ai_canonname != NULL) {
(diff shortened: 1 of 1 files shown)


eworm-de:
Not quite sure this is the best and/or correct way to fix this. What exactly was the motivation for this error message?

djmdjm:
I think we should relax valid_domain() to allow names prefixed with underscores instead. I'll look at getting that merged.

PR #392, opened 2023-04-06
userauth_pubkey (client): remove duplicate checks

martelletto:
remove duplicate checks introduced with SSH_BUG_RSASIGMD5's removal in https://github.com/openssh/openssh-portable/commit/2a7e3449.


djmdjm:
It looks like the first of these check was removed some time since you posted this. I just removed the other one in 9c97b6af8e05 - thanks!

PR #393, opened 2023-04-17
ssh-keygen: generate, screen - arg for stdout

skull-squadron:
Although [2007 patch](https://lists.mindrot.org/pipermail/openssh-unix-dev/2007-March/025141.html) didn't take, it's inconsistent and a surprise these are missing.

Allows: ssh-keygen -M generate -O bits=4096 - | parallel --pipe -N1 ssh-keygen -M screen -f - - >> /etc/moduli


djmdjm:
Similar change merged in d9cd208e89a471a3 - thanks!

PR #397, opened 2023-05-12
gss-serv.c: `MAXHOSTNAMELEN` -> `HOST_NAME_MAX`

SoapGentoo:
`MAXHOSTNAMELEN` is not defined in POSIX, which breaks on musl: https://pubs.opengroup.org/onlinepubs/9699919799/functions/gethostname.html

Bug: https://bugs.gentoo.org/834044


thesamesam:
@daztucker Could you take a look? We're trying to reduce the number of downstream patches we've got in Gentoo and this is one of the easy wins, cheers

daztucker:
Applied, thanks.

PR #399, opened 2023-05-16
Handle the return value of atomicio as elsewhere.

rmsh1216:


rmsh1216:
> This change is incorrect, it will leak memory if the atomicio() call fails.
> 
> There is no reason to handle atomicio() failures here - if it fails then all that happens is the progress meter doesn't get displayed and so it is totally safe to ignore the return value.

Explictly ignoring return codes by using (void) may be better.
(void)atomicio (vwrite, STDOUT_FILENO, obuf, strlen (obuf));

PR #400, opened 2023-05-18
Add gnome-ssh-askpass4 for GNOME 40+

jtojnar:
This uses unstable Gcr API to get GNOME shell prompt.

Prompted by https://gitlab.gnome.org/GNOME/seahorse/-/issues/352 (GNOME Shell always asks for permission to grab input and since `gnome-ssh-askpass` does not have a desktop file, it will not be able to remember the permission.)

### DIFF
--- a/contrib/Makefile
+++ b/contrib/Makefile
@@ -1,7 +1,7 @@
 PKG_CONFIG = pkg-config
 
 all:
-	@echo "Valid targets: gnome-ssh-askpass1 gnome-ssh-askpass2 gnome-ssk-askpass3"
+	@echo "Valid targets: gnome-ssh-askpass1 gnome-ssh-askpass2 gnome-ssk-askpass3 gnome-ssk-askpass4"
 
 gnome-ssh-askpass1: gnome-ssh-askpass1.c
 	$(CC) $(CFLAGS) `gnome-config --cflags gnome gnomeui` \
@@ -17,6 +17,11 @@ gnome-ssh-askpass3: gnome-ssh-askpass3.c
 	$(CC) $(CFLAGS) `$(PKG_CONFIG) --cflags gtk+-3.0` \
 		gnome-ssh-askpass3.c -o gnome-ssh-askpass3 \
 		`$(PKG_CONFIG) --libs gtk+-3.0 x11`
+
+gnome-ssh-askpass4: gnome-ssh-askpass4.c
+	$(CC) $(CFLAGS) `$(PKG_CONFIG) --cflags gcr-4 gio-2.0` \
+		gnome-ssh-askpass4.c -o gnome-ssh-askpass4 \
+		`$(PKG_CONFIG) --libs gcr-4 gio-2.0`
 
 clean:
 	rm -f *.o gnome-ssh-askpass gnome-ssh-askpass[123]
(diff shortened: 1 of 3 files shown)


apollo13:
Hi Jan, I enabled this on my system and it already looks great. Do you think there is any chance of showing the full text though? It currently gets cut off:
image

jtojnar:
@apollo13 Thanks for testing. Maybe we should use `gcr_prompt_set_description` instead of `gcr_prompt_set_message`?

apollo13:
That would look like this:
image

Now it seems to click to long "words". Given that the ssh askpass prompts show the key fingerprint, this will be "long" and it would be great if it where readable, even if forcible broken up over multiple rows.

jtojnar:
> Given that the ssh askpass prompts show the key fingerprint, this will be "long" and it would be great if it where readable, even if forcible broken up over multiple rows.

Gcr offers no control about how this is displayed. We should probably report this against gnome-shell.

lsjostro:
Arch AUR has package this PR already 😄 :shipit: 
https://aur.archlinux.org/packages/gnome-ssh-askpass4-git

kirelagin:
It would also make sense to update the message that the `all` target echoes and the README (which, admittedly is already outdated as it does not include the gnome3 version).

jtojnar:
Thanks. Rebased and updated the `make all` message and `README`.

kirelagin:
I am also having this weird effect: the “Confirm user presence” message pops up first and then, after a fraction of a second, changes to “Enter PIN and confirm user presence”. I mean, it’s perfectly fine functionality, but this looks a bit weird and I am not sure why it is doing this...

djmdjm:
> Let’s instead use the GNOME Shell’s native prompt through the unstable Gcr API.

Is the GCR API still unstable? Once it's stable then let me know and I'll look at merging this

jtojnar:
I do not expect it to be stabilised anytime soon, the whole stack seems to be undermaintained.

Anyway, I do not think it should be a problem – it has not changed for years, and we can always fix it in the unlikely case the API changes. Either way, such breakage wouldn’t be much worse than the current state of the contrib helpers.

djmdjm:
Do you want to add you name to the copyright line in gnome-ssh-askpass4.c?

djmdjm:
I guess an alternative is to do a GTK4 port and include a .desktop file to avoid the permission thing? (edit: the alternatives are not exclusive)

jtojnar:
I followed the comment in askpass2:

https://github.com/openssh/openssh-portable/blob/f9dc519259804702cab0fa0ca8b193a360e3ec38/contrib/gnome-ssh-askpass2.c#L25

I can add a copyright comment if desired, it does not matter to me.

GTK 4 [removed support for grabs](https://docs.gtk.org/gtk4/migrating-3to4.html#stop-using-grabs), IIRC because Wayland does not have that concept due to its security model. So having a popup grab input might not be possible outside of compositors that support protocol extensions like [Hyprland](https://wayland.app/protocols/hyprland-focus-grab-v1).

djmdjm:
It's totally up to you, I just figured that since you did most of the work you should get at least some of the credit.

Anyway, I'll go ahead and commit this. If you decide you want your name there then I'm happy to add it.

djmdjm:
This has been merged. Thanks!

PR #402, opened 2023-05-26
command [argument ...] in ssh(1): a footgun

Artoria2e5:
* The argument functionality in ssh is a common place of misuse, as it does not actually preserve the boundary. We should discourage people from using it.
* Manual is currently vague about what shell is used to run the passed command. Spell it out.

See, well, https://marc.info/?l=openssh-unix-dev&m=168509072920594&w=2.

### DIFF
--- a/ssh.1
+++ b/ssh.1
@@ -94,7 +94,8 @@ their identity to the remote machine using one of several methods
 If a
 .Ar command
 is specified,
-it will be executed on the remote host instead of a login shell.
+it will be executed on the remote host (using the user's shell)
+instead of a login shell.
 A complete command line may be specified as
 .Ar command ,
 or it may have additional arguments.
--- a/ssh.c
+++ b/ssh.c
@@ -1129,6 +1129,10 @@ main(int ac, char **av)
 		}
 	} else {
 		/* A command has been specified.  Store it into the buffer. */
+		if (ac > 1)
+			fprintf(stderr,
+				"Warning: arguments given to command. "
+				"Concatenating with spaces.\n");
 		for (i = 0; i < ac; i++) {
 			if ((r = sshbuf_putf(command, "%s%s",
 			    i ? " " : "", av[i])) != 0)
--- a/ssh.c
+++ b/ssh.c
@@ -1116,33 +1116,6 @@ main(int ac, char **av)
 		fatal("sshbuf_new failed");
 
 	/*
-	 * Save the command to execute on the remote host in a buffer. There
-	 * is no limit on the length of the command, except by the maximum
-	 * packet size.  Also sets the tty flag if there is no command.
-	 */
-	if (!ac) {
-		/* No command specified - execute shell on a tty. */
-		if (options.session_type == SESSION_TYPE_SUBSYSTEM) {
-			fprintf(stderr,
-			    "You must specify a subsystem to invoke.\n");
-			usage();
-		}
-	} else {
-		/* A command has been specified.  Store it into the buffer. */
-		if (ac > 1)
-			fprintf(stderr,
-				"Warning: arguments given to command. "
-				"Concatenating with spaces.\n");
-		for (i = 0; i < ac; i++) {
-			if ((r = sshbuf_putf(command, "%s%s",
-			    i ? " " : "", av[i])) != 0)
-				fatal_fr(r, "buffer error");
-		}
-	}
-
-	ssh_signal(SIGPIPE, SIG_IGN); /* ignore SIGPIPE early */
-
-	/*
 	 * Initialize "log" output.  Since we are the client all output
 	 * goes to stderr unless otherwise specified by -y or -E.
 	 */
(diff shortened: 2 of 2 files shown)


Artoria2e5:
Well, ain't that inconvenient. The tests are flagging me for putting extra stuff in stderr. They aren't wrong per se...

ssh.c
```diff
@@ -1129,6 +1129,10 @@ main(int ac, char **av)
 		}
 	} else {
 		/* A command has been specified.  Store it into the buffer. */
+		if (ac > 1)
+			fprintf(stderr,
+				"Warning: arguments given to command. "
+				"Concatenating with spaces.\n");
```
djmdjm, line 1152: No sorry - many, many people use ssh with commandline arguments every day both manually and in scripts. This would just be noise.

Artoria2e5: I don't disagree that many people use it, but to me it still is a big footgun with a severity ranging from "use this example to scare newbies" to "throw semicolons at it and you might get a command injection".

There's also the more modest approach mentioned in the mailing list: we expand the manual page for examples of this stuff used "responsibly", whatever that could mean. That will take much more writing and is less visible.

Artoria2e5: We *could* do some extra scanning to avoid the basic `ssh x@x sh /etc/xxx` use cases. Like look for spaces, which is the most common failure case; or something like `;|&` which can launch extra commands. But simple scanning will still flag specially quoted arguments and "just use a single `command` parameter" would be a less obvious solution compared to what this message implies.

Earnestly:
Is it possible to fix this broken design by not squashing argv in the first place? `sh -c` allows arguments to be supplied so that they can be safely used as positional parameters within the command string without having to interpolate them as `ssh` requires.

Would it also be possible to just not wrap `ForceCommand` in `sh -c` at all?

Artoria2e5:
Hmmmm.... let me check my old email link to see what I considered back then. Okay, I'm mostly back on track.

Regarding `sh -c` + positional: currently this stuff is called via https://github.com/openssh/openssh-portable/blob/3ad669f81aabbd2ba9fbd472903f680f598e1e99/session.c#L1709. I didn't consider positional arguments back then, but I still don't think it would work without making what currently works fail. ssh don't just use `sh`, it uses the user's preferred login shell.  Adding positional arguments would add two requirements for the shell: [which currently does not need to be POSIX (it only needs to understand `-c` with a string)]

1. it needs to accept extra arguments after the `-c blahblah`, preferably with  `--` between "blahblah" and the rest.
2. Just `sh -c 'blahblah' -- a b c` doesn't work; you got to reference the positional args like `sh -c 'blahblah "$@" -- a b c`, and that `"$@"` is going to narrow down the scope of support a lot. 

Regarding ForceCommand: I am afraid not. The `ForceCommand` thing is probably (I didn't check) read from the config file as one big string, so it won't make sense to without a shell: `execve` wants a neat array of things.

I'm at least glad that someone's sharing my annoyance.

ssh.1
```diff
@@ -94,7 +94,8 @@ their identity to the remote machine using one of several methods
 If a
 .Ar command
 is specified,
-it will be executed on the remote host instead of a login shell.
+it will be executed on the remote host (using the user's shell)
```
Artoria2e5, line 97: Hm, wouldn't it be "executed using the user's login shell with  .Op c" instead?

Earnestly:
Oh yes, fixing this very, very old bug is going to break many uses of `ssh`.

When using `-c` with `sh` it is effectively taking the first argument as a command string (`-c` does not take a command string, it is a special mode) and the remaining as arguments starting from zero.

The problem openssh and others such as dropbear have is that they squash the command-line into a single string, obliterating any quoting information requiring people escape shell input in an attempt safely interpolate/inject arbitrary input.

djmdjm:
Sorry, but we don't plan to change ssh/sshd's behaviour here.

Changing the quoting and use of shell for execution would break many working scripts that depend on how it currently works, which has been stable for ~25 years and even longer if you consider ssh as a successor to rsh.

Changing how arguments are passed to the server requires a SSH protocol extension too, as the specified protocol just uses a string.

Earnestly:
> Changing the quoting ...

For what it's worth, this is not changing quoting. It's about changing how arguments are passed to commands and it would benefit every command, not just shells.

> Changing how arguments are passed to the server requires a SSH protocol extension too, as the specified protocol just uses a string.

Yes, maybe it would require more work but is a really terrible situation where everyone[1] uses openssh insecurely by default, including many orchestration wrappers, which exposes them to code injection for no good reason besides vestigial design.

*1. There are a few, very few, who know how - and even realise the need - to quote arguments according to not only a potential shell but also ForceCommands.*

I fear that ultimately your conclusion of WONTFIX because old is likely to shared broadly, even if it was the addition of a new flag that wouldn't interfere with existing code.

Artoria2e5:
... and to add to the point, a manual section never broke anything...

PR #403, opened 2023-05-31
Add wtmpdb support as Y2038 safe wtmp replacement

thkukuk:
wtmp support as implemented in glibc is not Y2038 safe and glibc developers don't plan to fix that 
(see https://www.thkukuk.de/blog/Y2038_glibc_utmp_64bit/ and https://www.thkukuk.de/blog/Y2038_glibc_wtmp_64bit/)

This patch adds support for wtmpdb, a Y2038 safe alternative to wtmp.
For most applications the PAM module is working fine and good enough, but since sshd does not know the TTY yet when calling PAM, native support for wtmpdb is necessary to include the TTY in the data, too.

### DIFF
--- a/.github/configs
+++ b/.github/configs
@@ -266,6 +266,10 @@ case "${TARGET_HOST}" in
 	# Native linker is not great with PIC so OpenSSL is built w/out.
 	CONFIGFLAGS="${CONFIGFLAGS} --disable-security-key"
 	;;
+    fbsd14-ppc64)
+	# Disable security key tests for bigendian interop test.
+	CONFIGFLAGS="${CONFIGFLAGS} --disable-security-key"
+	;;
     hurd)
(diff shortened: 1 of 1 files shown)


djmdjm:
Aren't systems moving to 32-bit time_t? That would (backwards-incompatibly) change the wtmp format too, right?

thkukuk:
> Aren't systems moving to 32-bit time_t? That would (backwards-incompatibly) change the wtmp format too, right?

Nobody will move to 32bit time_t. And wtmpdb is the solution to move to 64bit time_t as everybody is doing today.
And yes, switching to 64bit time_t does break current wtmp API and on disk data, that's why wtmpdb got introduced, so solve that and many more problems of wtmp.

thkukuk:
Sorry, somehow I messed this up, there is a new PR#540

PR #404, opened 2023-06-01
Security Key: allow specifying a prompt command.

ebardsley:
If if specified, this will be executed instead of printing the default prompt to confirm user presence.

_Note: I'm using this to implement a graphical notifier to prompt to confirm user presence, which IMO clutters up the terminal less and doesn't get lost when ssh is called as a subprocess that redirects stderr. It also has the pleasant side effect of getting rid of the "User presence confirmed" message._

_I'm not entirely sure that this is the right way to do regression testing for this (it does pass `make LTESTS=sk-prompt t-exec`) nor am I entirely certain what the original intent of notify_start's interaction with SSH_ASKPASS was, since it seems very hard to trigger, and perhaps should be refactored. Please advise if you'd like this done differently._

### DIFF
--- a/misc.h
+++ b/misc.h
@@ -228,8 +228,8 @@ struct notifier_ctx;
 
 char	*read_passphrase(const char *, int);
 int	 ask_permission(const char *, ...) __attribute__((format(printf, 1, 2)));
-struct notifier_ctx *notify_start(int, const char *, ...)
-	__attribute__((format(printf, 2, 3)));
+struct notifier_ctx *notify_start(const char *, int, const char *, ...)
+	__attribute__((format(printf, 3, 4)));
 void	notify_complete(struct notifier_ctx *, const char *, ...)
 	__attribute__((format(printf, 2, 3)));
 
--- a/readconf.c
+++ b/readconf.c
@@ -173,8 +173,8 @@ typedef enum {
 	oStreamLocalBindMask, oStreamLocalBindUnlink, oRevokedHostKeys,
 	oFingerprintHash, oUpdateHostkeys, oHostbasedAcceptedAlgorithms,
 	oPubkeyAcceptedAlgorithms, oCASignatureAlgorithms, oProxyJump,
-	oSecurityKeyProvider, oKnownHostsCommand, oRequiredRSASize,
-	oEnableEscapeCommandline,
+	oSecurityKeyPromptCommand, oSecurityKeyProvider, oKnownHostsCommand,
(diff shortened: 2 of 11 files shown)


ebardsley:
That CI failure on Mac 11/12 is disappointing, especially since I wrote this on MacOS 12. Any insight into what's different about those test runners would be appreciated.

djmdjm:
IMO this would be less intrusively achieved by finessing the way that the askpass program itself is used. E.g. https://github.com/djmdjm/openssh-wip/pull/21

ebardsley:
> IMO this would be less intrusively achieved by finessing the way that the askpass program itself is used. E.g. [djmdjm/openssh-wip#21](https://github.com/djmdjm/openssh-wip/pull/21)

I patched that and tested it and it seems to also accomplish what I'm looking for - allowing a notify-specific "askpass" that's not strongly tied to either `BatchMode` or `DISPLAY`

I do think there's some advantage to configuration directives and the associated `Host/Match`-based customization, but if you're willing to commit the env-based version upstream, that works for me!

djmdjm:
So one thing I've been wanting to do is extending the askpass interface to be more modern and flexible. In particular, I'd like to pass structured information to the askpass program, e.g. the endpoint address, username, key in question, etc.

Obviously I haven't got as far as a design for this yet :(

ebardsley:
FWIW, my current user presence "askpass" is the following wrapper around [terminal-notifier](https://github.com/julienXX/terminal-notifier) which throws up a bubble that disappears once presence is confirmed. Although not strictly necessary, being able to customize the message a bit more than the current "Confirm user presence for key ..." message (or a fixed string) seems appropriate. Perhaps more environment variables ala `SSH_ASKPASS_PROMPT` to the subprocess would be useful for that information, eg `SSH_ASKPASS_{HOST,USER,KEY}` etc

```
#!/bin/bash

trap "terminal-notifier -remove ssh >/dev/null 2>&1" EXIT
terminal-notifier -title ssh -message "$*" -group ssh
while true; do sleep 100; done
```

ebardsley:
@djmdjm It sounded like you were going to implement your `SSH_ASKPASS_REQUIRE` feature instead of this PR.  Please let me know if you'd like me to do something with this PR, otherwise I'll just leave it sitting here and close it once the askpass feature makes it in

djmdjm:
I honestly haven't decided - happy to leave both open until I have more time to think about it

PR #406, opened 2023-06-14
pkcs11: Do not trim last character of fixed-length fields

Jakuje:
The function rmspace trimmed last character of all the fixed-length fields causing inaccurate information being reported to the user when the whole buffer was used by the pkcs11 library. This is common for the serial number and very unlike for other fields such as manufacturer ID.

This commit changes the rmspace function to not trim the last character. As a consequence the resulting buffers might not be null terminated so we modify the debug prints to make sure they do not go over the buffer limits by explicitly using precision modifiers.


djmdjm:
Fixed in 6958f00acf3

PR #410, opened 2023-07-04
Ensure strtonum returns a number that is always positive in int range

AtariDreams:
We should be using INT_MAX, not UINT_MAX, and checking for errors via errstr, as opposed to allowing negative numbers.


ssh-add.c
```diff
@@ -863,16 +864,16 @@ main(int argc, char **argv)
 			confirm = 1;
 			break;
 		case 'm':
-			minleft = (int)strtonum(optarg, 1, UINT_MAX, NULL);
```
daztucker, line 866: > We should be using INT_MAX, not UINT_MAX

minleft (and maxsign) are u_int, so the existing limit is correct (although the cast is wrong).

> checking for errors via errstr, as opposed to allowing negative numbers.

It won't allow negative numbers since the minimum value specified to strtonum is "1", thus if it returns zero it is unambiguously an error.  In fact given that it's stored in a u_int it'll probably even work for values >INT_MAX, although we should change the cast.

AtariDreams:
@daztucker Fixed!

daztucker:
I've applied the cast change, but as I said above I do not see the need for the rest of this change since the existing error check is unambiguous.  Can you give an example of when the existing code does the incorrect thing?

daztucker:
You did not address my previous comment: " I do not see the need for the rest of this change since the existing error check is unambiguous. Can you give an example of when the existing code does the incorrect thing?"

PR #413, opened 2023-07-05
Remove dead code in pack_public_key_ed25519

AtariDreams:
if (ret != 0)
		free(response->public_key);

is dead code; the very last operation before the memcpy and the setting of ret to 0 is the malloc of response->public_key, which means response->public_key is always NULL if pack_public_key_ed25519 fails.


djmdjm:
I'd rather not remove this - if the code subsequently changes to add another operation then it would be easy to introduce a memleak on error

PR #416, opened 2023-07-05
Prefer mkdtemp over mktemp

AtariDreams:
mktemp is deprecated and potentially dangerous.


djmdjm:
this is an import of openbsd's `src/usr.bin/nc` - IMO you'd be better off proposing this change there and we'll sync it back

PR #419, opened 2023-07-16
Prefer fputc over fputs where possible

AZero13:
fputc is meant for single characters, fputs is for strings.


kalvdans (commented):
I see nothing wrong with the patch, just noting that gcc already does this rewrite at compile-time. Sticking to always calling fprintf instead of mixing function calls enhances readability in my opinion.

PR #422, opened 2023-07-28
remove unnecessary if statement

eyalasulin999:
in file `sshconnect2.c`, function `userauth_pubkey`
checking twice if `id->key` is not NULL


daztucker:
Applied upstream, thanks.

PR #425, opened 2023-08-07
regress/key-options.sh: update future key expiry date to far in the future

kanavin:
This allows testing Y2038 with system time set to after that (i.e.2040), so that actual Y2038 issues can be exposed, and not masked by key expiry time errors.

### DIFF
--- a/regress/key-options.sh
+++ b/regress/key-options.sh
@@ -120,5 +120,5 @@ check_valid_before() {
 check_valid_before "default"	""				"pass"
 check_valid_before "invalid"	'expiry-time="INVALID"'		"fail"
 check_valid_before "expired"	'expiry-time="19990101"'	"fail"
-check_valid_before "valid"	'expiry-time="20380101"'	"pass"
+check_valid_before "valid"	'expiry-time="25250101"'	"pass"
 


kanavin:
Ping, please.

djmdjm:
Does this work properly on systems with 32bit time_t?

kanavin:
> Does this work properly on systems with 32bit time_t?

On such systems it fails, predictably:
```
key option expiry-time default
key option expiry-time invalid
key option expiry-time expired
key option expiry-time valid
key option failed valid
```

There are two ways to interpret that:

- systems with 32 bit time_t are a ticking time bomb which will spectacularly explode in 2038, and this failure highlights that they must be upgraded or replaced ASAP.

- there are legitimate reasons to keep running such systems for a few more years, and they should be on life support for a while yet, which means not using dates that they can't represent.

I can see both sides to this. If you prefer first option, the patch can be taken as it is. If you prefer the second option, then the test should be using expiry in 2038 if time_t is 32 bit, and expiry in 2525 if time_t is bigger. I can implement this, but not sure how, and would appreciate suggestions.

daztucker:
Applied, but instead of replacing the test I added your new one, configured to be skipped on platforms that still have a 32bit time_t.

kanavin:
Thanks, but unfortunately this is not completely what I'd like to have. We should also be skipping the original test (where expiry is set to 2038) on 64 bit time_t platforms, so that such platforms can be tested already now with the date set to the post-2038 future. That aspect was lost with your changes :(

I'll make a new PR.

kanavin:
New PR here https://github.com/openssh/openssh-portable/pull/544

PR #428, opened 2023-08-11
lim cannot be less than zero because rlim_t is usually treat as an un…

rmsh1216:
…signed type


djmdjm:
No, rlim_t is signed on some platforms

daztucker (commented):
POSIX[1] says rlim_t is an unsigned type so it shouldn't be negative.  It could be zero, though, and you're removing that check.

[1] https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/resource.h.html

djmdjm:
POSIX might say that, but FreeBSD has signed rlim_t https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=209729#c1

PR #429, opened 2023-08-16
Artificial delays preventing user enumeration timing attacks...

beldmit:
...should not happen in case when auth method is none because it doesn't reveal any new information and should have an upper limit.

Othervise in case of bad network it tends to increase drastically.


djmdjm:
This got merged a while ago

PR #430, opened 2023-08-17
protocol: Fix type also for last privatekeyN

Jakuje:
The key specification was wrongly marking all keys as strings until 24fee8973abdf1c521cd2c0047d89e86d9c3fc38, which fixed this for key1 and key2, but not for the keyN.

This is mostly cosmetic fix as the OpenSSH supports only one key in a file, but it can take some time to wrap a head around this to verify which one is correct.


djmdjm:
Merged, thanks

PR #434, opened 2023-08-30
Add support for the special ${RANDOM} environment variable.

erjoalgo:
This can be used to generate strings unique to each ssh connection. For example, it can be used to generate unique RemoteForward unix socket paths to avoid conflicts between multiple connections to a given remote host.


erjoalgo:
For more details of this feature request: https://bugzilla.mindrot.org/show_bug.cgi?id=3609

misc.c
```diff
@@ -1301,7 +1301,17 @@ vdollar_percent_expand(int *parseerror, int dollar, int percent,
 			}
 			var = xmalloc(len + 1);
 			(void)strlcpy(var, string, len + 1);
-			if ((val = getenv(var)) == NULL) {
+			char RANDOM[7];
```
daztucker, line 1304: OpenSSH is compilable with a C89 compiler, and those don't allow declarations after code, which this is.  Our c89 test should have caught this but didn't, for a reason I've just fixed.

Also, all-caps things are macros not variables.  See the OpenBSD style guide: https://man.openbsd.org/style

misc.c
```diff
@@ -1301,7 +1301,17 @@ vdollar_percent_expand(int *parseerror, int dollar, int percent,
...
+			val = NULL;
+			if (strcmp(var, "RANDOM") == 0) {
+				/*
+				 * Generate a random 5-digit number to support the
+				 * special ${RANDOM} environment variable.
+				 */
+				sprintf(RANDOM, "%d", arc4random_uniform(1e6));
```
daztucker, line 1311: sprintf is not bounds-checked.  Please use snprintf instead.

daztucker, line 1311: Using arc4random here adds an indirect dependency on OpenSSL's libcrypto.  We put some effort into avoiding that dependency for binaries like scp, sftp and sftp-server which is why many of the Github Actions tests fail.

This could be avoided by moving this function into one of the files only linked into ssh and sshd (but see above about whether or not this approach is the right way to solve this, leaving aside the implementation).

misc.c
```diff
@@ -1301,7 +1301,17 @@ vdollar_percent_expand(int *parseerror, int dollar, int percent,
...
+				val = (char*)&RANDOM;
```
daztucker, line 1312: val is a char* and RANDOM is a char[] so you shouldn't need the casts and &

misc.c
erjoalgo: I replied in the bugzilla bug with an attempt to clarify the reasons why this feature would be useful for me, I understand this is still a discussion.

Thanks for helping me understand why some of those github action tests fail. I'm not very familiar with the codebase so if you could point me into what you think would be the right file where to move this function that would be very helpful.

misc.c
erjoalgo: Fixed.

misc.c
```diff
@@ -1301,7 +1301,17 @@ vdollar_percent_expand(int *parseerror, int dollar, int percent,
 			}
 			var = xmalloc(len + 1);
 			(void)strlcpy(var, string, len + 1);
-			if ((val = getenv(var)) == NULL) {
+			char RANDOM[7];
+			val = NULL;
+			if (strcmp(var, "RANDOM") == 0) {
+				/*
+				 * Generate a random 5-digit number to support the
```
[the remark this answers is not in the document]
erjoalgo: The RANDOM range used in both bash and ksh is actually [less](https://man.openbsd.org/ksh.1#RANDOM) than the full 5 digits: 

> [RANDOM](https://man.openbsd.org/ksh.1#RANDOM)
> A random number generator. Every time RANDOM is referenced, it is assigned the next random number in the range 0-32767

A range of 5 digits is more than 100x the number of days of the year, personally the bash ${RANDOM} facility with an even smaller range has worked well for me in many situations. But if you have a suggestion for how many digits should be used here I will be happy to update it.

misc.c
erjoalgo: Fixed.

misc.c
erjoalgo: Thanks, fixed.

PR #435, opened 2023-09-02
enable ssh connection to a unix socket

oliverkurth:
This is based on a PR #431 and #162 which enables connecting to a unix socket where `getaddrinfo(3)` supports `AF_UNIX` and adds connecting to a unix socket where it is not (which is the case for MacOS and Linux and probably most others). This is achieved by changing `resolve_host` to return a faked up `struct addrinfo` for a unix socket, when the "hostname" starts with a '/'. The code is partially reused from `channels.c` / `connect_to_helper()`.

This also enforces `HashKnownHosts` when a unix socket is used, to prevent corrupting the `known_hosts` file with special characters - file paths can contain spaces, `!`, `*` and others which are not allowed in host names but do have special meaning in `known_hosts`.

Tested successfully with MacOS 13.5 (using gcc 13) and Ubuntu 22.04.

Example usage:
```
ssh -A okurth@/run/sshd.sock
```

There are multiple ways to set up a ssh server listening on a unix socket. One is setting
```
[Socket]
ListenStream=/var/run/sshd.sock
Accept=yes
```
in a systemd socket config (for example `/etc/systemd/system/ssh.socket`).

Other ways are using `socat`/`netcat`, or port forwarding with `ssh`, using a unix socket on the listening end and TCP on the connecting end.

- [ ] amend man page
- [ ] automatic testing


sshconnect.c
```diff
@@ -469,19 +470,20 @@ ssh_connect_direct(struct ssh *ssh, const char *host, struct addrinfo *aitop,
...
 			if (ai->ai_family != AF_INET &&
-			    ai->ai_family != AF_INET6) {
+			    ai->ai_family != AF_INET6 &&
+			    ai->ai_family != AF_UNIX) {
 				errno = EAFNOSUPPORT;
 				continue;
 			}
-			if (getnameinfo(ai->ai_addr, ai->ai_addrlen,
+			if (ai->ai_family != AF_UNIX && getnameinfo(ai->ai_addr, ai->ai_addrlen,
 			    ntop, sizeof(ntop), strport, sizeof(strport),
```
kalvdans, line 479: getnameinfo supports AF_UNIX on glibc. It returns localhost in ntop and the path in strport.

kalvdans: Perhaps the syntax to connect to a unix socket should be "ssh localhost -p /path/to/socket"? It makes some sense that the path goes into the "port" part instead of the host.

ssh.c
```diff
@@ -257,6 +258,31 @@ resolve_host(const char *name, int port, int logerr, char *cname, size_t clen)
 	int gaierr;
 	LogLevel loglevel = SYSLOG_LEVEL_DEBUG1;
 
+	if (name[0] == '/') {
+		struct sockaddr_un *sunaddr;
+
+		if (strlen(name) > sizeof(sunaddr->sun_path)) {
```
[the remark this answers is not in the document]
oliverkurth: Oh, good catch! That needs fixing in `channels.c` / `connect_to_helper()` too then - I copied from there. I guess that needs a separate PR. Do you want to create one for that?

sshconnect.c
oliverkurth: > getnameinfo supports AF_UNIX on glibc. 

Hm, maybe not on MacOS, and that's why I changed it. I'll check that later.

> Perhaps the syntax to connect to a unix socket should be "ssh localhost -p /path/to/socket"? It makes some sense that the path goes into the "port" part instead of the host part.

I think it would be confusing. It's not used that way for the `-L` and `-R` options. And I am not sure how that would work for the `HostAlias` option.

ssh.c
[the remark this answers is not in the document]
kalvdans: Yep! It wouldn't be very BSD-ish to go back to use strncpy to win one more character in the path. New PR fixing all occurrances at https://github.com/openssh/openssh-portable/pull/436 .

sshconnect.c
kalvdans: Yes, you are probably right. We should never talk about host/port when it comes to AF_UNIX. So that's why relying on getnameinfo is bad even if it happen to support AF_UNIX. Resolving this discussion thread.

ssh.c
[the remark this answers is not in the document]
oliverkurth: It's actually worse than that, and has nothing to do with `strlcpy()`: `sun_path` needs to be terminated with a `NUL` character, so that has to be counted. If not, anything trying to access it as a string would run into the memory that follows.

sshconnect.c
oliverkurth: I made small change so it prints correctly prints out "Connecting to..." when debugging.

ssh.c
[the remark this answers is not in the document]
oliverkurth: Fixed.

sshconnect.c
```diff
@@ -500,7 +506,7 @@ ssh_connect_direct(struct ssh *ssh, const char *host, struct addrinfo *aitop,
 				break;
 			} else {
 				oerrno = errno;
-				debug("connect to address %s port %s: %s",
+				debug("connect to %s%s: %s",
 				    ntop, strport, strerror(errno));
```
kalvdans, line 510: Adapt this error message for unix paths as well.

sshconnect.c
```diff
@@ -513,7 +519,7 @@ ssh_connect_direct(struct ssh *ssh, const char *host, struct addrinfo *aitop,
 
 	/* Return failure if we didn't get a successful connection. */
 	if (sock == -1) {
-		error("ssh: connect to host %s port %s: %s",
+		error("ssh: connect to %s%s: %s",
 		    host, strport, errno == 0 ? "failure" : strerror(errno));
```
kalvdans, line 523: Adapt error message for unix paths.

kalvdans (changes requested):

sshconnect.c
oliverkurth: It's not needed here. For a UNIX socket, `strport` will be an empty string, so `%s%s` will just print what's in `host`, which contains the path. See:
```
% /usr/local/bin/ssh okurth@/foo.sock
ssh: connect to /foo.sock: No such file or directory
```

sshconnect.c
oliverkurth: Fixed.

sshconnect.c
kalvdans: Yes, but it is broken for IP addresses:
```
$ ./ssh user@localhost -p 3333
ssh: connect to localhost3333: Connection refused
```

kalvdans (commented):
Test works but spams a non-fatal error message:
```
$ make t-exec TEST_SSH_UNSAFE_PERMISSIONS=1
[...]
run test connect-unix.sh ...
/home/chn/repo/openssh-portable/regress/test-exec.sh: 32: kill: No such process

ok simple connect
```

Thanks a lot for writing a test!

sshconnect.c
oliverkurth: Oh, I missed that, thanks! This is from one of pjd's original changes.

And fixed, with 5e813709

kalvdans: I think pjd's original changes were fine, but later destroyed by my attempts to simplify it :/ Thanks for restoring!

sshconnect.c
```diff
@@ -500,8 +506,13 @@ ssh_connect_direct(struct ssh *ssh, const char *host, struct addrinfo *aitop,
...
 			} else {
 				oerrno = errno;
-				debug("connect to address %s port %s: %s",
-				    ntop, strport, strerror(errno));
+				if (ai->ai_family != AF_UNIX) {
+					debug("connect to %s port %s: %s",
+					    ntop, strport, strerror(errno));
+				} else {
+					debug("connect to %s: %s",
+					    host, strerror(errno));
```
kalvdans, line 514: Remove these three lines, as the same message will anyway be printed as an error immediatelly after. (for unix domain sockets, we always have exactly one entry in the struct addrinfo linked list)

oliverkurth: I'd prefer to leave it there, for symmetry reasons.

oliverkurth:
Alright, test is mostly working now, except two jobs:
* CIFuzz / Fuzzing fails with:
```
+ clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -std=c++11 -DCIPHER_NONE_AVAIL=1 -I. -L. -Lopenbsd-compat -g regress/misc/fuzz-harness/agent_fuzz.cc -o /github/workspace/build-out/agent_fuzz sk-dummy.o agent_fuzz_helper.o ssh-sk.o -lssh -lopenbsd-compat -lz -Wl,-Bstatic -lcrypto -Wl,-Bdynamic -fsanitize=fuzzer
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x25
agent_fuzz_helper.o: in function `process_add_smartcard_key':
agent_fuzz_helper.c:(.text.process_add_smartcard_key[process_add_smartcard_key]+0xdb6): undefined reference to `pkcs11_make_cert'
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
2023-12-31 03:26:07,809 - root - ERROR - Building fuzzers failed.
2023-12-31 03:26:07,809 - root - ERROR - Error building fuzzers for (commit: 2b76606bae5cf45aa14de283e7778a811804ef33, pr_ref: refs/pull/435/merge).
```
I don't think it's related.

* ubuntu-20.04, valgrind-1 is related to the change, and fails with:
```
run test connect-unix.sh ...
Extra argument OBJ=/home/runner/work/openssh-portable/openssh-portable/regress.
FATAL: no sshd.socket created
```
But I don't know how to fix this.

regress/connect-unix.sh
```diff
@@ -0,0 +1,45 @@
+tid="simple connect"
+
+SCRIPT_DIR=`dirname $0`
+
+start_sshd_unix()
+{
+	${SUDO} env NC=${NC} SSHD=${SSHD} OBJ=${OBJ} ${SCRIPT_DIR}/sshd-unix.sh&
```
kalvdans, line 7: One reason for the CI error `FATAL: no sshd.socket created` is because if a timing issue. Here we run the whole script `sshd-unix.sh` in the background, which first cleans up `sshd.socket` before starting netcat to create a new file. This creates a timing window where the wait for creating of `sshd.socket` finishes too early. I suggest the following change:
```diff
modified   regress/connect-unix.sh
@@ -4,7 +4,7 @@ SCRIPT_DIR=`dirname $0`
 
 start_sshd_unix()
 {
-	${SUDO} env NC=${NC} SSHD=${SSHD} OBJ=${OBJ} ${SCRIPT_DIR}/sshd-unix.sh&
+	${SUDO} env NC=${NC} SSHD=${SSHD} OBJ=${OBJ} ${SCRIPT_DIR}/sshd-unix.sh
 
 	trace "wait for sshd-unix"
 	i=0;
modified   regress/sshd-unix.sh
@@ -7,7 +7,7 @@ run_sshd_unix() {
 	rm -f ${OBJ}/sshd.fifo && mkfifo ${OBJ}/sshd.fifo
 	rm -f ${OBJ}/sshd.socket
 	touch ${OBJ}/sshd-unix.log ; chmod 0644 ${OBJ}/sshd-unix.log
-	cat ${OBJ}/sshd.fifo | ${SSHD} -i -f ${OBJ}/sshd_config "$@" -E ${OBJ}/sshd-unix.log | ${NC} -l -U ${OBJ}/sshd.socket > ${OBJ}/sshd.fifo
+	cat ${OBJ}/sshd.fifo | ${SSHD} -i -f ${OBJ}/sshd_config "$@" -E ${OBJ}/sshd-unix.log | ${NC} -l -U ${OBJ}/sshd.socket > ${OBJ}/sshd.fifo &
 }
 
 run_sshd_unix
```

I haven't been able to run the test suite in full on my own machine, so I can't tell if this fixes the CI error. But at least it should be more logically correct.

oliverkurth: Thanks! I just pushed a change with that fix.

oliverkurth:
@svmhdvn I am using this on a daily basis, and I find it very useful (I use it to connect to a VM running under VMware Fusion, without having to bother with the network).

Unfortunately, I did not get any response from the maintainers, which has been very disappointing. But maybe I haven't reached out to the right people.

svmhdvn:
I agree, I think this is a very useful PR, I'd like to help push this so that we don't have to maintain local patches to security-sensitive software like this. @djmdjm who is the right maintainer to take a look at this? It seems to be in a good working state.

Caio99BR:
Any update on this?

oliverkurth:
No. If anyone has an idea how to move this forward, please let me know.

vt-alt:
The feature is very interesting as many people want to avoid exposing ssh to the network for use cases of connecting to sandboxed or containerized environments. Availability of ssh protocol allows other tools supporting ssh to integrate with them natively.

Caio99BR:
@oliverkurth can you rebase it to fix conflicts?

oliverkurth:
> @oliverkurth can you rebase it to fix conflicts?

Done.

I squashed all changes, and dropped the test changes - the test suite is a mess, I'll stay out of that.

PR #436, opened 2023-09-03
Always leave a trailing zero at the end of sun_path

kalvdans:
It simplifies things, being able to check length with strlen, and allow copying with strlcpy.


readconf.h
```diff
@@ -20,7 +20,6 @@
 
 #define SSH_MAX_HOSTS_FILES	32
 #define MAX_CANON_DOMAINS	32
-#define PATH_MAX_SUN		(sizeof((struct sockaddr_un *)0)->sun_path)
```
oliverkurth, line 23: Why did you move this?

kalvdans: It was only used internally in the corresponding C file. It is good to keep header files small, to avoid polluting the namespace and to avoid excess recompilation in case the define is changed.

kalvdans:
I need some help to reproduce the CI failures. I could not reproduce on a mac at work:
```
insight@mac openssh-portable % mkdir -p /tmp/usr/local/etc/
insight@mac openssh-portable % ./ssh-keygen -A -f /tmp     
ssh-keygen: generating new host keys: ED25519 
insight@mac openssh-portable % sw_vers -productVersion                 
13.4
```
But then, the CI environment is quite different. Can you try on your end, @oliverkurth ?

oliverkurth:
> But then, the CI environment is quite different. Can you try on your end, @oliverkurth 

I saw that, and tried to reproduce on my Mac, but it works fine for me too.

The CI is a mess, and I have failures for the unix socket PR that I am not able to reproduce. It's very frustrating.

kalvdans:
> > But then, the CI environment is quite different. Can you try on your end, @oliverkurth
> 
> I saw that, and tried to reproduce on my Mac, but it works fine for me too.

Thanks a lot for trying to reproduce!

> The CI is a mess, and I have failures for the unix socket PR that I am not able to reproduce. It's very frustrating.

I force-pushed to re-trigger CI, but the failure is consistent. Good that it isn't intermittent at least :)

oliverkurth:
Looks like the segfault in `ssh-keygen` on MacOS is a known issue: https://bugzilla.mindrot.org/show_bug.cgi?id=3584

It's not your fault 🙂

PR #437, opened 2023-09-06
openssh regress/Makefile: print logs if test fails

mikkorapeli-linaro:
Some tests are failing in yocto project CI runs uner qemu and reproduction has failed. Print the captured sshd and ssh client logs if test fails. This should help to fix the root causes.

Reference: https://bugzilla.yoctoproject.org/show_bug.cgi?id=15178

### DIFF
--- a/regress/Makefile
+++ b/regress/Makefile
@@ -237,7 +237,7 @@ t-exec:	${LTESTS:=.sh}
 		done; \
 		if [ "x$${skip}" = "xno" ]; then \
 			echo "run test $${TEST}" ... 1>&2; \
-			(env SUDO="${SUDO}" TEST_ENV=${TEST_ENV} ${TEST_SHELL} ${.CURDIR}/test-exec.sh ${.OBJDIR} ${.CURDIR}/$${TEST}) || exit $$?; \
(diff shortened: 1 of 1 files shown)


daztucker:
That Makefile is also used by OpenBSD's test infrastructure so I don't want to change that behaviour.

Our CI test driver already outputs the logs on failure (https://github.com/openssh/openssh-portable/blob/694150ad92765574ff82a18f4e86322bd3231e68/.github/run_test.sh#L23)  and saves them as a build artifact. (https://github.com/openssh/openssh-portable/blob/694150ad92765574ff82a18f4e86322bd3231e68/.github/workflows/c-cpp.yml#L117).  You could do either of those things.

mikkorapeli-linaro:
Ok, thanks for the tip.

With this patch applied the yocto project testing with qemu managed to capture a failure which is quite puzzling for us:

https://bugzilla.yoctoproject.org/show_bug.cgi?id=15178
-> http://autobuilder.yocto.io/pub/non-release/20230907-29/testresults/qemux86-64-ptest/openssh.log
or https://bugzilla.yoctoproject.org/attachment.cgi?id=4973

It looks like the ssh data socker is closed before 0 and 100 size banners sent out by the server? Could it be a timing/scheduling issue on high CPU/memory/IO load with qemu?

PR #440, opened 2023-09-09
sshd.c: Prevent infinite loop in ppoll()

deleriux:
The `ppoll` call in `sshd.c` derives its `nfds` value from options->max_startups.

However, if this value is above the processes' RLIMIT_NOFILE soft limit, in Linux the `ppoll` call errors out with -EINVAL.

This is not caught correctly and produces an uncontrolled infinite loop that leaves ssh in a unprocessable state for new connections.

To avoid this, both check RLIMIT_NOFILE and set `rlim_cur` to ensure that `sshd` can correctly `ppoll`.

A fatal error is returned if the user tries to set a MaxStartups above the `rlim_max` setting.

### DIFF
--- a/sshd.c
+++ b/sshd.c
@@ -47,6 +47,7 @@
 #include <sys/types.h>
 #include <sys/ioctl.h>
 #include <sys/socket.h>
+#include <sys/resource.h>
 #ifdef HAVE_SYS_STAT_H
 # include <sys/stat.h>
 #endif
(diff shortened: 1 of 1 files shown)


djmdjm:
> This is not caught correctly and produces an uncontrolled infinite loop that leaves ssh in a unprocessable state for new connections.

How is it not caught correctly? Since 379b30120, this has been the ppoll() call in sshd: 
```c
/* Wait until a connection arrives or a child exits. */
ret = ppoll(pfd, npfd, NULL, &osigset);
if (ret == -1 && errno != EINTR) {
        error("ppoll: %.100s", strerror(errno));
        if (errno == EINVAL)
                cleanup_exit(1); /* can't recover */
}
```

I don't see how this could infinitely loop.

deleriux:
No, you're right. I obviously didn't check this code in the latest release from the one I found this in!

deleriux:
Closing as it's an already fixed bug.

PR #443, opened 2023-09-15
Expose SSH_AUTH_INFO_0 always to PAM auth modules.

eehakkin:
The commit e8f474554e3bda102a797a2fbab0594ccc66f097 exposed SSH_AUTH_INFO_0 to PAM auth modules when a keyboard-interactive authentication method is in use. This does the same when a password authentication method is in use.


djmdjm:
This looks sensible to me. @daztucker any thoughts?

djmdjm:
merged - thanks

PR #445, opened 2023-09-27
[Draft] Using EVP_PKEY in OpenSSH

beldmit:
An attempt to use EVP_PKEY instead of RSA/DSA/EC structures


dh.c
```diff
@@ -283,6 +288,103 @@ dh_pub_is_valid(const DH *dh, const BIGNUM *dh_pub)
 int
 dh_gen_key(DH *dh, int need)
 {
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
```
bob-beck, line 291: If you are just going to switch over to EVP_PKEY, this should be possible without wholesale code duplication for OpenSSL 3.0 -  The EVP_pkey interface isn't significantly different between 1.1.1 and 3.0.

kexdh.c
```diff
@@ -200,4 +302,48 @@ kex_dh_dec(struct kex *kex, const struct sshbuf *dh_blob,
 	sshbuf_free(buf);
 	return r;
 }
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
```
bob-beck, line 305: Again, this is bringing in OpenSSL 3.0 stuff, not just converting to EVP_PKEY. I think a first step here is to convert to PKEY *without* including anything for 3.0, and to just have a minimal change that converts to the EVP_PKEY api with the 1.1.1. interface.

dh.c
bob-beck: I think as a first step you should separate out all the OpenSSL 3.0 stuff from this, and make a PR that *ONLY* converts to the EVP_PKEY API with the LibreSSL (1.1.1) api.

PR #446, opened 2023-09-28
Use compiler intrinsics where supported

AtariDreams:
This improves the performance of certain functions that use it.


djmdjm:
Sorry, be we generally take the upstream implementations of the PQ algorithms with as few modifications as possible. XMSS is disabled and probably going to be removed anyway once we add support for SPHINCS+

PR #447, opened 2023-09-28
Use memmove instead of memcpy

AtariDreams:
This is actually an incorrect usage of memcpy. Memcpy is when the memory is guaranteed to not overlap or be related to each other. memmove is when this may not be the case or is not the case.


aaaaaa123456789:
As far as I can tell, there's no possible overlap here? For there to be any overlap, `l` would have to be 1 or 2; 1 is precluded by the loop's condition and 2 is impossible because this is inside an `if (l & 1)` block... am I wrong?

PR #448, opened 2023-10-09
Update sshpty.c

weedcookie:
According to man pages `open` return -1 on error, which means a value bigger than or equal to zero is a valid file descriptor.


daztucker:
> According to man pages `open` return -1 on error

That's true, however your patch is incorrect.  If you look at what it's trying to do:

> * Verify that we are successfully disconnected from the controlling tty.

 /dev/tty is present when a process has controlling tty.  This means that if the open *succeeds* then the process still has a controlling tty, which means that the attempt to disconnect from it *failed*.

PR #450, opened 2023-10-16
Update openssl-devel dependency

FabioPedretti:
since openssh 9.4, openssl >= 1.1.1 is required
build with --without-openssl elsewhere
according to https://repology.org/project/openssl/versions openssl 1.1.1 is available on fedora >= 29 and rhel >= 8 successfully build tested, installed and run on rhel 6


daztucker:
Applied, thanks.

PR #451, opened 2023-10-16
Remove ancient RHL 6.x config in RPM spec.

FabioPedretti:
### Note: please have a detailed look at this, since I am actually not very familiar building packages for RH. Anyway I tested it on RHEL 6 (where this config was applied) and it looks everything is OK.

It looks like build6x options were intended for RHL 6.x (the Red Hat distro predating Fedora, not RHEL), but were then applied to RHEL.

Completely remove support for this ancient configuration.

Successfully built, installed and run on RHEL 6. This also remove a build warning about deprecation of PreReq.

### DIFF
--- a/contrib/redhat/openssh.spec
+++ b/contrib/redhat/openssh.spec
@@ -23,14 +23,6 @@
 # Use GTK2 instead of GNOME in gnome-ssh-askpass
 %global gtk2 1
 
-# Use build6x options for older RHEL builds
-# RHEL 7 not yet supported
-%if 0%{?rhel} > 6
-%global build6x 0
-%else
-%global build6x 1
-%endif
-
 %global without_openssl 0
 # build without openssl where 1.1.1 is not available
 %if 0%{?fedora} <= 28
@@ -52,14 +44,6 @@
 # RedHat <= 7.2 and Red Hat Advanced Server 2.1 are examples.
 # rpm -ba|--rebuild --define 'no_gtk2 1'
 %{?no_gtk2:%global gtk2 0}
-
-# Is this a build for RHL 6.x or earlier?
-%{?build_6x:%global build6x 1}
-
-# If this is RHL 6.x, the default configuration has sysconfdir in /usr/etc.
-%if %{build6x}
-%global _sysconfdir /etc
-%endif
 
 # Options for static OpenSSL link:
 # rpm -ba|--rebuild --define "static_openssl 1"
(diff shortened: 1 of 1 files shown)


nkadel:
RHEL 6 systems that get this RPM installed will ignore their old hostkeys in "/etc/" and generate new ones in "/etc/ssh/". This is..... begging for an unwelcome surprise as clients find mis-matched hostkeys in their $HOME/.ssh/known_hosts file and refuse to connect. I'm tempted to say "it's RHEL 6, who cares?", but banks and hospitals tend to run critical systems far, far beyond the expiration date written on the cereal box, sometimes for really good reasons to access old data with unique software. The current location is baked into ansible, chef, puppet, and systemd-init, and waagent. I would not touch any of this for another few years.

djmdjm:
Maybe @Jakuje could comment here?

Jakuje:
I would say go ahead and remove them. I do not think anybody will build openssh from source on RHEL6 (or RHL 6) from upstream repo/tarball these days with the upstream spec file. If they run RHEL6 (hope not RHL6), they have contact for somebody to maintain the packages for them and I hope that somebody will not blindly pick up spec file from upstream when there is perfectly good working spec file in dist git.

FabioPedretti:
Hi, note that the patch is not removing RHEL 6 (released in 2010), but just RHL6 (released in 1999).

The patch was intended to fix building on RHEL 6, which I still used (IIRC the patch just fixed a warning on RHEL 6).

sincerelysaucy:
If it's going to stick around, the could should be labeled for RHEL 6, not "RHL 6". No on, and I mean *no one* referred to Red Hat 6 as RHL 6. And I helped companies update about 20,000 hosts form Red Hat 5 to Red Hat 6.

FabioPedretti:
> No on, and I mean _no one_ referred to Red Hat 6 as RHL 6

The openssh.spec file is using RHL 6.x for referring to the 1999 Red Hat Linux 6:

```
# Is this a build for RHL 6.x or earlier?
...
# If this is RHL 6.x, the default configuration has sysconfdir in /usr/etc.
```

I just kept it in the commit message.

FabioPedretti:
What do you propose to change, exactly? I could eventually update the commit message...

nkadel:
In a perfect world, I would toss out contrib/redhat and add a new contrib/rhel , and update it appropriately for RHEL rather than Red Hat release use. There are other issues with the files there, such as an "sshd.init" rather than a systemd compatible "sshd.service". The modern SRPM for RHEL has 5 distinct *.service files, there is a  whole lot of material to bring in. And it also has 59 patch files applied on top of OpenSSH p1, it's kind of scary.

FabioPedretti:
I sent this patch (as well as others I sent in the past that were merged) because I am still maintaining some RHEL 6 servers, and with these patches I was/am able to successfully build/upgrade/run openssh on them (just for the security fixes, as I just did it with 9.8p1).

Actually it looks like I am the only one building rpm with the included spec file on RHEL 6 (without my previous patches building was broken).

I think your proposal is out of scope for my modest PR (which is a minor fix to the spec file), I am not planning to do any other evolution work on this (and actually I hopefully should get rid of last RHEL 6 servers soon...).

nkadel:
I suspect you're right about being the only one building for RHEL 6. Yeah, take the patch and let's start a ticket to update and relabel these for RHEL use, especially for systemd use.

daztucker:
I think the consensus is to apply this?  @djmdjm ok to apply?

djmdjm:
Done in 6b9cd09556 - thanks

PR #337, opened 2022-08-17
Flush PAM_TEXT_INFO immediately

Nigma1337:
These changes make PAM_TEXT_INFO and PAM_ERROR_MSG flush immediately.

Discussed here: https://bugzilla.mindrot.org/show_bug.cgi?id=2876 (back in 2018 lol). The old accumulating messages behavior was needed due to SSHv1.

Perhaps we should make this a config option as James suggests in the bugzilla thread.

PR #452, opened 2023-10-17
auth-pam: Immediately report instructions to clients and fix handling in ssh client

3v1n0:
As it has been discussed some long time ago in [bug #2876](https://bugzilla.mindrot.org/show_bug.cgi?id=2876), I'm adding support for immediate reporting of instructions to clients as defined by RFC4256. There was a similar attempt on #337, however this implied few more changes (see each commit for further details), in particular:
 - ~~Add an option to disable this behavior server-side~~
 - Fix `ssh` client to properly handle the instructions, parsing them properly as per the standard format
---

~~I'm not fully sure if the server-side setting is needed given that SSHv1 is not supported anymore by sshd, but probably it still makes sense to disable this behavior form server-side, even though using the device selection is possible too now.~~

### DIFF
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -990,15 +990,15 @@ sshpam_respond(void *ctx, u_int num, char **resp)
 	switch (ctxt->pam_done) {
 	case 1:
 		sshpam_authenticated = 1;
-		return (0);
+		return KbdintResultSuccess;
 	case 0:
 		break;
 	default:
-		return (-1);
+		return KbdintResultFailure;
 	}
 	if (num != 1) {
 		error("PAM: expected one response, got %u", num);
-		return (-1);
+		return KbdintResultFailure;
 	}
 	if ((buffer = sshbuf_new()) == NULL)
 		fatal("%s: sshbuf_new failed", __func__);
@@ -1015,10 +1015,10 @@ sshpam_respond(void *ctx, u_int num, char **resp)
 	}
 	if (ssh_msg_send(ctxt->pam_psock, PAM_AUTHTOK, buffer) == -1) {
 		sshbuf_free(buffer);
-		return (-1);
+		return KbdintResultFailure;
 	}
 	sshbuf_free(buffer);
-	return (1);
+	return KbdintResultAgain;
 }
 
 static void
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -132,11 +132,16 @@ typedef pid_t sp_pthread_t;
 #define pthread_join	fake_pthread_join
 #endif
 
+typedef int SshPamDone;
+#define SshPamError -1
+#define SshPamNone 0
+#define SshPamAuthenticated 1
+
 struct pam_ctxt {
 	sp_pthread_t	 pam_thread;
 	int		 pam_psock;
 	int		 pam_csock;
-	int		 pam_done;
+	SshPamDone	 pam_done;
 };
 
 static void sshpam_free_ctx(void *);
@@ -904,7 +909,7 @@ sshpam_query(void *ctx, char **name, char **info,
 				**prompts = NULL;
 				*num = 0;
 				**echo_on = 0;
-				ctxt->pam_done = -1;
+				ctxt->pam_done = SshPamError;
 				free(msg);
 				sshbuf_free(buffer);
 				return 0;
@@ -931,7 +936,7 @@ sshpam_query(void *ctx, char **name, char **info,
 				import_environments(buffer);
 				*num = 0;
 				**echo_on = 0;
-				ctxt->pam_done = 1;
+				ctxt->pam_done = SshPamAuthenticated;
 				free(msg);
 				sshbuf_free(buffer);
 				return (0);
@@ -944,7 +949,7 @@ sshpam_query(void *ctx, char **name, char **info,
 			*num = 0;
 			**echo_on = 0;
 			free(msg);
-			ctxt->pam_done = -1;
+			ctxt->pam_done = SshPamError;
 			sshbuf_free(buffer);
 			return (-1);
 		}
@@ -988,10 +993,10 @@ sshpam_respond(void *ctx, u_int num, char **resp)
 
 	debug2("PAM: %s entering, %u responses", __func__, num);
 	switch (ctxt->pam_done) {
-	case 1:
+	case SshPamAuthenticated:
 		sshpam_authenticated = 1;
 		return KbdintResultSuccess;
-	case 0:
+	case SshPamNone:
 		break;
 	default:
 		return KbdintResultFailure;
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -136,6 +136,7 @@ typedef int SshPamDone;
 #define SshPamError -1
 #define SshPamNone 0
 #define SshPamAuthenticated 1
+#define SshPamAgain 2
 
 struct pam_ctxt {
 	sp_pthread_t	 pam_thread;
@@ -868,6 +869,8 @@ sshpam_query(void *ctx, char **name, char **info,
 	**prompts = NULL;
 	plen = 0;
 	*echo_on = xmalloc(sizeof(u_int));
+	ctxt->pam_done = SshPamNone;
+
 	while (ssh_msg_recv(ctxt->pam_psock, buffer) == 0) {
 		if (++nmesg > PAM_MAX_NUM_MSG)
 			fatal_f("too many query messages");
@@ -888,15 +891,13 @@ sshpam_query(void *ctx, char **name, char **info,
 			return (0);
 		case PAM_ERROR_MSG:
 		case PAM_TEXT_INFO:
-			/* accumulate messages */
-			len = plen + mlen + 2;
-			**prompts = xreallocarray(**prompts, 1, len);
-			strlcpy(**prompts + plen, msg, len - plen);
-			plen += mlen;
-			strlcat(**prompts + plen, "\n", len - plen);
-			plen++;
-			free(msg);
-			break;
+			*num = 0;
+			free(*info);
+			*info = msg; /* Steal the message */
+			msg = NULL;
+			ctxt->pam_done = SshPamAgain;
+			sshbuf_free(buffer);
+			return (0);
 		case PAM_ACCT_EXPIRED:
 		case PAM_MAXTRIES:
 			if (type == PAM_ACCT_EXPIRED)
@@ -1001,6 +1002,8 @@ sshpam_respond(void *ctx, u_int num, char **resp)
 		return KbdintResultSuccess;
 	case SshPamNone:
 		break;
+	case SshPamAgain:
+		return KbdintResultAgain;
 	default:
 		return KbdintResultFailure;
 	}
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -446,6 +446,9 @@ sshpam_thread_conv(int n, sshpam_const struct pam_message **msg,
 			break;
 		case PAM_ERROR_MSG:
 		case PAM_TEXT_INFO:
+			debug3("PAM: Got message of type %d: %s",
+			       PAM_MSG_MEMBER(msg, i, msg_style),
+			       PAM_MSG_MEMBER(msg, i, msg));
 			if ((r = sshbuf_put_cstring(buffer,
 			    PAM_MSG_MEMBER(msg, i, msg))) != 0)
 				fatal("%s: buffer error: %s",
--- a/auth.h
+++ b/auth.h
@@ -51,6 +51,7 @@ struct sshauthopt;
 typedef struct Authctxt Authctxt;
 typedef struct Authmethod Authmethod;
 typedef struct KbdintDevice KbdintDevice;
+typedef int KbdintResult;
 
 struct Authctxt {
 	sig_atomic_t	 success;
@@ -114,6 +115,10 @@ struct Authmethod {
 	struct authmethod_cfg *cfg;
 	int	(*userauth)(struct ssh *, const char *);
 };
+
+#define KbdintResultFailure -1
+#define KbdintResultSuccess 0
+#define KbdintResultAgain 1
 
 /*
  * Keyboard interactive device:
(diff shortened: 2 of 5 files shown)


3v1n0:
FYI this has been tested also with PuTTy, and it properly supports the changes without any modification.

JoeStewart93:
Thanks for raising this @3v1n0! I'm running into issues with PAM/SSH and would love to see your fix implemented. Hopefully this can gain some traction.

auth.h
```diff
@@ -111,6 +112,12 @@ struct Authmethod {
 	int	*enabled;
 };
 
+enum KbdintResult {
+	KbdintResultFailure = -1,
+	KbdintResultSuccess,
+	KbdintResultAgain,
+};
+
```
djmdjm, line 120: This is stylistic, but we generally prefer `#define` to enum in OpenSSH. Defined symbols are easier to grep for

sshd_config.5
```diff
@@ -969,6 +969,16 @@ or
 .Cm no .
 .Cm ChallengeResponseAuthentication
 is a deprecated alias for this.
+.It Cm KbdInteractiveLegacyInstructions
```
djmdjm, line 972: What clients would break without this option? AFAIK OpenSSH clients should handle it ok (modulo formatting bugs). Are you aware of any others that have problems? I'd rather not add an option without a clearly identified need.

misc.c
```diff
@@ -3063,3 +3063,42 @@ lib_contains_symbol(const char *path, const char *s)
 	return ret;
 #endif /* HAVE_NLIST_H */
 }
+
+ssize_t
+write_msg(const char *str, int flags)
```
djmdjm, line 3068: I think you should use one of the utf8.c `*mprintf()` functions here, otherwise you are writing arbitrary characters from a (usually fine but potentially untrustworthy) server to the user's tty. The `logit()` calls that you're replacing do filter these using `strnvis()`, but as you observe this is too restrictive wrt UTF8

misc.c
```diff
@@ -3063,3 +3063,42 @@ lib_contains_symbol(const char *path, const char *s)
...
+	} else if ((output = open(_PATH_TTY, O_WRONLY)) != -1) {
+		needs_close = 1;
+	} else {
+		if (flags & WRITE_MSG_REQUIRE_TTY) {
+			errno = ENOTTY;
+			return -1;
+		}
+
+		output = output_fileno;
+	}
```
djmdjm, line 3089: I'm not convinced this section is necessary or correct. I guess it's mimicking the behaviour of `read_passphrase()`, but it isn't complete as that function will also call out to `$SSH_ASKPASS` and has some additional logic besides. 

You might be better off using the readpass.c `notify_start()` API, as it tries to do all this.

sshconnect2.c
```diff
@@ -1953,10 +1956,16 @@ input_userauth_info_req(int type, u_int32_t seq, struct ssh *ssh)
 	    (r = sshpkt_get_cstring(ssh, &inst, NULL)) != 0 ||
 	    (r = sshpkt_get_cstring(ssh, &lang, NULL)) != 0)
 		goto out;
-	if (strlen(name) > 0)
-		logit("%s", name);
-	if (strlen(inst) > 0)
-		logit("%s", inst);
+	if (strlen(name) > 0) {
+		debug("%s", name);
```
djmdjm, line 1960: recommend using `debug_f("kbd int name: %s", name);` here

sshconnect2.c
```diff
@@ -1953,10 +1956,16 @@ input_userauth_info_req(int type, u_int32_t seq, struct ssh *ssh)
...
+		if (write_msg(name, WRITE_MSG_REQUIRE_TTY) == -1)
+			goto out;
+	}
+	if (strlen(inst) > 0) {
+		debug("%s", inst);
```
djmdjm, line 1965: likewise, more context in the debug message would be helpful here too

auth-pam.c
```diff
@@ -1077,6 +1091,22 @@ KbdintDevice mm_sshpam_device = {
...
 
+KbdintDevice sshpam_device_legacy = {
+	"pam-legacy-instructions",
+	sshpam_legacy_init_ctx,
+	sshpam_query,
+	sshpam_respond,
+	sshpam_free_ctx
+};
+
+KbdintDevice mm_sshpam_device_legacy = {
```
djmdjm, line 1102: Reiterating an earlier comment: I don't think we should add this unless there is a clear motivating case of a client that is broken by these changes. Do you have an example of one?

An alternative to adding a whole other method is to add a `compat.c` pattern to detect these clients and automatically adapt.

djmdjm (commented):
Thanks for looking at the PAM code. Some comments on these changes

auth.h
3v1n0: Oh, I see... Well I thought they were also more prone to errors, but will do.

sshd_config.5
3v1n0: No, all the ones I've tested (not many, but at least 3-4 including putty and some old ones) had not issues with it.

The reason why the option has been added was this comment: https://bugzilla.mindrot.org/show_bug.cgi?id=2876#c10

And since it was one of the blocker, I thought was somewhat needed.

However, if this is not a concern, I can definitely simplify things by avoiding that.

auth-pam.c
3v1n0: Yeah, I indeed have no case for this, so we can drop it and if the problem shows for some older clients (most wouldn't work anyways since they don't support other features) then we can reiterate adding compatibility code.

Are you fine with that?

Another reason why this was added was not to change the behavior for some clients who are expecting the old way, mostly because if a client pre-this-PR will connect to a server with this PR, it may not receive valid UTF-8 chars output, and this could be an issue.

By providing an option, it's possible to use the old behavior. But that's the only case I may think we could create problems.

djmdjm:
I think implementing it without backwards compatibility is fine, we could
land this early in the release cycle and see if anyone finds breakage with
non-OpenSSH clients. If there are problems we can use a compat.[ch] hack to
turn it off for the broken clients somehow.

djmdjm:
>
> Oh, I see... Well I thought they were also more prone to errors, but will do.
>
>
There's very little type safety gained by enums, so it doesn't help as much
as we'd like :(

3v1n0:
> I think implementing it without backwards compatibility is fine, we could land this early in the release cycle and see if anyone finds breakage with non-OpenSSH clients. If there are problems we can use a compat.[ch] hack to turn it off for the broken clients somehow.

Fair... My only concerns is for those new clients (modulo this PR), that may still want the old behavior because otherwise they may get wrong utf-8 output, but if this is not a problem, we can avoid supporting `-o KbdInteractiveDevices=pam-legacy-instructions` too.

djmdjm:
On Wed, 6 Dec 2023 at 08:26, Marco Trevisan ***@***.***>
wrote:

> Fair... My only concerns is for those new clients (modulo this PR), that
> may still want the old behavior because otherwise they may get wrong utf-8
> output, but if this is not a problem, we can avoid supporting -o
> KbdInteractiveDevices=pam-legacy-instructions too.
>

Wouldn't those clients either 1) not be using notification only PAM modules
because they don't work properly now or 2) using them before a module that
prompts and therefore be seeing the messed up UTF-8 (as the messages are
accumulated) already?

3v1n0:
> On Wed, 6 Dec 2023 at 08:26, Marco Trevisan ***@***.***> wrote: Fair... My only concerns is for those new clients (modulo this PR), that may still want the old behavior because otherwise they may get wrong utf-8 output, but if this is not a problem, we can avoid supporting -o KbdInteractiveDevices=pam-legacy-instructions too.
> Wouldn't those clients either 1) not be using notification only PAM modules because they don't work properly now or 2) using them before a module that prompts and therefore be seeing the messed up UTF-8 (as the messages are accumulated) already?

1) Eh, but it's not up to the clients to decide if the server uses a pam module that may require that or not.

However, I've been conservative since there are tons of setups I may not have considered, and so providing working-already workaround for old clients seemed to me sane.

But I'll drop this. :)

misc.c
3v1n0: This is done too now, not sure if it can be done better.

3v1n0:
@djmdjm I feel that [this code](https://github.com/openssh/openssh-portable/pull/452/files#diff-ccd4282bd0b96f2181cd1e0e00448d0a5b507e8f65509aa1b5fa49ed2d04fc54L915-R929) and similar one can be now probably dropped too, right?

misc.c
3v1n0: So this can be done with those indeed, but they look like too complex for this purpose, so I used another behavior but let me know if you prefer using *mprintf funcs.

3v1n0:
@djmdjm can you have a look again on that please?

sshconnect2.c
```diff
@@ -1953,10 +1962,26 @@ input_userauth_info_req(int type, u_int32_t seq, struct ssh *ssh)
...
+		debug_f("kbd int name: %s", name);
+		if (msg == NULL)
+			goto out;
+		notifier = notify_start(0, "%s", msg);
+		notify_complete(notifier, NULL);
+		notifier = NULL;
+		free(msg);
+	}
+	if (*inst != '\0') {
+		char *msg = parse_ssh_string(inst);
```
tobhe, line 1976: same question as above

tobhe (commented):
@3v1n0 Added a few questions inline.

misc.c
```diff
@@ -3063,3 +3063,28 @@ lib_contains_symbol(const char *path, const char *s)
 	return ret;
 #endif /* HAVE_NLIST_H */
 }
+
+char *
+parse_ssh_string(const char *msg)
```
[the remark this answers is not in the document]
3v1n0: Oh... You are right... Using notifier this is not needed anymore indeed. I think I had tested this but very likely with an old compiled-ssh build.

So all this can be dropped now.

tobhe (approved):
Thanks, looks better now! The return value defines are not strictly necessary but I think they improve readability.

tobhe:
@djmdjm any chance you could have another look at this? Ubuntu plans to ship this in 24.04 and I'd very much prefer if we had an upstream approved fix instead of an out-of-tree patch.

3v1n0:
@djmdjm ping?

dmulder:
The Samba team's Himmelblau (Azure Entra ID auth) depends on this. What is the hold up?

Yusong-Gao:
> @djmdjm ping?

Hi, I have a question that is if the PR solve the backwards compatibility problem.

It could't show qrcode on terminal when I use the old ssh-client to connect the new ssh-server. This meaning  have to ask user to use ssh-client version >= x.x.

Like this:
image

3v1n0:
> > @djmdjm ping?
> 
> Hi, I have a question that is if the PR solve the backwards compatibility problem.
> 
> It could't show qrcode on terminal when I use the old ssh-client to connect the new ssh-server. In other words we have to ask user to use ssh-client version >= x.x.
> 
> Like this:
> image
> 

This fixes it for new clients, but I feel that we should expose a pam env variable with the server version to make pam modules aware of it and act consequently

Yusong-Gao:
> > > @djmdjm ping?
> > 
> > 
> > Hi, I have a question that is if the PR solve the backwards compatibility problem.
> > It could't show qrcode on terminal when I use the old ssh-client to connect the new ssh-server. In other words we have to ask user to use ssh-client version >= x.x.
> > Like this:
> > image
> 
> This fixes it for new clients, but I feel that we should expose a pam env variable with the server version to make pam modules aware of it and act consequently

The problem will be reproduced If i use ssh-clients from Ubuntu 20.04 and the ssh-servers is 24.04 latest. This commit is not backport to the old branch in Ubuntu code.

3v1n0:
> Watching this and many other pull requests, I'm starting to think that github is failing to notify contributors that their action is due, or that the notification gets buried and never becomes visible again under a pile of rubble.

Well, to ensure this is not the case the set has been sent also to the mailing list: https://marc.info/?l=openssh-unix-dev&m=173946483424943&w=2

dmulder:
I often get complaints about this issue. Both openSUSE and Ubuntu are now carrying this patch downstream because this PR has been stalled for so long. @djmdjm would you please look at this again?

3v1n0:
FYI these patches were also sent separated to the SSH ML too: https://lists.mindrot.org/pipermail/openssh-unix-dev/2025-February/thread.html

auth-pam.c
```diff
@@ -888,15 +891,13 @@ sshpam_query(void *ctx, char **name, char **info,
 			return (0);
 		case PAM_ERROR_MSG:
 		case PAM_TEXT_INFO:
-			/* accumulate messages */
```
daztucker: BTW the reason it accumulates messages instead of passing them back as individual kbdint messages is that the same code used to support both SSH1 and SSH2, and in SSH1 the TIS authentication method could only support a single challenge and a single response.  SSH1 is now gone, so this restriction no longer applies.

auth2-chall.c
```diff
@@ -170,7 +170,7 @@ kbdint_next_device(Authctxt *authctxt, KbdintAuthctxt *kbdintctxt)
 			    "keyboard-interactive", devices[i]->name))
 				continue;
 			if (strncmp(kbdintctxt->devices, devices[i]->name,
-			    len) == 0) {
+			    len) == 0 && strlen(devices[i]->name) == len) {
```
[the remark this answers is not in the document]
3v1n0: Yep, sadly it took longer than expected to land a such simple fix.

daztucker (commented):
look ok to me.  @djmdjm ok for this specific commit?

https://github.com/daztucker/openssh-portable/commit/d6cc7b72b00be125d216e25ca0f0eea85a0b7482 shows the followup cleanups.

auth2-chall.c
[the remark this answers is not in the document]
daztucker: That's true.  That said, there are only two of us doing this in mostly our spare time, which means we don't always have time to get to everything everyone wants.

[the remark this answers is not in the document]
3v1n0: Sure, don't worry... I understand it's not the easiest of the project to maintain either.

Thanks for handling this!

dmulder:
@daztucker and @djmdjm why is this _still_ stalled here?

PR #453, opened 2023-10-18
Evp pkey nokex

beldmit:
Reduced version of #445
RSA/EC in sshkey are moved to EVP API to some extent, key exchange is unttouched


regress/unittests/sshkey/common.c
```diff
@@ -83,48 +86,100 @@ load_bignum(const char *name)
...
+	RSA *rsa = NULL;
+	BIGNUM *res = NULL;
+#endif
 
 	ASSERT_PTR_NE(k, NULL);
-	ASSERT_PTR_NE(k->rsa, NULL);
-	RSA_get0_key(k->rsa, &n, NULL, NULL);
+	ASSERT_PTR_NE(k->pkey, NULL);
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+	EVP_PKEY_get_bn_param(k->pkey, OSSL_PKEY_PARAM_RSA_N, &n);
```
richsalz, line 101: Does it make sense to have one routine, `ssh_gen_bn_from_rsa` and then move the several clusters of these ifdef/else/endif lines into one place?

beldmit: I am not sure I understand how it can be done. Both named params in new API and key params in old RSA API are not pretty convenient to deal with from an universal function

ssh-rsa.c
```diff
@@ -36,53 +40,42 @@
...
+ssh_rsa_size(const struct sshkey *k)
 {
-	const BIGNUM *rsa_n;
-
-	if (key->rsa == NULL)
+	if (k->pkey == NULL)
 		return 0;
-	RSA_get0_key(key->rsa, &rsa_n, NULL, NULL);
-	return BN_num_bits(rsa_n);
+	return EVP_PKEY_bits(k->pkey);
```
Jakuje, line 50: i would keep here some sanity check the key is a RSA key.

ssh-rsa.c
```diff
@@ -36,53 +40,42 @@
...
-		return 0;
-	return 1;
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+	if (EVP_PKEY_eq(a->pkey, b->pkey) == 1)
+		return 1;
+#else
+	if (EVP_PKEY_cmp(a->pkey, b->pkey) == 1)
+		return 1;
+#endif
+	return 0;
```
Jakuje, line 78: same here as for the ecdsa keys -- I would say that a compat define and simplification would help here.

ssh-rsa.c
beldmit: Fixed, thanks!

ssh-keygen.c
```diff
@@ -372,20 +376,44 @@ static void
...
 		break;
 #endif
+#endif
+	case KEY_DSA:
+		if (!PEM_write_DSA_PUBKEY(stdout, k->dsa))
+			fatal("PEM_write_DSA_PUBKEY failed");
+		break;
 	default:
 		fatal_f("unsupported key type %s", sshkey_type(k));
 	}
```
levitte, line 419: I would replace all of this with:

``` C
	if (k->pkey != NULL) {
		if (!PEM_write_PUBKEY(stdout, k->pkey))
			fatal("PEM_write_PUBKEY failed");
	} else {
		fatal_f("unsupported key type %s", sshkey_type(k));
	}
```

(that said, without knowing the proper conditions for "unsupported key", please adjust appropriately)

levitte: Side note: `do_convert_to_pkcs8` is a surprising name for a function that outputs SubjectPublicKeyInfo in PEM form.  There's no PKCS#8 in sight...

beldmit: @levitte many thanks! It made the code much more compact.

ssh-ecdsa.c
```diff
@@ -65,30 +68,22 @@ ssh_ecdsa_size(const struct sshkey *key)
...
-		return 0;
-	if (EC_POINT_cmp(grp_a, pub_a, pub_b, NULL) != 0)
-		return 0;
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+	if (EVP_PKEY_eq(a->pkey, b->pkey) == 1)
+		return 1;
+#else
+	if (EVP_PKEY_cmp(a->pkey, b->pkey) == 1)
+		return 1;
+#endif
```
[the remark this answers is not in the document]
bob-beck: So this is a general comment about this entire change.  I made it in 455 and this is convenienty a different PR so the comment is now lost, and I'll state it again here.  The EVP_PKEY interface in 3.0 is not that different from the 1.1.1 level of API supported in libressl.  This is an example of a needless change because 3.0 STILL HAS EVP_PKEY_cmp and can do this just as it did before.  

The result here is a mess of #ifdef soup promoting the new API instead of the old one, merely because 3.0 is claiming to mark some funcitons as deprecated and the new api is preferred for new code.  this is fine, and we could even have discussions later about ensure *some* of these API's were universally supported, but that will not happen if this goes in as #ifdef soup trying to make everything the newest 3.0 way.. 

As I mentioned, I'd love to see this in a form where we simply have this coverted to the 1.1.1 EVP_PKEY api first. and at that point, look at what the minimal differences are to get this to compile on 3.0, which i expect are absolutely not this.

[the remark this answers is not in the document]
beldmit: I don't see any problem with removing #ifdef in this particular function (and the same for EC). In general I'd say we need some compatibility layer.

regress/unittests/sshbuf/test_sshbuf_getput_crypto.c
```diff
@@ -223,29 +240,74 @@ sshbuf_getput_crypto_tests(void)
...
+	    OSSL_PKEY_PARAM_GROUP_NAME, ec256_sn, strlen(ec256_sn)), 1);
+#else
 	eck = EC_KEY_new_by_curve_name(ec256_nid);
 	ASSERT_PTR_NE(eck, NULL);
 	ecp = EC_POINT_new(EC_KEY_get0_group(eck));
 	ASSERT_PTR_NE(ecp, NULL);
+#endif
 	MKBN(ec256_x, bn_x);
 	MKBN(ec256_y, bn_y);
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
```
bob-beck, line 257: So (this comment applies to all of the bag of ifdefs like this)

This define should not be used in this code like this.  It precludes libressl from ever providing a compatible API
since both packages define OPENSSL_VERSION_NUMBER - Unless when providing 3.0 API you would prefer that we again jump the shark and sit on 0x40000000L or greater to get around this sort of ifdef mathemagics. 

again, I reiterate this is doing too much at once and should just concentrate on the compatible to 1.1.1. EVP_PKEY changes, but if a define for 3.0 is to be used, please make it a define for *OPENSSL 3.0*  not numeric magic like this.

djmdjm:
This was landed in openssh-9.9

PR #454, opened 2023-11-07
man ssh_config: Clarifications around ChrootDirectory

nh2:
For upstream:

Here are some clarifications to the man page that would likely have saved thousands of hours :)


djmdjm:
Thanks, all have been committed upstream and synced back now.

nh2:
@djmdjm Thanks for doing that!

PR #456, opened 2023-12-19
Optimize aesctr_inc to avoid undefined behavior

AtariDreams:
We can ensure that an underflow for "i" never happens by only subtracting after the check for "i" being 0.


djmdjm:
I don't follow - there's no undefined behaviour here. The counter value is signed and checked before it's used to dereference anything.

djmdjm:
It doesn't look like underflow could happen here as mentioned in the previous comment. Reopen if you have evidence otherwise.

PR #457, opened 2023-12-19
Fix compilation error in ssh-pcks11-client.c

AtariDreams:
Compilation fails becaus of an undefined reference to helper_by_ec, because we forgot the preprocessor conditional that excludes that function from being called in unsupported configurations.


djmdjm:
merged as 6b8be2ccd7dd0918 -- thanks

PR #458, opened 2023-12-19
Build openssh on Mac OS not using libutil

AtariDreams:
Many functions called by libutil on macOS are deprecated, and we have a fallback, which is using the utmpx functions.


djmdjm:
What benefit does this yield?

PR #459, opened 2023-12-19
Update macOS targets for Sonoma

AtariDreams:
Sonoma is out, so we should test on there.


djmdjm:
I think @daztucker already added Sonoma, at least I see it in the github actions results.

PR #462, opened 2024-01-05
Fix ssh_config Match directive typo

rad4day:
The ssh_config.5 manpage incorrectly states that there is a Match criteria keyword called `Tag`. According to a later paragraph and the source code, this directive is actually called `tagged`


djmdjm:
Thanks, this was applied upstream and synced back as ab73f9678eb

PR #463, opened 2024-01-19
Fix proxy multiplexing mode

montag451:
The changes introduced by 7603ba71264e7fa938325c37eca993e2fa61272f broke proxy multiplexing mode. This commit makes it work again.


djmdjm:
Thanks for the fix. This has been applied upstream and synced back as 9844aa2521. It will be in the openssh-9.7 release, which is due fairly soon

PR #467, opened 2024-03-05
channels: Fix memory leak in channel_proxy_downstream

maks-mishin:
Dynamic memory, referenced by 'listen_host', is allocated by calling function 'sshbuf_get_cstring' at channels.c:3237 and lost at channels.c:3250.

Found by RASU JSC.


djmdjm:
Thanks - this has been committed upstream and synced back as c47e1c9c7911f. It will be included in the OpenSSH 9.7 release next week.

PR #468, opened 2024-03-08
[ssh-agent] Fix 100% CPU spin in signal handler

bhamiltoncx:
The ssh-agent signal handler included a call to cleanup_socket(), which wrote a log via sshlog().

This is not async-signal-safe; only async-signal-safe APIs are safe to call from inside a signal handler.

This PR fixes the issue by:

1) Creating a pair of pipe FDs
2) Adding the read end of the pipe FD to the set of FDs which
   ssh-agent polls
3) Changing the signal handler to set a global boolean value to
   true and write to the write end of the pipe FD
4) Checking the boolean at the start of each loop to see if ssh-agent
   should clean up and exit

Fixes: https://bugzilla.mindrot.org/show_bug.cgi?id=3670

### DIFF
--- a/ssh-agent.c
+++ b/ssh-agent.c
@@ -165,6 +165,18 @@ time_t parent_alive_interval = 0;
 /* pid of process for which cleanup_socket is applicable */
 pid_t cleanup_pid = 0;
 
+/*
+ * Pair of pipe() fds used to indicate the process should clean up and exit
+ * after receiving a signal.
+ */
+static int signal_cleanup_pipe[2] = { -1, -1 };
+
+/**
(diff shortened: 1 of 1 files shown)


bhamiltoncx:
Another approach from @djmdjm ensures there's a non-zero timeout passed to `poll()` so we don't have to juggle `pipe()` fds. This is more consistent with other main loops in openssh code which can be interrupted.

https://bugzilla.mindrot.org/attachment.cgi?id=3793&action=diff

bhamiltoncx:
This was superseded by a6a740a4948d10a622b505135bb485c10f21db5e.

PR #469, opened 2024-03-11
Fix DEREF_AFTER_NULL in mm_answer_keyallowed()

maks-mishin:
After having been compared to a NULL value at monitor.c:1206, pointer 'key' is dereferenced in function `auth2_record_key`.

Found by RASU JSC.


djmdjm:
key=NULL is impossible here as sshkey_froms() won't return a NULL key. I'll look at removing the unnecessary key==NULL checks instead

PR #472, opened 2024-03-12
sshconnect2: Remove redundant NULL check

maks-mishin:
NULL-pointer check for key in sshconnect2.c:1714 is not necessary because it is performed in sshconnect2.c:1708 by calling of `sshkey_is_cert()` function.

Found by RASU JCS.


kalvdans (approved):

nuschpl:
IMHO those few CPU cycles saved are not worth potential risk of security vulnerability being introduced when the code of two seperate `if` blocks will be refactored in future.

PR #473, opened 2024-03-14
Print user@host when connecting with sftp

neverpanic:
Prior to d989217110932490ba8ce92127a9a6838878928b, using sftp user@host would print "Connected to user@host", because the 'host' variable would also contain the username.

Parsing switched in the above mentioned commit for good reasons, but some users prefer to see the username if it was specified on the command line and never noticed that it did not show up when they specified a path on connection.

Restore the behavior of showing the username if one was specified and make it consistent regardless of whether a path was passed on the command line or not.

### DIFF
--- a/sftp.c
+++ b/sftp.c
@@ -2670,7 +2670,10 @@ main(int argc, char **argv)
 
 	if (!quiet) {
 		if (sftp_direct == NULL)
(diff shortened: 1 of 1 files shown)


neverpanic:
The two CI failures seem to be due to network issues and unrelated to the PR.

PR #474, opened 2024-03-14
clientloop: Remove redundant NULL checks in process_escapes()

maks-mishin:
These NULL-checks for pointer `c` are redundant because the check is already at clientloop.c:1181.

Found by RASU JSC.


maks-mishin:
Good day, I am writing to kindly remind you that I as awaiting your response. Thanks you for your attention.

PR #475, opened 2024-03-18
Fix for CVE-2020-15778

bijaytiwari:
The command "scp" can be used to pass additional linux commands to execute on the destination machine. The fix is provided in the PR.

```scp  /sourcefile remoteserver:'`touch /tmp/exploit.sh`/targetfile'```

### DIFF
--- a/session.c
+++ b/session.c
@@ -393,8 +393,11 @@ int
 do_exec_no_pty(struct ssh *ssh, Session *s, const char *command)
 {
 	pid_t pid;
+	verbose("Calsoft: do_exec_no_pty. Command - %s pid: %d", command, getpid());
+
 #ifdef USE_PIPES
 	int pin[2], pout[2], perr[2];
+	verbose("Calsoft: Use Pipes");
 
 	if (s == NULL)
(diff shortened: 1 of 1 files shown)


daztucker:
scp (the protocol, such as it is) inherently relies on interactions with the remote shell, and that's essentially unfixable.  What your example command essentially does is "ssh remoteserver '`touch ...`", and if you wanted to prevent this kind of thing you could use a restricted shell on the server.  There have been attempts at that in the past (eg scponly, rssh) with varying levels of success.  Attempts to fix this for scp-the-protocol are futile, which is why scp-the-command now defaults to using sftp-the-protocol, since it does not rely on the remote shell.

On top of that, your change does not actually fix the problem (eg you can trivially do the same thing with '$(touch /whatever)"), but it will break any perfectly valid remote command that happens to have a backtick in it.  You would need to catch every possible shell trick, and since this invokes the user's shell, you can't even assume it's a Bourne-type would need to do that for every possible shell that a user might have (what about tcsh?  fish?)

PR #476, opened 2024-03-18
getrrsetbyname: Fix null-deref with response->answer

maks-mishin:
Pointer 'rr', which is dereferenced at getrrsetbyname.c:276, is compared to a NULL value at getrrsetbyname.c:319.

Found by RASU JCS.


djmdjm:
I'll look at getting this fixed upstream, but it's amusing the analyser missed another dereference just a few lines later: https://github.com/openssh/openssh-portable/pull/476/commits/055543846e1a76b6addbd15d6c0882a08bfec223#diff-32a975c1ea4e4ce98e3d001c8be80d498926d26459768d97fe241173e73ea0a5R289

PR #477, opened 2024-03-26
sftp-server: Fix home-directory extension to match the specification

Jakuje:
The `home-directory` extension does not match the specification in PROTOCOL. Right now, it:
 * does not accept the empty string username
 * It returns always the current user home directory

The proposed change updates the extension to catch both of the corner cases.

### DIFF
--- a/sftp-server.c
+++ b/sftp-server.c
@@ -1706,14 +1706,16 @@ process_extended_home_directory(u_int32_t id)
 		fatal_fr(r, "parse");
 
 	debug3("request %u: home-directory \"%s\"", id, username);
-	if ((user_pw = getpwnam(username)) == NULL) {
+	if (username[0] == '\0') {
+		user_pw = pw;
+	} else if ((user_pw = getpwnam(username)) == NULL) {
 		send_status(id, SSH2_FX_FAILURE);
 		goto out;
 	}
 
-	verbose("home-directory \"%s\"", pw->pw_dir);
+	verbose("home-directory \"%s\"", user_pw->pw_dir);
 	attrib_clear(&s.attrib);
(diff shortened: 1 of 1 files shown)


sftp-server.c
```diff
@@ -1706,14 +1706,16 @@ process_extended_home_directory(u_int32_t id)
 		fatal_fr(r, "parse");
 
 	debug3("request %u: home-directory \"%s\"", id, username);
-	if ((user_pw = getpwnam(username)) == NULL) {
+	if (username[0] != '\0') {
```
galorithm, line 1709: I think it should be `username[0] == '\0'` as when the username is empty string, then we want the home directory for the current user `pw`

Jakuje: Indeed. Thanks for the prompt comment! Fixed.

Jakuje:
@djmdjm any chance somebody would be able to look into this?

djmdjm:
Thanks, committed upstream and synced back as 385ecb31e147dfe

rmsh1216:
> The `home-directory` extension does not match the specification in PROTOCOL. Right now, it:
> 
> * does not accept the empty string username
> * It returns always the current user home directory
> 
> The proposed change updates the extension to catch both of the corner cases.

The empty string username means that the name of the user is empty? How can I create an empty string username?

Jakuje:
> The empty string username means that the name of the user is empty? How can I create an empty string username?

No, it means the string in the SFTP message that represents username is empty. For reference, see the SFTP protocol extension implemented by OpenSSH in PROTOCOL:

https://github.com/openssh/openssh-portable/blob/master/PROTOCOL#L697

PR #478, opened 2024-03-26
Fix detection of setres*id on GNU/Hurd

sthibaul:
Like Linux, proper _SOURCE macros need to be set to get declarations of various standard functions, notably setres*id. Now that Debian is using -Werror=implicit-function-declaration this is really required. While at it, define other _SOURCE macros like on GNU/Linux, since GNU/Hurd uses the same glibc.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1347,6 +1347,12 @@ EOD
 	AC_DEFINE([BROKEN_SETVBUF], [1],
 	    [LynxOS has broken setvbuf() implementation])
 	;;
+*-*-gnu*)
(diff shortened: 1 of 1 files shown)


daztucker:
Seems reasonable.  What does ./config.guess report on that platform?

sthibaul:
It's `i686-unknown-gnu0.9`

sthibaul:
A ping on this?

daztucker:
Applied thanks.  It's on both master and V_9_8 branch so it will be in the next release.

PR #479, opened 2024-03-28
Allow ssh-askpass on Wayland by checking for $WAYLAND_DISPLAY

dkg:
Currently, no part of ssh (including the agent!) will even consider running ssh-askpass unless $DISPLAY is set.  But some systems run a graphical environment (e.g. Wayland) where some versions of ssh-askpass (e.g. ssh-askpass-gnome) will still work just fine.

So expand this baseline check to to permit invoking ssh-askpass if the sentinel wayland environment variable is present as well.

### DIFF
--- a/readpass.c
+++ b/readpass.c
@@ -127,8 +127,9 @@ read_passphrase(const char *prompt, int flags)
 	const char *askpass_hint = NULL;
 	const char *s;
 
-	if ((s = getenv("DISPLAY")) != NULL)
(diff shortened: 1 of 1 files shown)


dkg:
I think this fix will also address not only invocations of `ssh-askpass` from `ssh-agent`, but should also fix any sessions that use (for example) `ControlMaster=ask` configurations.

djmdjm:
Applied - thanks. This is 697359be

PR #482, opened 2024-04-04
Forbid others from talking to your ssh-agent

kalvdans:
Sets the permissions on the forwarded ssh-agent socket to 0600 (rw-------) to avoid other users from talking to your agent. root will still be able to talk to your agent.


djmdjm:
What does this achieve that the mode `0700` directory that already contains the socket and the `getpeereid()` permissions checks in ssh-agent.c doesn't?

kalvdans:
It doesn't add anything, sorry to not actually try connecting with another user before posting. You are completely right. Closing PR.

PR #483, opened 2024-04-05
Include forgotten `<unistd.h>` header for systemd notification

62832:
The [latest commit](https://github.com/openssh/openssh-portable/commit/08f579231cd38a1c657aaa6ddeb8ab57a1fd4f5c) adding support for direct systemd notification is currently missing the `<unistd.h>` header directive needed for `write(...)` and `close(...)`, causing a failing build when this feature is enabled such as in the `clang-12-Werror` workflow. This should fix that.


daztucker:
Applied, thanks.

PR #484, opened 2024-04-05
add support for the newline-fd readiness protocol

capezotte:
This is a partial equivalent to `sd-notify` that's already implemented in a few non-systemd supervisors — namely, [s6](https://skarnet.org/software/s6/notifywhenup.html), which originally proposed this protocol, and [dinit](https://davmac.org/projects/dinit/man-pages-html/dinit-service.5.html#ready). With creative fd wrangling, a shell script is able to recognize it.

> Daemons write a line to a file descriptor of their choice, then close that file descriptor, when they're ready to serve. This is a generic mechanism that some daemons already implement.

Even if there's no interest in merging this (as non-systemd Linux is such a rare sight that the systemd notification protocol is under `port-linux.[ch]` in this repository), I'd at least like to know if this is a sound implementation of it.


djmdjm:
How widely used are these init systems? This feature might be a bit too niche...

One problem you're likely to run into is the unconditional `closefrom()` in sshd.c:main() that will significantly constrain the possible fd numbers you are able to pass in. I'd suggest altering your patch to check whether the supplied fd number is going to get clobbered and erroring out if it is.

I'd also suggest moving the actual messaging to `platform_post_listen()`, which probably means you'll need some way of plumbing the fd number though to that file.

capezotte:
Would stdout be a suitable outlet for the newline? It appears unused when `sshd` is a daemon.

(Actually, if any option writes something to an fd other than stderr after listening, it'd be hacky but suitable for these supervisors, and something like this patch wouldn't be needed).

_______

> How widely used are these init systems? This feature might be a bit too niche...

Admittedly, it's niche even for "non-systemd Linux" standards — most are using OpenRC or sysv, which work with sshd's default behavior of forking == ready. s6 is finding some use as a [container init](https://github.com/just-containers/s6-overlay/), but that's a context where sshd doesn't make much sense.

Even if this does not get merged, thanks for the feedback!

PR #485, opened 2024-04-06
Update ssh-keyscan.c

ya0guang:
It seems like `POLLHUP` in the `else if` branch will never be triggered but causes confusion.


Mariumm11 (approved):

djmdjm:
Thanks, merged as a9d6d7d93c

PR #487, opened 2024-04-10
Add `reserved` fields to the agent protocol specification

wiktor-k:
The ssh client appends empty `reserved` fields to the [destination constraint extension][0]. This was found when working on our Rust-based [library for writing SSH agents][1] which includes tests for messages captured using the OpenSSH client.

This patch adjusts the documentation of the agent protocol to include the reserved fields.

I have not found out the reasoning for this reserved value in commit 5e950d7657 and as such I didn't expand the documentation to include further context but I think it would be a good idea to add it.

Thanks for your time! :wave: 

[0]: https://github.com/openssh/openssh-portable/blob/88351eca17dcc55189991ba60e50819b6d4193c1/authfd.c#L469
[1]: https://github.com/wiktor-k/ssh-agent-lib/blob/2605fba8497075d35daedc1f78522f3e2c17bf8d/src/proto/extension.rs#L71

### DIFF
--- a/PROTOCOL.agent
+++ b/PROTOCOL.agent
@@ -49,10 +49,13 @@ Where a constraint consists of:
 
 	string		from_username (must be empty)
 	string		from_hostname
+	string		reserved
 	keyspec[]	from_hostkeys
 	string		to_username
 	string		to_hostname
+	string		reserved
 	keyspec[]	to_hostkeys
+	string		reserved
 
 And a keyspec consists of:
 


PROTOCOL.agent
```diff
@@ -49,9 +49,11 @@ Where a constraint consists of:
 
 	string		from_username (must be empty)
 	string		from_hostname
+	string		reserved
 	keyspec[]	from_hostkeys
 	string		to_username
 	string		to_hostname
+	string		reserved
 	keyspec[]	to_hostkeys
```
baloo, line 58:
```suggestion
string reserved
```

There is also this reserved field:
https://github.com/openssh/openssh-portable/commit/5e950d765727ee0b20fc3d2cbb0c790b21ac2425#diff-62fc3a49d4e4e6c55819579611e5a53767c9dcc66aafa74fd3aa067ea0b5d31dR495

wiktor-k: Thanks for the double-check @baloo! I've squashed your change with mine and added you as `Co-authored-by` :)

djmdjm:
Thanks, this has been merged upstream and synced back as da757b022bf18c6f7d04e

PR #489, opened 2024-04-17
uint64_t -> u_int64_t

James-ZHANG:
uint64_t is part of `<stdint.h>` but that's not included in `misc.h`.


kalvdans:
Better include <stdint.h> in misc.h in my opinion.

James-ZHANG:
The main branch now already includes `stdint.h` in `misc.c`: https://github.com/openssh/openssh-portable/blob/4442bbc2fc661277a6dabfedb756a7e15ee8b8b8/misc.c#L45-L47

So I think this PR is obsolete.

kalvdans:
Yes, it seems to be fixed in https://github.com/openssh/openssh-portable/commit/18ea857770e84825a3a6238bb37f54864487b59f

PR #490, opened 2024-04-24
[loginrec.c] Fix "undeclared 'ut'" error by replacing it with 'utx'

pmiadzvedzeu:
In the `utmpx_perform_login` function there are no `ut` declarations, but there is `utx`. I believe this mistake was made during copy-paste from `utmp_perform_login` function where `ut` is used.

### DIFF
--- a/loginrec.c
+++ b/loginrec.c
@@ -1033,7 +1033,7 @@ utmpx_perform_login(struct logininfo *li)
 		return (0);
 	}
 # else
-	if (!utmpx_write_direct(li, &ut)) {
+	if (!utmpx_write_direct(li, &utx)) {
 		logit("%s: utmp_write_direct() failed", __func__);
 		return (0);
 	}


daztucker:
> Hello, @daztucker, @djmdjm All of the 32 failed checks (for now) show a CI setup error. I don't think it's my fault. Could you please check this?

Yep that's not you.  It looks like Github have managed to break either their ubuntu-latest package repo, or update the ubuntu-latest image to point to a different broken repo.  My guess is the latter, since the unbuntu-22.04 image seems unaffected.

E: Failed to fetch https://packages.microsoft.com/ubuntu/22.04/prod/dists/jammy/InRelease  Clearsigned file isn't valid, got 'NOSPLIT' (does the network require authentication?)
E: The repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease' is no longer signed.

Back to the PR: which platform(s) and/or configure options did you observe the error on:

daztucker:
> It looks like Github have managed to break either their ubuntu-latest package repo

and from more recent runs it looks like they've since fixed it, so I've rerun the failed tests.

The openssl-master test failure is on us, OpenSSL moved to 3.4 and configure did not accept that.  I've just committed a change to configure that should fix that, so if you pull in those changes you should get a clean test run.

kalvdans:
Could you change the `#ifdef`'s to normal C code: if (UTMPX_USE_LIBRARY) instead? Then both code paths will be syntax checked no matter what it is defined to.

djmdjm:
I've merged this because the previous code was clearly wrong, but Darren's question is still standing:

> Back to the PR: which platform(s) and/or configure options did you observe the error on?

PR #492, opened 2024-04-28
fix: Should check for STDERR_FILENO to cover all standard file descriptors

RealHurrison:
Should check for STDERR_FILENO to cover all standard file descriptors


djmdjm:
Thanks, I committed a slightly simplified version of this in 2e69a724051488e3fb3cd115

PR #493, opened 2024-04-29
Fix lost quit messages

rmsh1216:
We should immediately output the buffered data to stderr after jumping out of the loop, rather than after sending the disconnect message, because we may have no chance to output the data after fatal_fr().


djmdjm:
This was already fixed in 1c0d81357921f8d3bab06841df649edac515ae5b

PR #494, opened 2024-05-01
Fix ssh-agent certificate not recognized with IdentitiesOnly

AlexpFr:
- [Bugzilla – Bug 3681 - SSH Agent Certificate Not Recognized with 'IdentitiesOnly' Configured](https://bugzilla.mindrot.org/show_bug.cgi?id=3681)

The certificate present in the SSH agent is not added to the keys to be tested when 'IdentitiesOnly yes' is configured in the ~/.ssh/config file.

```bash
$ cat ~/.ssh/config
Host exemple.org
	IdentityFile ~/.ssh/id_ed25519.pub
	IdentitiesOnly Yes
```

```bash
$ ssh-add -l
256 SHA256:<FINGERPRINT> <COMMENT> (ED25519)
256 SHA256:<FINGERPRINT> <COMMENT> (ED25519-CERT)
```

```bash
$ ls ~/.ssh/
config    id_ed25519.pub
```

With the same configuration, 'ssh' defaults to looking for a certificate in ~/.ssh/id_ed25519.pub-cert::

```bash
$ ssh -vvv user@exemple.org
debug1: identity file ~/.ssh/id_ed25519.pub type 3
debug1: identity file ~/.ssh/id_ed25519.pub-cert type -1
```

I believe the expected behavior should also include searching for the certificate in the agent.

I have attempted a very simple fix.

The certificate is added from the agent with 'IdentitiesOnly Yes'. The code is redundant, but it works.

Yet another inconsistency: 'ssh-keygen' generates certificates in the form id_ed25519-cert.pub, whereas "ssh" searches for id_ed25519.pub-cert.

### DIFF
--- a/sshconnect2.c
+++ b/sshconnect2.c
@@ -1669,6 +1669,7 @@ pubkey_prepare(struct ssh *ssh, Authctxt *authctxt)
 	struct sshkey *key;
 	int disallowed, agent_fd = -1, i, r, found;
 	size_t j;
+	size_t k;
 	struct ssh_identitylist *idlist;
 	char *cp, *ident;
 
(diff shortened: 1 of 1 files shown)


djmdjm:
> I believe the expected behavior should also include searching for the certificate in the agent.

Why? Your configuration explicitly requests only a plain key be used:

```
Host exemple.org
	IdentityFile ~/.ssh/id_ed25519.pub
	IdentitiesOnly Yes
```

If you want the certificate too then why not specify it?

djmdjm:
Closing this for lack of response. It looks like user configuration is the problem here.

PR #495, opened 2024-05-09
set SSH_KEY_ID if present

maxfortun:
When authenticating via `TrustedUserCAKeys` it is useful to know the key_id in order to be able to correlate the used key to the original identity for which the key was signed. This can be leveraged to federate authentication. The process is described [here](https://github.com/maxfortun/sshd-trusted-user).

### DIFF
--- a/session.c
+++ b/session.c
@@ -1184,6 +1184,15 @@ do_setup_env(struct ssh *ssh, Session *s, const char *shell)
 		child_set_env(&env, &envsize, "SSH_ORIGINAL_COMMAND",
 		    original_command);
 
+	if(
+		s->authctxt->auth_method_key != NULL
+		&& s->authctxt->auth_method_key->cert != NULL
+		&& s->authctxt->auth_method_key->cert->key_id != NULL
+	) {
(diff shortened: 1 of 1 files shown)


djmdjm:
I think you should be able to do this using `ExposeAuthInfo` without code changes. This is a more general mechanism, but it would require you to parse keys

maxfortun:
I will check this out. Thanks.

maxfortun:
@djmdjm , you are absolutely right. I was able to extract the key id. That totally solves my problem.
Thank you again.
```
cat $SSH_USER_AUTH | cut -d' ' -f2- | ssh-keygen -L -f - | grep '^ *Key ID:'|cut -d'"' -f2
```

maxfortun:
There is only one edge-case to this. If the file system is out of space, the file won't be written and a user won't be able to login. Hypothetically, with an ENV only solution, the auth may still work.

djmdjm:
I don't think that possibility is enough to justify the additional code, sorry.

maxfortun:
Well, that possibility has happened :) Thus the code :) But I understand. Thank you for looking into this.

PR #496, opened 2024-05-09
send COLORTERM environment variable

caarlos0:
most terminals that support truecolor set the `COLORTERM=truecolor` environment variable.

currently, openssh only sends `TERM` by default. I believe we could send COLORTERM, too, so things running on the target server may properly detect colors.

That said, it has been a long time since I last did any C, and I don't know the codebase, but this is a try... happy to work more on it with some guidance if you think this is a good thing to do.

Cheers!

### DIFF
--- a/session.c
+++ b/session.c
@@ -1063,6 +1063,8 @@ do_setup_env(struct ssh *ssh, Session *s, const char *shell)
 		child_set_env(&env, &envsize, "TZ", getenv("TZ"));
 	if (s->term)
 		child_set_env(&env, &envsize, "TERM", s->term);
+	if (s->colorterm)
+		child_set_env(&env, &envsize, "COLORTERM", s->colorterm);
 	if (s->display)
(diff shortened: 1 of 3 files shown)


djmdjm:
This isn't sufficient, as TERM has special handling in the ssh session protocol (see https://datatracker.ietf.org/doc/html/rfc4254#section-6.2) that doesn't exist for COLORTERM.

Fortunately you can get COLORTERM sent without any code changes at all, using the existing environment variable passing support. Just put `SendEnv COLORTERM` in your `~/.ssh/config` and `AcceptEnv COLORTERM` in your `/etc/ssh/sshd_config` and it should work.

caarlos0:
Hey! Thanks for the response!

Yep, I'm aware of `SendEnv`... 

My idea was to send/accept `COLORTERM` by default, so the users don't have to, and all the CLIs that use it will get true color support.

Not sure if it makes sense, nor if its something SSH wants as a project, but I thought it would be nice.

PR #497, opened 2024-05-15
add malicious patch

Mic92:
nothing to see


djmdjm:
This is a good way to get your account blocked. I don't know what you are trying to achieve, but we have zero appetite for it.

PR #499, opened 2024-06-06
require public key for signing only when necessary

sisp:
Signing a file with `ssh-keygen -Y sign -n <namespace> -f <key_file> <file>` does not actually require a public key when `<key_file>` is the path to a private key file, but `ssh-keygen` is currently failing when the public key file does not exist. To avoid being unnecessarily strict, I've moved the check for the existence of a public key file. Now, an attempt at loading the public key file is made in the same place as before, but its existence is only enforced when checking the SSH agent for the key. IIUC, this is the only place where the public key is _required_.

WDYT? :slightly_smiling_face:

### DIFF
--- a/ssh-keygen.c
+++ b/ssh-keygen.c
@@ -2735,16 +2735,16 @@ sig_sign(const char *keypath, const char *sig_namespace, int require_agent,
 	if (sig_process_opts(opts, nopts, &hashalg, NULL, NULL) != 0)
 		goto done; /* error already logged */
 
-	if ((r = sshkey_load_public(keypath, &pubkey, NULL)) != 0) {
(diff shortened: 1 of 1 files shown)


djmdjm:
I don't see how this could work, because the code later does this:

```c
        if (signer == NULL) {
                /* Not using agent - try to load private key */
                if ((privkey = load_sign_key(keypath, pubkey)) == NULL)
                        goto done;
                signkey = privkey;
        } else { 
```

and `load_sign_key()` checks the pubkey against the private key unconditionally:

```c
        if (!sshkey_equal_public(pubkey, privkey)) {
                error("Public key %s doesn't match private %s",
                    keypath, privpath);
                goto done;
        }
```

which will fail if pubkey is NULL

PR #500, opened 2024-06-12
support authentication indicators in GSSAPI

abbra:
RFC 6680 defines a set of GSSAPI extensions to handle attributes associated with the GSSAPI names. MIT Kerberos and FreeIPA use name attributes to add information about pre-authentication methods used to acquire the initial Kerberos ticket. The attribute 'auth-indicators' may contain list of strings that KDC has associated with the ticket issuance process.

Use authentication indicators to authorise or deny access to SSH server. GSSAPIIndicators setting allows to specify a list of possible indicators that a Kerberos ticket presented must or must not contain. More details on the syntax are provided in sshd_config(5) man page.

Fixes: https://bugzilla.mindrot.org/show_bug.cgi?id=2696


abbra:
All failures seem to be about `penalty.sh` dependency of a t-exec which is unrelated to the PR changes, it seems.

gss-serv-krb5.c
```diff
@@ -76,6 +76,52 @@ ssh_gssapi_krb5_init(void)
...
+			chptr = strchr(options.gss_indicators[i], '=');
+			chptr[0] = '\0';
+			if (strcmp(client->indicators[j],
+			     options.gss_indicators[i]) == 0) {
+				*matched = j;
+			}
+			chptr[0] = '=';
+			denial = (chptr[1] == 'd' || chptr[1] == 'D') ? 1 : 0;
+			denials |= denial;
+			allows |= (1 - denial);
```
djmdjm, line 107: would `allows |= !denial` be more clear here?

djmdjm:
This looks pretty reasonable but I don't think any of the active developers have the GSSAPI knowledge to review the GSSAPI bits of this change. We'll have to find someone with more experience to look at it.

gss-serv-krb5.c
abbra: You are right, I replaced it by `allows |= !denial`.

abbra:
@djmdjm I am one of FreeIPA, SSSD, and Samba developers and contribute to MIT Kerberos as well.  This code is based on my indicators work for SSSD's `pam_sss_gss` which we have in production since 2021.

If you want more external eyes, I can ask my fellow Red Hat colleagues who maintain OpenSSH and wrote GSSAPI code in past too.

cryptomilk:
I'm working on libssh, MIT Kerberos and Samba. I can review the GSSAPI bits.

gss-serv.c
```diff
@@ -268,6 +268,91 @@ ssh_gssapi_parse_ename(Gssctxt *ctx, gss_buffer_t ename, gss_buffer_t name)
...
+			value.value = NULL;
+			display_value.value = NULL;
+			ctx->major = gss_get_name_attribute(&ctx->minor, gss_name,
+					&attrs->elements[i], &authenticated,
+					&complete, &value, &display_value, &more);
+			if (ctx->major != GSS_S_COMPLETE) {
+				(void) gss_release_buffer(&ctx->minor, &value);
+				(void) gss_release_buffer(&ctx->minor, &display_value);
+				(void) gss_release_buffer_set(&ctx->minor, &attrs);
+				return (ctx->major);
```
cryptomilk, line 337: I would do a `goto out;` here and add the `out:` below so you have a destructor.

abbra: Fixed, thanks.

cryptomilk:
Patch looks good to me.

gss-serv-krb5.c
```diff
@@ -76,6 +76,52 @@ ssh_gssapi_krb5_init(void)
...
+				*matched = j;
+			}
+			chptr[0] = '=';
+			denial = (chptr[1] == 'd' || chptr[1] == 'D') ? 1 : 0;
+			denials |= denial;
+			allows |= !denial;
+			if (*matched) {
+				return (1 - denial);
+			}
+		}
```
beldmit, line 111: Use `match_pattern_list` ?

beldmit (changes requested):
Overall impression is fine, one place to change

gss-serv-krb5.c
abbra: Rewrote to use `match_pattern_list`. This forced a change in the way how a rule is presented, though not too different:
```
GSSAPIIndicators [!]name1,[!]name2,...
```
Where a negation would mean a denial to the presence of the indicator.
A result is that a rule like
```
GSSAPIIndicators !hardened,pkinit,idp,passkey,otp
```
would only allow access with passwordless pre-authentication method in FreeIPA (either using smartcards, external IdP authentication, FIDO2 token or TOTP/HOTP token).

This simplifies a bit both configuration part and matching rule.

abbra: I added this as a fixup on top of the original commit to show the difference. Let me know if this is better one. I also would like to add it to `Match` handling so that one could have per-group or per-user indicator rules. Let me know if this would be useful.

gss-serv.c
```diff
@@ -268,6 +268,89 @@ ssh_gssapi_parse_ename(Gssctxt *ctx, gss_buffer_t ename, gss_buffer_t name)
...
+		/* No indicators in the ticket */
+		return (0);
+	}
+
+	count = 0;
+	for (i = 0; i < attrs->count; i++) {
+		/* skip anything but auth-indicators */
+		if (strncmp(AUTH_INDICATORS_TAG,
+			    attrs->elements[i].value,
+			    sizeof(AUTH_INDICATORS_TAG) - 1) != 0)
```
Jakuje, line 305: Is the indicator guaranteed to be NULL terminated to be safely comparable with `strncmp` when it will be shorter than the name we look for? I see in the header of this function that it should be utf8, which might imply this, but I would rather be safe to make sure we do not go behind the buffer boundary.

Additionally, are there any prefix names? The way how it is written matches just the non-null characters, which could theoretically match attributes such as `auth-indicators-something` if I see right.

Jakuje (commented):
@abbra added one comment inline with one more potential issue.

We are now working with a GSoC student to implement test coverage for libssh and GSSAPI authentication in https://gitlab.com/libssh/libssh-mirror/-/merge_requests/490
I think it would be great if something similar could be in OpenSSH upstream testsuite, which could land a test for this functionality. Let me know if you would find it useful. If so, we can try to help contribute something similar.

gss-serv.c
abbra: Yes, indicators are UTF-8 strings and encoded so in the tickets, according to RFC 8129 (sequence of utf8string ASN.1 encoding, to be precise). In both Heimdal and MIT Kerberos any ASN.1-encoded utf8string is converted into a nul-terminated string.

As for the prefixes, you might be right that we should do exact comparison and not the prefixed one. There is nothing stopping a KDC to send another name attributes that start with `auth-indicators-` prefix although nothing is done like that today.

abbra:
I updated the code to follow recent discussions. Let me know whether this is all fine for merge.

I have recorded a demo here: https://www.youtube.com/watch?v=OauHdZYKGKk. This demo shows the use of `GSSAPIIndicators` option in a `Match` block to constrain two users (`unenforced-user` and `enforced-user`) to authenticate with GSSAPI when their initial Kerberos tickets were obtained in a particular way. In this demo we just deny any ticket that was not obtained with OTP or a smartcard-based pre-authentication mechanism. Since FreeIPA associated a `hardened` indicator with SPAKE pre-authentication method (used for password-based auth), we also reject `hardened` indicator.

In the demo, first we attempt to access as `unenforced-user`, with a ticket that contains `hardened` indicator. Then we access as `enforced-user`, with a ticket that contains `otp` indicator. In the first case we fail and system logs show that clearly. In the second case we succeed. Again, logs demonstrate that a particular indicator was inspected and accepted.

abbra:
I have been on vacation for past few weeks so may be I missed something...
@djmdjm Is there anything else preventing a further review and inclusion of the GSSAPI authentication indicators work?

beldmit:
This code looks good to me after the changes I've requested are done and we are interested in having it upstream.
@djmdjm could you please look at it if time permits?

beldmit:
@abbra is it just a rebase or also changes?

abbra:
Just a rebase to clear a small conflict with the renamed option.

abbra:
@bob-beck suggested to add a CI test. I'll look at this.

gss-serv-krb5.c
```diff
@@ -100,11 +127,38 @@ ssh_gssapi_krb5_userok(ssh_gssapi_client *client, char *name)
 	}
 	if (krb5_kuserok(krb_context, princ, name)) {
```
djmdjm: IMO this function would be clearer if it shortcircuits when preconditions fail, i.e.

```C
	if (!krb5_kuserok(krb_context, princ, name))
		goto out;
	if (options.gss_indicators == NULL)  {
		retval = 1;
		goto out;
	}
	if (!client->indicators) {
		logit("GSSAPI authentication indicators enforced "
			"but not matched. krb5 principal %s denied",
		(char *)client->displayname.value);
		goto out;
	}
...
 out:
	if (retval == 1)  {
		logit("Authorized to %s, krb5 principal %s (%s)",
		    name, (char *)client->displayname.value, errmsg);
	}
	krb5_free_principal(krb_context, princ);
	return retval;
```

gss-serv.c
```diff
@@ -269,6 +269,92 @@ ssh_gssapi_parse_ename(Gssctxt *ctx, gss_buffer_t ename, gss_buffer_t name)
...
+	if (attrs == GSS_C_NO_BUFFER_SET) {
+		/* No indicators in the ticket */
+		return (0);
+	}
+
+	count = 0;
+	for (i = 0; i < attrs->count; i++) {
+		/* skip anything but auth-indicators */
+		if (((sizeof(AUTH_INDICATORS_TAG) - 1) != attrs->elements[i].length) ||
+		    strncmp(AUTH_INDICATORS_TAG,
```
djmdjm: memcmp here?

gss-serv-krb5.c
abbra: Thanks, it makes sense. I rewrote the logic to reduce if-conditions.

gss-serv-krb5.c
```diff
@@ -85,26 +112,62 @@ static int
...
-		retval = 1;
-		logit("Authorized to %s, krb5 principal %s (krb5_kuserok)",
-		    name, (char *)client->displayname.value);
-	} else
+
+	retval = krb5_kuserok(krb_context, princ, name) == TRUE;
+	if (retval == 0)
+		goto out;
+
+	if (options.gss_indicators == NULL)
```
djmdjm: Similarly, AFAIK the function should return 1 in this case too.

gss-serv.c
```diff
@@ -269,6 +269,94 @@ ssh_gssapi_parse_ename(Gssctxt *ctx, gss_buffer_t ename, gss_buffer_t name)
...
+		return (ctx->major);
+	}
+
+	if (attrs == GSS_C_NO_BUFFER_SET) {
+		/* No indicators in the ticket */
+		return (0);
+	}
+
+	count = 0;
+	for (i = 0; i < attrs->count; i++) {
```
djmdjm: Can this loop be merged with the one below? You can update `count` and use `xrecallocarray()` to grow the array as you go. As it stands, if the behaviour of the 2nd loop differs from the first, then it could write out of `client->indicators` bounds.

gss-serv-krb5.c
```diff
@@ -85,26 +112,62 @@ static int
...
 		krb5_free_error_message(krb_context, errmsg);
 		return 0;
 	}
-	if (krb5_kuserok(krb_context, princ, name)) {
-		retval = 1;
-		logit("Authorized to %s, krb5 principal %s (krb5_kuserok)",
-		    name, (char *)client->displayname.value);
-	} else
+
+	retval = krb5_kuserok(krb_context, princ, name) == TRUE;
```
djmdjm: the logic is a little hard to follow here.

Maybe it would be clearer (and safer) to have a dedicated `success` variable that is explicitly set only when the function has succeeded. I.e. after the `options.gss_indicators == NULL` test below and after `ssh_gssapi_check_indicators()`  returns a positive match.

djmdjm (changes requested):

abbra:
@djmdjm thanks for the comments. I addressed them.

We still need to add a test but it is a wider effort as you pointed out in https://lists.mindrot.org/pipermail/openssh-unix-dev/2025-December/042303.html. We have a lot of tests downstream in Fedora/RHEL and plan to adopt them to upstream CI.

PR #501, opened 2024-06-21
Sx is for internal references only

domenpk:
Therefore, the links will not work. Replace with .Cm which seems more suitable.


djmdjm:
Merged as d76424bf27; thanks

PR #503, opened 2024-06-29
ssh-keygen: print keyfile when reading passphrase

qwerty287:
Hello!

When ssh-keygen needs a passphrase to decrypt a private key, it just prints the message `Enter passphrase:`.

However, applications like KSSHAskPass are using this prompt to determine which key is currently being used so the correct passphrase can be requested from the password stores.

This changes the message to the one used for most other requests, `Enter passphrase for key '…':`.

I never developed with C before, so if there's something that needs a change, please just comment!

Thanks!


ChrisTG742:
I think this output should only be optional, activated by a commandline-switch.
Adding additional screen-output by default could break any existing scripts that use ssh-keygen, for example automation-scripts that are used in UI-wizards when installing a fresh Linux distribution.
So I'd suggest to have that output disabled by default and add another commandline-option to enable it.

ChrisTG742:
> @ChrisTG742 What do you think about [my suggestion](https://github.com/openssh/openssh-portable/pull/503#issuecomment-2200863602)?

Sounds reasonable to me. However, it's not up to me to decide as I'm not the maintainer of this repo.
I just indicate that it could cause trouble if you change the standard-output.

djmdjm:
IMO the best solution here is to make `read_passphrase()` accept a printf(3)-style format string. Then it will be trivial to add the filename wherever it is used. I'll take a look at this, but I'm not sure it will make the 9.9 release.

PR #504, opened 2024-07-03
fix utmpx ifdef

christoph-zededa:
02e16ad95fb1f56ab004b01a10aab89f7103c55d did a copy-paste for utmpx, but forgot to change the ifdef appropriately

This fix also has been done in the alpine repo: https://github.com/alpinelinux/aports/blob/master/main/openssh/fix-utmp.patch#L31

### DIFF
--- a/loginrec.c
+++ b/loginrec.c
@@ -723,7 +723,7 @@ set_utmpx_time(struct logininfo *li, struct utmpx *utx)
 void
 construct_utmpx(struct logininfo *li, struct utmpx *utx)
 {
-# ifdef HAVE_ADDR_V6_IN_UTMP
+# ifdef HAVE_ADDR_V6_IN_UTMPX
 	struct sockaddr_in6 *sa6;
 #  endif
 	memset(utx, '\0', sizeof(*utx));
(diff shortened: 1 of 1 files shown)


dkwo:
with 9.9p1, i also get https://github.com/void-linux/void-packages/pull/52297
```
usr/bin/ld: loginrec.o: in function `record_failed_login':
loginrec.c:(.text+0x116a): undefined reference to `construct_utmp'
```
is this part of the same issue?

djmdjm:
> with 9.9p1, i also get [void-linux/void-packages#52297](https://github.com/void-linux/void-packages/pull/52297)
> 
> ```
> usr/bin/ld: loginrec.o: in function `record_failed_login':
> loginrec.c:(.text+0x116a): undefined reference to `construct_utmp'
> ```
> 
> is this part of the same issue?

Almost certainly not; that was probably caused by 671c440786a5a66216

dkwo:
@djmdjm That was fixed in Void by this patch https://github.com/void-linux/void-packages/blob/master/srcpkgs/openssh/patches/musl-btmp.patch
We also carry a bunch of other patches, maybe this one https://github.com/void-linux/void-packages/blob/master/srcpkgs/openssh/patches/ssh-musl-gssapi.patch could be upstreamed too?

djmdjm:
thanks, I just applied both of those

djmdjm:
btw, I've pushed all these commits to the stable V_9_9 branch

PR #506, opened 2024-07-06
Update scp.c

Builder-Ed:
The former code might not be efficient enough because the new variable seemed to be of no use.


djmdjm:
No, the comment there explains why this is needed. E.g. https://github.com/openssh/openssh-portable/blob/master/scp.c#L1407

PR #510, opened 2024-07-16
regress/test-exec: use the absolute path in the SSH env

quaresmajose:
The SSHAGENT_BIN was changed in [1] to SSH_BIN but the last one don't use the absolute path and consequently the function increase_datafile_size can loops forever if the binary not found.

[1] https://github.com/openssh/openssh-portable/commit/a68f80f2511f0e0c5cef737a8284cc2dfabad818

### DIFF
--- a/regress/test-exec.sh
+++ b/regress/test-exec.sh
@@ -175,6 +175,11 @@ if [ "x$TEST_SSH_OPENSSL" != "x" ]; then
 fi
 
 # Path to sshd must be absolute for rexec
+case "$SSH" in
+/*) ;;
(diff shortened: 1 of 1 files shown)


quaresmajose:
The CI fails on the configure step of some `windows-20xx` but the patch don't touch any of this so it could only be a false positive.
```
C/C++ CI / windows-2019 default (pull_request) Failing after 3m
C/C++ CI / windows-2022 default (pull_request) Failing after 3m
```

daztucker:
Applied upstream, thanks.  (It'll get pulled into this repository next time we sync.)

PR #511, opened 2024-07-17
Add SM3 secure hash algorithm

uudiin:
Add OSCCA SM3 secure hash algorithm (OSCCA GM/T 0004-2012 SM3).


FreezB11:
@uudiin umm can i do something like, i want to help so can you let me know

PR #514, opened 2024-08-02
Fix the issue where without_openssl is always set to 1.

wuwx:
In Fedora systems, %{?rhel} is empty. In RHEL systems, %{?fedora} is empty. Therefore, the original code always sets without_openssl to 1.

### DIFF
--- a/contrib/redhat/openssh.spec
+++ b/contrib/redhat/openssh.spec
@@ -31,13 +31,13 @@
 %global build6x 1
 %endif
 
-%global without_openssl 0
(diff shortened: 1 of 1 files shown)


djmdjm:
I think it would be better to leave the default of `without_openssl` at 0 and only override it for known-bad versions

wuwx:
i've set without_openssl back to 0, and check fedora and rhel defined now

djmdjm:
That's committed and will be in OpenSSH 9.9, due in a few weeks

PR #519, opened 2024-09-25
fix error support clang

sgzGary:
I meet problems like this when I use clang:
[  172s] ssh-ecdsa.c:267:6: error: call to undeclared function 'is_ecdsa_pkcs11'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[  172s]   267 |         if (is_ecdsa_pkcs11(key->ecdsa)) {
[  172s]       |             ^
[  172s] 1 warning and 1 error generated.
so i need to add head support

### DIFF
--- a/ssh-ecdsa.c
+++ b/ssh-ecdsa.c
@@ -42,7 +42,7 @@
 #include "digest.h"
 #define SSHKEY_INTERNAL
 #include "sshkey.h"
-
+#include "ssh-pkcs11.h"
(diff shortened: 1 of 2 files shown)


djmdjm:
I don't know what you are doing but there is no function `is_ecdsa_pkcs11()` in OpenSSH. Maybe you are using some third-party patch? If so, you need to send this to the patch authors.

PR #520, opened 2024-09-28
Shut down forked PAM handle

Xyene:
Shutting down the forked PAM handle must be done as per the Linux-PAM documentation, specifically after setuid.

https://fossies.org/linux/Linux-PAM-docs/doc/adg/html/adg-interface-by-app-expected.html#adg-pam_end

This is not a new requirement, but until recently (~2021) there was no consequence to not doing so.

`pam_cap` now requires the handle to be shut down correctly in order to configure ambient capabilities for the session. Importantly, these must be configured after setuid, as setuid clears the ambient capability vector.

See this comment (and thread) for more details: https://github.com/shadow-maint/shadow/pull/408#issuecomment-919673098


auth-pam.c
```diff
@@ -661,6 +661,18 @@ sshpam_store_conv(int n, sshpam_const struct pam_message **msg,
...
 
+void
+sshpam_cleanup_in_child(void)
+{
+	if (sshpam_handle == NULL)
+		return;
+
+#ifdef PAM_DATA_SILENT
+	/* macOS PAM doesn't support PAM_DATA_SILENT.  */
+	pam_end(sshpam_handle, PAM_SUCCESS | PAM_DATA_SILENT);
```
3v1n0, line 672:
```suggestion
	pam_end(sshpam_handle, sshpam_err | PAM_DATA_SILENT);
```

Shouldn't be still be ended even outside pam-linux?

Xyene: I'm not sure it's safe to do so without `PAM_DATA_SILENT`, otherwise `pam_end` may result in modules trying to perform cleanup work meant for the parent (deleting directories, logging, etc.)

AndrewGMorgan: If the forked process is going to be run, I would expect `sshpam_err` to always be `PAM_SUCCESS`. When one of the `pam_authenticate()` etc calls fail to return `PAM_SUCCESS`, I would have expected this fork to not have happened.

+1 to @Xyene - this does not change the need for the parent to call `pam_end(sshpam_err)` and that call should not have `PAM_DATA_SILENT` or'd in.

Xyene:
I moved the cleanup to after `do_setup_env`, since its original position before broke `pam_env.so` setting environment variables.

PR #521, opened 2024-09-30
auth-pam: Check the user didn't change during PAM transaction

3v1n0:
PAM modules can change the user during their execution, in such case ssh would still use the user that has been provided giving potentially access to another user with the credentials of another one.

So prevent this to happen, by ensuring that the final PAM user is matching the one that initiated the transaction.

Similar to https://github.com/util-linux/util-linux/pull/3206 and https://github.com/sudo-project/sudo/pull/412

### DIFF
--- a/auth-pam.c
+++ b/auth-pam.c
@@ -467,6 +467,28 @@ sshpam_thread_conv(int n, sshpam_const struct pam_message **msg,
 	return (PAM_CONV_ERR);
 }
 
+static int
+check_pam_user(Authctxt *authctxt)
+{
+	const char *pam_user;
+
+	if (authctxt == NULL || authctxt->pw == NULL || authctxt->pw->pw_name == NULL)
+		fatal("%s: PAM authctxt user not initialized", __func__);
+
+	sshpam_err = pam_get_item(sshpam_handle,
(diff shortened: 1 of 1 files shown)


djmdjm:
Thanks - I just pushed a slightly-modified version of this as 140bae1df2b7246bb43439d039bf994159973585

The main change is that, in `sshpam_init()` it disallows changing the username at all. Previously it would allow username changes if `sshpam_init()` was called again. I don't see any point in permitting that.

km1967:
This also affects macOS, which ships with 

$ sshd -V
OpenSSH_10.2p1, LibreSSL 3.3.6

The case is when there is (intentional) mismatch  between RealName and RecordName

$ dscl . read /Users/administrator2021 RealName RecordName NFSHomeDirectory
NFSHomeDirectory: /Users/administrator2021
RealName: admin
RecordName: administrator2021

previously

ssh admin2... worked, as well as ssh administrator2021@...

But since some update of MacOS / openssh only ssh administrator2021@... works.

Now ssh admin@... gets this error:

debug1: PAM user "admin" does not match expected "administrator2021"
PAM user mismatch

Note 

id admin
and 
id administrator2021

give identical output

$ sdiff -s <(id admin) <(id administrator2021)
$

PR #522, opened 2024-10-14
fix duplicate PAM msgs, missing loginmsg reset

hellermf:
Without this change in **mm_answer_pam_account** all messages added in **auth-pam.c sshpam_query(...)** `case PAM_SUCCESS` end up sent here, then are still sitting in the **loginmsg** buffer and printed a second time in **session.c do_login(...)**

I initially encountered this problem using `pam_duo.so` (described by some other user in the link below) however it is not specific to that PAM module, I also get messages printed twice when pam_echo.so, pam_debug.so, and others are included in the _auth_ PAM chain used by sshd.

https://community.cisco.com/t5/protecting-applications/duplicate-success-logging-you-in-after-a-successful-login/td-p/4879911

> Autopushing login request to phone...
> Success. Logging you in...
> Autopushing login request to phone...
> Success. Logging you in...

Without this code change, given the logic inside **do_login(...)** it is possible to work-around the problem on per-user basis with `touch ~/.hushlogin` but that also has the side-effect of disabling printing of the MOTD by sshd if that is configured ...and it is per-user not system-wide.

I gave some thought to whether there is any circumstance where the behavior of the original code is desired but thus far I cannot imagine one, although I am not deeply familiar with OpenSSH codebase.


hellermf:
Hi!
I am just wondering if there is anything I can do to help move this PR forward. For example, would it help if I filed a corresponding bug report in Bugzilla? Is there any additional information I should provide?

daztucker:
applied, thanks.  I also cherry picked it into the 10.2 release branch, so it'll be in the next release either way.

PR #523, opened 2024-10-15
Create openssh-portable

mustafacco7:


daztucker:
what is a devcontainer.json and why would we want one?

PR #525, opened 2024-10-21
Updated gitignore to ignore sshd-session and sshd-auth targets

amballip:
sshd-session and sshd-auth targets were missing in .gitignore file, showing up as untracked files in git status command


daztucker:
Applied, thanks.

PR #528, opened 2024-11-02
Inherit DESTDIR from the environment.

sortie:
autoconf packages conventionally inherit the DESTDIR variable from the environment.


djmdjm:
Merged, thanks

PR #529, opened 2024-11-02
Fix configure implicit declaration and format warnings.

sortie:


configure.ac
```diff
@@ -4344,7 +4345,11 @@ int main(void)
 	long long num = 0x7fffffffffffffffll;
 #endif
 	strcpy(expected_out, "9223372036854775807");
+#if (SIZEOF_LONG_INT == 8)
+	snprintf(buf, mazsize, "%ld", num);
+#else
 	snprintf(buf, mazsize, "%lld", num);
```
djmdjm, line 4351: I don't see how this is correct: num is a long long, so the %lld format string should be fine

sortie: num is a long sometimes per the ifdef its defined in. Have a look at the context of this patch :)

daztucker:
Applied, thanks.

PR #531, opened 2024-11-02
Remove unused <netinet/in_systm.h> and <resolv.h> inclusions.

sortie:
netinet/in_systm.h defines n_short, n_long, and n_time which are not used anywhere in openssh.

resolv.h is not used in the affected files.

### DIFF
--- a/defines.h
+++ b/defines.h
@@ -55,7 +55,6 @@ enum
 /*
  * Definitions for IP type of service (ip_tos)
  */
-#include <netinet/in_systm.h>
 #include <netinet/ip.h>
 #ifndef IPTOS_LOWDELAY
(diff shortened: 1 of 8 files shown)


djmdjm:
resolv.h is needed for base64 functions on some platforms

```
hostfile.c:97:8: error: call to undeclared function '__b64_pton'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
        ret = __b64_pton(b64salt, salt, salt_len);
```

daztucker:
I've partially applied this removing most of the in_systm.h includes, but dropping the resolv.h parts based on Damien's comment.  Thanks.

PR #532, opened 2024-11-04
Don't use login as a verb

ThinLinc-Zeijlon:
I found these while passing by!

Motivation:
- "We can handle one or more **login** types".
- "But I **log in** to my server".


ThinLinc-Zeijlon:
It doesn't look like the failing test has anything to do with the changes i made. Not sure how to re-run `C/C++ CI / windows-2019 cygwin-release (pull_request)`

daztucker:
Looks like there was an infrastructure failure setting up the VM for the test.  Requested a rerun of the failed test.

daztucker:
Applied thanks.  (The session.c one went upstream, it'll get merged back into this tree next time we sync it).

PR #534, opened 2024-11-07
Fix regular users expired password reset issue

chmmeng:
When PASSWD_NEEDS_USERNAME is enabled and a password expires, a regular user like test will see the error: passwd: Only root can specify a user name upon login. Allowing regular users to change their own password is a reasonable requirement, so when PASSWD_NEEDS_USERNAME is enabled, a condition needs to be added to address this.

### DIFF
--- a/session.c
+++ b/session.c
@@ -1438,8 +1438,11 @@ do_pwchange(Session *s)
 		setexeccon(NULL);
 #endif
 #ifdef PASSWD_NEEDS_USERNAME
-		execl(_PATH_PASSWD_PROG, "passwd", s->pw->pw_name,
-		    (char *)NULL);
+               if (getuid() != 0) {
+                       execl(_PATH_PASSWD_PROG, "passwd", (char *)NULL);
+               } else {
+                       execl(_PATH_PASSWD_PROG, "passwd", s->pw->pw_name, (char *)NULL);
+               }		
 #else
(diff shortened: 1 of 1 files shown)


djmdjm:
On which platform did you experience this? AFAIK the existing code is correct for some of the three platforms where PASSWD_NEEDS_USERNAME is enabled (Solaris, UnixWare, OpenServer)

chmmeng:
I encountered this issue on a CentOS-like platform. Unlike Solaris, UnixWare, and OpenServer, CentOS doesn’t require PASSWD_NEEDS_USERNAME for standard users, as only root can specify a username with passwd. This difference seems to cause unexpected behavior in our environment.

djmdjm:
I'm more confused now - no Linux variant sets PASSWD_NEEDS_USERNAME in configure, are you setting it manually? If so, why?

chmmeng:
yes,we set it manually to resolve a bug.
there are two users: root and sysop uid=0,and the sysop account has expired. When logging in as sysop, it prompts to change the root password instead of the sysop password. Here are the steps 

useradd sysop 
Edit /etc/passwd, change the UID and GID of the sysop user to 0, save and exit.
passwd sysop to set a password for the sysop user.
Log in via SSH using the sysop account.
Exit the SSH session.
As root, execute passwd -e sysop to expire the sysop password.
SSH into the server using the sysop account. It prompts to reset the password, but the username mentioned is root, not sysop.

PR #535, opened 2024-11-07
Add notify_hostkeys configuration

chmmeng:
Submit the notify_hostkeys Configuration Switch

Performance Optimization:

In high-concurrency environments, sending multiple host keys during each connection can impact connection establishment times. This configuration switch allows administrators to disable the feature when performance is critical.

Flexibility and Customization:

Different environments may have varying needs. For example, in trusted networks or testing environments, frequent host key notifications may not be necessary. This switch provides the flexibility to adjust the behavior accordingly.

Reduced Network Overhead:

In bandwidth-constrained networks, transmitting multiple host keys can introduce unnecessary load. Disabling this feature helps optimize data transmission.
Balance Between Security and Usability:

Some administrators prefer a simpler connection experience without frequent key notifications. This switch allows them to strike a balance between security and performance based on their specific needs.

### DIFF
--- a/servconf.c
+++ b/servconf.c
@@ -216,6 +216,7 @@ initialize_server_options(ServerOptions *options)
 	options->sshd_session_path = NULL;
 	options->sshd_auth_path = NULL;
 	options->refuse_connection = -1;
+	options->notify_hostkeys = -1;
 }
 
 /* Returns 1 if a string option is unset or set to "none" or 0 otherwise. */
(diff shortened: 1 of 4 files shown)


djmdjm:
Could you share some information on how the hostkeys notifications are problematic for you? The notification messages in the common case are short (just one or more public keys) and consume effectively no CPU overhead for the common case where the client already has the full set of hostkeys. These extensions are used even on extremely busy servers, such as Github's.

BTW an easy way to effectively disable the notifications is to have the server offer only one hostkey at a time.

chmmeng:
Thank you for the suggestion! We’re operating in a cloud environment where our systems, especially for banking clients, require frequent SSH connections due to high demands for automated operations and dynamic resource management. In this cloud setting, even minimal delays in SSH connection times accumulate quickly, affecting overall efficiency.

Specifically, in cloud-based banking environments, we frequently establish and close connections, which can make the hostkey notifications a performance bottleneck. The added notifications during each connection setup, although lightweight, lead to additional network overhead, particularly in high-frequency operations.

Given these requirements, we’re exploring ways to streamline SSH connections by disabling or limiting these notifications, and we’ll consider your recommendation to try offering a single hostkey to observe its effect on performance.

PR #536, opened 2024-11-09
readpass: add fallback to tty if default askpass unavailable

YHNdnzj:
Currently, when read_passphrase() is called with RP_ALLOW_STDIN + !isatty(STDIN_FILENO) or $SSH_ASKPASS_REQUIRE=prefer, and running in GUI environment (w/ $DISPLAY or $WAYLAND_DISPLAY), askpass is effectively enforced. This behavior is not ideal though when no askpass program is installed, as it results in hard failure.

Instead, check the existence of the default askpass path early, and if unavailable fall back to tty in the 2 cases mentioned above.

### DIFF
--- a/readpass.c
+++ b/readpass.c
@@ -123,7 +123,8 @@ char *
 read_passphrase(const char *prompt, int flags)
 {
 	char cr = '\r', *askpass = NULL, *ret, buf[1024];
-	int rppflags, ttyfd, use_askpass = 0, allow_askpass = 0;
+	int rppflags, ttyfd, use_askpass = 0, allow_askpass = 0,
+        askpass_fallback_tty = 0;
 	const char *askpass_hint = NULL;
 	const char *s;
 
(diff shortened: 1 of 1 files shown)


djmdjm:
I think doing it this way might be a little cleaner. Does it solve your problem?

https://github.com/djmdjm/openssh-wip/pull/46/commits/0ad874541fc7f8cafa6f65a7c588b50218958cdf

YHNdnzj:
> I think doing it this way might be a little cleaner. Does it solve your problem?
> 
> [djmdjm/openssh-wip@0ad8745](https://github.com/djmdjm/openssh-wip/commit/0ad874541fc7f8cafa6f65a7c588b50218958cdf)

A cleaner approach is definitely appreciated. However, it appears to me that your patch would break compat if `$SSH_ASKPASS` points to executable in `$PATH`, rather than being absolute. That's why I resorted to only checking the existence for default askpass program, and rely on `exec*p()` returning error otherwise.

PR #541, opened 2024-11-22
mdoc2man: balance nested square brackets

raboof:
I noticed the square brackets in `destination [command [argument...]` in the synopsis for the `ssh.1` manpage were not balanced, this balances them.

(I also proposed this as a patch on the openssh-unix-dev mailing list, Oct 20th, message id 811cbc01-1c58-4ed8-82fb-bcb34364016f@app.fastmail.com, but not sure if that's still on someone's radar or got lost)


daztucker:
Applied, thanks.

PR #542, opened 2024-11-29
Fix security keys without attestation

michael-dev:
Using libfido2 with windows://hello results in security key returning no attestation data. This currently fails due to fido_cred_verify_self failing.

According to https://github.com/Yubico/libfido2/issues/840 this is not a bug in libfido2, but openssh instead has to skip the verify call if no attestation is given.

This fixes the issue by skipping attestation verification during key generation if there is no attestation.

Fixes https://github.com/PowerShell/Win32-OpenSSH/issues/2040 and https://github.com/PowerShell/Win32-OpenSSH/issues/2279

### DIFF
--- a/sk-usbhid.c
+++ b/sk-usbhid.c
@@ -961,13 +961,15 @@ sk_enroll(uint32_t alg, const uint8_t *challenge, size_t challenge_len,
 			    fido_strerr(r));
 			goto out;
 		}
-	} else {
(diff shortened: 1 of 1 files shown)


djmdjm:
This looks reasonable to me, but @martelletto knows way more about this than I do

martelletto:
LGTM

djmdjm:
Merged as d3a7ff7cecbc and will be in openssh-10.0, due in a couple of months

PR #543, opened 2024-12-02
Fix compilation with DEBUG_SK enabled

shivak:
In `ssh_ecdsa_sk_verify`, the `datalen` variable was renamed to `dlen` -- but not in this debugging block.


daztucker:
applied, thanks.

PR #544, opened 2024-12-05
regress/key-options.sh: do not run test with 2038 expiry date on 64 bit time_t systems

kanavin:
This allows testing Y2038 with system time set to after that (i.e. 2040), so that actual Y2038 issues can be exposed, and not masked by key expiry errors.


kanavin:
This is a follow-up to #425

daztucker:
Applied, thanks.

PR #546, opened 2024-12-09
Fix out-of-tree build failure for standalone sk-libfido2

thelastlin:
The compiler can't find `sk-usbhid.c` when building out-of-tree.

This fixes the issue by provide the explicit path of `sk-usbhid.c` in compile command.


daztucker:
applied, thanks.

PR #550, opened 2025-01-05
Fix handling of Haiku remote hosts in ssh-copy-id.

infrastation:
As described. This is an improved revision of haikuports/haikuports#11587.


djmdjm:
Thanks, but we just include the upstream version of ssh-copy-id from https://git.hands.com/ssh-copy-id - you're best off reporting this to phil@hands.com or via the openssh-unix-dev mailing list.

infrastation:
Thank you, posted the next revision to the mailing list.

PR #552, opened 2025-01-16
Fix debug logging of user_specific_delay

fjl5:
In debug log, user_specific_delay always erroneously reports
```
debug3: user_specific_delay: user specific delay 0.000ms [preauth]
```
Maybe the debug log should also add `MIN_FAIL_DELAY_SECONDS`, to really report the returned value?


daztucker:
Applied upstream, thanks.  We'll pull the change into this repo during the next sync.

PR #554, opened 2025-01-20
Make 3rd party unit tests log "PERKELE" instead of "FUCKOFF"

martinpaljak:
I'm using this sample file as input to `known_hosts` parsing unit tests. I don't mind a bit of swearing but this Finnish word is a bit less direct than "FUCKOFF".

Reference: https://www.youtube.com/watch?v=g56Izmup9zw


djmdjm:
I don't understand - nothing in OpenSSH logs this string

martinpaljak:
@djmdjm in OpenSSH maybe not, but try this from the source tree root: `tail -1 ./regress/unittests/hostkeys/testdata/known_hosts | awk '{print $3}' | base64 -d`

pinacoelho:
Why should the Finns have all the fun ?

If the issue is the swearing, then change it to "GoAway", it's both easy to understand and not swearing.

PR #555, opened 2025-02-06
Debugging AWS-LC CI 

smittals2:
Debugging changes from bz3784. The AWS-LC CI is failing on the make tests step. We do not see this in our CI when building AWS-LC with OpenSSH - investigating the root cause. 

Updates: 
- Updating the ubuntu machine to 22.04 fixed the issue
- We are tripping up the seccomp filter in ubuntu 24.04 with syscall 21 (access)

Update (02/17/2025): 
We are continuing to look into alternatives for the "access" syscall and root causing why the seccomp filter only tripped up on Ubuntu 24.04 and not 22.04.


smittals2:
I see the patch has been merged in, closing the PR. Thank you!

PR #557, opened 2025-02-13
Pass -oControlMaster=no in scp

mrdomino:
Otherwise, some options (I noticed particularly ForwardAgent=no) will get applied across all future sessions on that control socket if the user has configured ControlMaster=auto for that host and we happen to be the session to start the socket.

I'm not 100% sure if I picked the right place to put this; maybe it belongs after ForwardAgent=no is appended later.

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -505,6 +505,7 @@ main(int argc, char **argv)
 	addargs(&args, "-oClearAllForwardings=yes");
 	addargs(&args, "-oRemoteCommand=none");
 	addargs(&args, "-oRequestTTY=no");
+	addargs(&args, "-oControlMaster=no");
 
 	fflag = Tflag = tflag = 0;
 	while ((ch = getopt(argc, argv,


daztucker:
Applied upstream and synced back as 5e60f59.  Thanks.

mrdomino:
Thank you!

deenoCO:
I do see this PR is closed but I was just reading the version 10.0/10.0p2.   I am not sure the appropriate protocol or procedure to continue discussion on this change.  Please advise where this should be besides this PR.

I see a motivation behind this change and the associated change to sftp.c in 5e60f59 to help users with simple configurations to get the longer term results that would be beneficial but this may be detrimental to other users and their use cases.   Specifically if I am using an editor/IDE to edit remote files using scp, then I use a persistent multiplexing connection so one is not created for every single file visited.  Adding the "-o<option>" to the command line overrides any user configuration even when it takes into account scp or sftp may be used.  I also see in the release notes "match sessiontype..." support.   My suggestion, which I have not investigated how to implement, would be to extend "match" or "match sessiontype" to specifically identify file transfer command is being used (I am unsure if both sftp and scp gets lumped into "subsystem") then let the user handle their own configuration.

I know a workaround for myself would be from a command line to ssh to the ssh to the target before using it from an editor/ide to visit remote files.  FWIW, I have a "pattern of ssh_config match aliases" used with specific configuration for copying files including creating a control master and multiplexed connection; the  changes in this PR will prevent that from continuing to work.   In the RnD envs I worked for ~40 years, it is very rare that DNS is configured for lab networks but also common that distinct lab networks reuse the same IP addresses which drove the "pattern of ssh_config match aliases" once ssh became available even before it was the dominant remote access method.

mrdomino:
Fwiw I did use `ssh $host true` prior to my `scp` commands as a workaround previously, after I figured out what was going on. It’s not hard to do, it’s just surprising and confusing and it took at least me a long time to figure out that that was the way to fix the “my SSH agent is mysteriously not working until I recreate my control socket” issue I was having.

Perhaps this is actually an argument for removing both `-oControlMaster=no` and `-oForwardAgent=no` from the options passed to scp. (Are there any other options passed to scp that are sticky across the whole control socket?)

deenoCO:
If I didn't read the release notes, I would have been trying to root cause why multiplexing & control master was not working.  But I did realize I just have to open an ssh connection without using scp/sftp.

I would "argue" that in general it can confusing to override users's configuration.  The flip side of this is ensuring there is enough information/flexibility for users to do the "correct thing" while having good defaults when they don't set the options (which in general I think ssh does well) but many users don't want to be bothered with doing the "correct thing".

I'll explore what sessiontype indicates (both in code and in use by config files) but that will have to wait until some free time (or less overbooked time) in about a month or so.  For now, I'm staying back on 9.9p2.    I also suspect I'm in a small minority of users that noticed this change and how it affect their editor/ide.

PR #558, opened 2025-02-27
Add support for memlocking sshd on Linux

d-tatianin:
Linux wakes up kcompactd threads in order to make more contiguous memory
available on the system, it does this by migrating live movable pages
(actively modifying live processes' page tables and constantly flooding
them with page invalidation IPIs, which can be up to millions per
second), which causes the process to become unresponsive for up to
seconds or even minutes in some severe cases. In case of sshd, we want
to always be able to connect to the system, even if it's under heavy
kcompactd load.

Introduce an option to protect sshd and its children sessions from being
compacted by kcompactd (this works in cojunction with
compact_unevictable_allowed = 0). Note that we depend on MCL_ONFAULT
being available, which was introduced in linux 4.4. MCL_ONFAULT allows
the system to lock pages lazily, thus drastically reducing memory usage
of a locked process (without MCL_ONFAULT, every existing mapping in the
process is instantly write-faulted).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -140,7 +140,7 @@ SSHD_SESSION_OBJS=sshd-session.o auth-rhosts.o auth-passwd.o \
 	auth2-gss.o gss-serv.o gss-serv-krb5.o \
 	loginrec.o auth-pam.o auth-shadow.o auth-sia.o \
 	sftp-server.o sftp-common.o \
-	uidswap.o $(SKOBJS)
+	uidswap.o platform-listen.o $(SKOBJS)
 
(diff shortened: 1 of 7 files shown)


d-tatianin:
CI failure looks unrelated to my changes (?)

d-tatianin:
Ping :)

daztucker:
Applied, thanks.

The CI failure is indeed unrelated (tcmalloc wants to use pipe2 under some conditions, which falls afoul of the process structure and sandboxing in a way we have not gotten to the bottom of yet.

PR #560, opened 2025-03-11
cipher-chachapoly: use OpenSSL's implementation of Poly1305

iucoen:
OpenSSL starting in version 3.0.0 has a new EVP_MAC API which allows POLY1305 algorithm to be used as a standalone MAC. OpenSSL's implementation is significantly faster, so we should use it when it's available.

On my RaspberryPi 3B+, with this change on an Intel N100 (Alderlake-N), I observed a 26% speedup from 276MB/s to 350MB/s.

### DIFF
--- a/cipher-chachapoly-libcrypto.c
+++ b/cipher-chachapoly-libcrypto.c
@@ -36,12 +36,18 @@
 
 struct chachapoly_ctx {
 	EVP_CIPHER_CTX *main_evp, *header_evp;
+#ifdef HAVE_EVP_MAC_INIT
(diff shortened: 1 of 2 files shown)


iucoen:
Any interest in taking this change?

djmdjm:
I've discussed this with one of the LibreSSL developers and I think we'll probably prefer to use the BoringSSL poly1305 API (which LibreSSL will also adopt), and add a shim to support EVP_MAC in OpenSSL. I'll wait until the LibreSSL people let me know once it's ready to look at.

PR #561, opened 2025-03-17
Log session length in disconnection message

uuumisha:
imho: It's convenient to know how long the session lasted.


djmdjm:
If you turn on `LogLevel=verbose` this information is already available on the client

```
Connection to blah closed.
Transferred: sent 4228, received 4588 bytes, in 2.0 seconds
Bytes per second: sent 2114.0, received 2294.0
```

FreezB11:
@djmdjm this thread can be closed init?

PR #565, opened 2025-04-16
ssh-agent: exit 0 from SIGTERM under systemd socket-activation

dkg:
When the ssh-agent service is configured to be launched under systemd socket-activation, the user can inspect the status of the agent with something like:

    systemctl --user status ssh-agent.service

If the user does:

    systemctl --user stop ssh-agent.service

it causes the `systemd --user` supervisor to send a SIGTERM to the agent, which terminates while leaving the systemd-managed socket in place.  That's good, and as expected. (If the user wants to close the socket, they can do "systemctl --user stop ssh-agent.socket" instead)

But because ssh-agent exits with code 2 in response to a SIGTERM, the supervisor marks the service as "failed", even though the state of the supervised service is exactly the same as during session startup (not running, ready to launch when a client connects to the socket).

This change makes ssh-agent exit cleanly (code 0) in response to a SIGTERM when launched under socket activation. This aligns the systemd supervisor's understanding of the state of supervised ssh-agent with reality.

This was also [reported on the mailing list](https://lists.mindrot.org/pipermail/openssh-unix-dev/2025-April/041884.html).

### DIFF
--- a/ssh-agent.c
+++ b/ssh-agent.c
@@ -2238,6 +2238,7 @@ main(int ac, char **av)
 	size_t npfd = 0;
 	u_int maxfds;
 	sigset_t nsigset, osigset;
+	int socket_activated = 0;
 
 	/* Ensure that fds 0, 1 and 2 are open or directed to /dev/null */
 	sanitise_stdfd();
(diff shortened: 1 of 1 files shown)


dkg:
a gentle nudge here -- it'd be great to get feedback on this.  If there's a problem, i'm happy to try to resolve it.

PR #567, opened 2025-05-05
Ignore bad hostkeys in known_hosts file

ZoltanFridrich:
When a short ssh-rsa hostkey is present in known_hosts, the connection will fail with the following error: `Bad server host key: Invalid key length`
What happens is that the ssh-rsa key is found in known_hosts file and is picked as prefered keyalg. Then when server sends its short rsa key, the verify_hostkey check fails and the error is printed. The connection fails even if valid keys are present on the server.

This patch changes the behavior to ignore bad hostkeys when scanning known_hosts file. The result is that a helpful message is printed into debug (example: `debug2: record_hostkey: /root/.ssh/known_hosts:1: ignoring hostkey: Invalid key length`) and the connection can be established by accepting new keys.

### DIFF
--- a/hostfile.c
+++ b/hostfile.c
@@ -63,6 +63,14 @@
 #include "hmac.h"
 #include "sshbuf.h"
 
+static int required_rsa_size = SSH_RSA_MINIMUM_MODULUS_SIZE;
+
+void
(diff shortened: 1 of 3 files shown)


djmdjm:
Looks good to me. @daztucker ok?

PR #568, opened 2025-05-09
Specify the default for MaxStartups later in the text

sanvila:
Hello. While reading about `MaxStartups` in `sshd_config(5)`, I noticed that the first paragraph explains the case where the option has a single value, then specifies the default value which has three numbers, then explains the case where the value has three numbers.

This is a little bit misleading and I think there is room for improvement. For example, the default value could be specified at the end, after the three numbers mode has been explained. This is what the MR does, but maybe it's still not good enough, as there is some redundancy between the example `e.g. "10:30:60"` and the default value which is given immediately afterwards.

(So, take the MR with a grain of salt and feel free to do any other rewording as you see it fits, it's just a way to communicate where is the problem).

Thanks.

### DIFF
--- a/sshd_config.5
+++ b/sshd_config.5
@@ -1367,11 +1367,11 @@ SSH daemon.
 Additional connections will be dropped until authentication succeeds or the
 .Cm LoginGraceTime
 expires for a connection.
-The default is 10:30:100.
 .Pp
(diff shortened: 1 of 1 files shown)


djmdjm:
Thanks - this has been merged as 7e68771f3cad61ec67303607afb3b85639288b29

PR #570, opened 2025-06-01
Support writing ED25519 keys in PKCS8 format

jbrobst:
Also make attempts to write ED25519 keys in PEM format an error.


djmdjm:
Merged as 2d0d26602f739b4a3ddde6c4dbc8f3ddab38ac0d - thanks

PR #573, opened 2025-06-12
sshbuf_dup_string: fix incorrect overflow check for l + 1

afonot:
This PR removes a dead-code overflow check in sshbuf_dup_string() and replaces it with a correct guard.

**What was wrong**

- l is a size_t; the expression l > SIZE_MAX is always false, so the protection never triggered.
- If l == SIZE_MAX, the expression l + 1 overflows to 0; allocating 0 bytes and then copying l bytes invokes undefined  behavior 

**Alternative considered — remove the length check entirely**

The sshbuf layer already enforces the invariant len <= SSHBUF_SIZE_MAX, so in normal operation l can never reach SIZE_MAX. In principle we could therefore drop the overflow guard and keep only the s == NULL test.

### DIFF
--- a/sshbuf-misc.c
+++ b/sshbuf-misc.c
@@ -254,7 +254,7 @@ sshbuf_dup_string(struct sshbuf *buf)
 	size_t l = sshbuf_len(buf);
 	char *r;
 
-	if (s == NULL || l > SIZE_MAX)
+	if (s == NULL || l >= SIZE_MAX)
 		return NULL;
 	/* accept a nul only as the last character in the buffer */
 	if (l > 0 && (p = memchr(s, '\0', l)) != NULL) {


sshbuf-misc.c
```diff
@@ -254,7 +254,7 @@ sshbuf_dup_string(struct sshbuf *buf)
 	size_t l = sshbuf_len(buf);
 	char *r;
 
-	if (s == NULL || l > SIZE_MAX)
+	if (s == NULL || l == SIZE_MAX)
```
daztucker, line 257: Looks OK to me.  As to removing the check altogether: I think it's worth keeping the check in case the external invariant changes at a later date.  @djmdjm ?

djmdjm: Maybe making `>=` expresses the intent better here while fixing the off-by-one?

djmdjm:
Looks good to me

sshbuf-misc.c
afonot: Done, thank you!

daztucker:
Applied upstream, thanks.

PR #577, opened 2025-06-20
mux: fix incorrect return value check in local forward cancellation

afonot:
channel_cancel_lport_listener() returns 1 on success and 0 on failure. The previous code incorrectly checked for `== -1`, a value the function never returns, so failure was not detected and the "port not found" error message was never shown when cancelling dynamic or local port forwards.


djmdjm:
This was merged as fd10cea0f16e928ae2b52fbeadccd475d0438eb4 a couple of months ago

PR #578, opened 2025-06-20
mdoc2man: support `Ns` inside `Ic`

raboof:
When encountering an `Ns` mdoc macro ('no space') inside an `Ic` block ('command'), such as for 'lines=number' in ssh-keygen.1, `mdoc2man` just output the macro instead of processing it.

This adds processing for `Ns` when seen inside an `Ic` block.


daztucker:
Applied, thanks.

PR #579, opened 2025-06-21
mdoc2man: process `Dl` macros

raboof:
`Dl` marks a single line as 'literal'. Since we don't output single lines differently in literal vs regular mode (we only insert line breaks for multi-line blocks in literal mode), we can just skip it.


daztucker:
Applied, thanks.

PR #580, opened 2025-06-27
sftp-server: handle errors from readdir()

msuhanov:
Hello.

On some platforms, the readdir() call may fail and set 'errno' to a security-related value.

In particular, on FreeBSD, 'errno' is set to EINTEGRITY if the underlying file system driver touched a directory block that failed the HMAC-based verification on an integrity-protected volume (i.e., a GELI volume with the authentication algorithm enabled).

On Linux, this is EIO for dm-verity and dm-integrity volumes (e.g., LUKS2 using AEAD). There is no way to distinguish between media/storage errors and authentication failures, except through reading the logs...

(I verified that this issue can be reproduced on both platforms.)

The current code ignores possible readdir() errors, treating them as the "end-of-directory" condition:

```
while ((dp = readdir(dirp)) != NULL) {
    [...]
}
if (count > 0) {
    send_names(id, count, stats);
    [...]
} else {
    send_status(id, SSH2_FX_EOF);
}
```

In such a case, an empty directory or a partial directory (depending on the platform and the file system driver used) is sent to the SFTP client, with no indication (at all) of this server-side error condition (which is, possibly, security-related).

Note: the only way to distinguish between "end-of-directory" and error conditions is to reset the 'errno' variable and, after the call, examine it.

### DIFF
--- a/sftp-server.c
+++ b/sftp-server.c
@@ -1131,7 +1131,7 @@ process_readdir(u_int32_t id)
 	DIR *dirp;
 	struct dirent *dp;
 	char *path;
-	int r, handle;
+	int r, handle, status = SSH2_FX_EOF;
 
 	if ((r = get_handle(iqueue, &handle)) != 0)
 		fatal_fr(r, "parse");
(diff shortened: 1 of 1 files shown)


djmdjm:
I think it is ok, except that I think we might need something similar in on the client side; see `upload_dir_internal()` in sftp-client.c

djmdjm:
@daztucker ok?

daztucker:
looks fine to me

PR #587, opened 2025-08-17
sftp: fix fatal when complete_match() truncates UTF8 multibyte character

TaoistBrickscarrier:
While sftp try to auto complete multibyte UTF8 file name list, 
complete_match() compares filenames one byte by one byte, 
and may truncates a prefix contains incomplete UTF8 multibyte character which leads to the fatal. 

Fix it in complete_ambiguous(), verify the bytes after match position, go back to beginning of the last valid UTF8 character.

### DIFF
--- a/sftp.c
+++ b/sftp.c
@@ -1876,6 +1876,7 @@ complete_ambiguous(const char *word, char **list, size_t count)
 
 	if (count > 0) {
 		u_int y, matchlen = strlen(list[0]);
+		u_int itemlen = matchlen, wordlen = strlen(word);
 
 		/* Find length of common stem */
 		for (y = 1; list[y]; y++) {
@@ -1887,6 +1888,10 @@ complete_ambiguous(const char *word, char **list, size_t count)
 
 			matchlen = x;
 		}
+
+		for (; matchlen > wordlen; matchlen--)
+			if (mblen(list[0] + matchlen, itemlen - matchlen) >= 0)
+				break;
 
 		if (matchlen > strlen(word)) {
 			char *tmp = xstrdup(list[0]);


djmdjm:
This looks ok to me. @daztucker ok?

djmdjm:
Here's a demonstration fwiw

```shell
$ cd /tmp                                                                       
$ printf "test-\xf0\x9f\x98\x9f" | xargs touch                                  
$ printf "test-\xf0\x9f\x98\x9d" | xargs touch                                  
$ ls test-*                                                                     
$ sftp -D /usr/libexec/sftp-server                                              
Attached to /usr/libexec/sftp-server.                                           
sftp> cp test-[hit tab here]                                                    
test-😝  test-😟                                                                
invalid multibyte character   
```

djmdjm:
Merged as 1f7556753869654ba5e2bf61e384c5da2db5ca6a - thanks!

djmdjm:
Unfortunately we had to back this fix out before the 10.1 release as it breaks the usual case for tab completion:

> sftp> ls
> foo      foo_bar
> 
> Now tab complete:
> 
> sftp> get fo<tab>
> foo      foo_bar
> 
> It does not complete 'get fo' to 'get foo' like it did before but it
> remains stuck at 'get fo', and if you do

So there is some kind of off-by-one error in this code.

TaoistBrickscarrier:
> Unfortunately we had to back this fix out before the 10.1 release as it breaks the usual case for tab completion:
> 
> > sftp> ls
> > foo      foo_bar
> > Now tab complete:
> > sftp> get fo
> > foo      foo_bar
> > It does not complete 'get fo' to 'get foo' like it did before but it
> > remains stuck at 'get fo', and if you do
> 
> So there is some kind of off-by-one error in this code.

That's odd, I have tried several times on different device, nothing like this happened.
But I studied the mblen(). 
With second parameter of value 0, mblen() should return 0 according to the ISO C 89,
but in some version mblen() is "restarted" to mbrlen() and returns -2.

I've made a check.

  Regards!

djmdjm:
I have a revised fix that I'll commit after the 10.1 release

djmdjm:
Revised fix committed as 0f3b8fd68a - thanks!

PR #589, opened 2025-08-22
fix: prevent OOM when pw_gecos is NULL on Android.

chinese0cabbage:
This PR addresses a **segmentation fault** caused by accessing an uninitialized `pw_gecos` field in `struct passwd` when using the Android NDK cross-compilation toolchain (`android-ndk-r26c`). While the macro `HAVE_STRUCT_PASSWD_PW_GECOS` is defined (indicating support for the `pw_gecos` field), Android's Bionic libc implementation returns `NULL` for `pw_gecos` in `getpwuid()`. This discrepancy leads to a crash in functions `xtrdup call strlen(str)` when dereferencing the null pointer.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -513,7 +513,7 @@ pwcopy(struct passwd *pw)
 	copy->pw_name = xstrdup(pw->pw_name);
 	copy->pw_passwd = xstrdup(pw->pw_passwd == NULL ? "*" : pw->pw_passwd);
 #ifdef HAVE_STRUCT_PASSWD_PW_GECOS
-	copy->pw_gecos = xstrdup(pw->pw_gecos);
+	copy->pw_gecos = xstrdup(pw->pw_gecos == NULL ? "" : pw->pw_gecos);
 #endif
 	copy->pw_uid = pw->pw_uid;
 	copy->pw_gid = pw->pw_gid;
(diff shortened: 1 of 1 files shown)


djmdjm:
An empty string would be better IMO, i.e.

```C
copy->pw_gecos = xstrdup(pw->pw_gecos == NULL ? "" : pw->pw_gecos);
```

While you're there, could you do the other string fields too? (excluding `pw_name`)

chinese0cabbage:
> An empty string would be better IMO, i.e.
> 
> ```c
> copy->pw_gecos = xstrdup(pw->pw_gecos == NULL ? "" : pw->pw_gecos);
> ```
> 
> While you're there, could you do the other string fields too? (excluding `pw_name`)

Thank you for your guidance. I approve of this solution, but the pw_passwd field has been filled with *. As a reference, I filled the pw_gecos field with null strings. According to your instructions, I performed null pointer checks on pw_gecos,pw_class,pw_dir, and pw_shell.

chinese0cabbage:
sorry to bother,how could I make this PR merged

djmdjm:
This has been merged as d343df4019b4369ce7f87e9bf6bbc80b81cd263d. I reworded the commit message for accuracy.

Thanks!

PR #592, opened 2025-09-01
Upgrade ssh_connection_hash from SHA1 to SHA256

xnox:
Upgrade ssh_connection_hash from SHA1 to SHA256, if length of this
value is considered an ABI, can also keep tohex legnth as before to
thus effectively use SHA256/160.

This change enables building and using ssh completely without SHA1.

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -362,7 +362,7 @@ ssh_connection_hash(const char *thishost, const char *host, const char *portstr,
 	struct ssh_digest_ctx *md;
 	u_char conn_hash[SSH_DIGEST_MAX_LENGTH];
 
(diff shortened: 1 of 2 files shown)


daztucker:
> if length of this value is considered an ABI, can also keep tohex length as before to thus effectively use SHA256/160.

I'ts not exactly an ABI, but it's often used in ControlPath Unix domain sockets, and those are usually quite restricted in length, eg on Linux that's 108 bytes:

$ grep UNIX_PATH_MAX /usr/include/linux/un.h 
#define UNIX_PATH_MAX	108
	char sun_path[UNIX_PATH_MAX];	/* pathname */

so increasing the size of this will render currently-working configurations broken.

PR #595, opened 2025-09-07
Upgrade ssh_connection_hash from SHA1 to SHA256

xnox:
Upgrade ssh_connection_hash from SHA1 to SHA256. Due to increased
length, instead of using hex encoding, use base64url encoding, which
for SHA256 is only slightly longer than hex SHA1.

This change enables building and using ssh completely without SHA1.

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -51,6 +51,7 @@
 
 #include "xmalloc.h"
 #include "ssh.h"
+#include "sshbuf.h"
 #include "ssherr.h"
 #include "cipher.h"
 #include "pathnames.h"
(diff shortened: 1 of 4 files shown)


sivizius:
base85(sha256(foo)) should be exactly the size of hex(sha1(foo)), is this an option?

xnox:
> base85(sha256(foo)) should be exactly the size of hex(sha1(foo)), is this an option?

From mailing lists, one of the requirements that it has to be URL and file-system safe name, base85 seems to use \ and / which are not safe on filesystems - things can fail with "no such directory" because of slashes.

sivizius:
> From mailing lists, one of the requirements that it has to be URL and file-system safe name

I see, then nevermind.

PR #596, opened 2025-10-02
include openssl/bn.h explicitly in more places

vapier:
These files use BN APIs, so explicitly include bn.h rather than rely on other openssl headers (sometimes) including it for us.


djmdjm:
Merged as a9cbe10da2b - thanks

PR #597, opened 2025-10-03
sshkey: check if mmap is supported

vapier:
If mmap isn't available, just use basic malloc functions.


djmdjm:
@daztucker has tried to move us away from checking HAVE_FOO_H tests recently. Is it possible to test HAVE_MMAP directly?

vapier:
the current autoconf doesn't check for mmap so the HAVE_MMAP symbol doesn't exist, but it'd be trivial to change the logic to check for the func

vapier:
is this version what you had in mind ?

sshkey.c
```diff
@@ -723,6 +725,7 @@ sshkey_sk_cleanup(struct sshkey *k)
 static int
 sshkey_prekey_alloc(u_char **prekeyp, size_t len)
 {
+#ifdef HAVE_MMAP
```
daztucker, line 728: you can have mmap() but not MAP_ANON or MAP_PRIVATE, so it needs to check those too (see openbsd-compat/arc4random.h) for an example, (although it doesn't check HAVE_MMAP and probably should).

configure.ac
```diff
@@ -2455,6 +2455,8 @@ AC_CHECK_FUNCS([endutxent getutxent getutxid getutxline getutxuser pututxline])
 AC_CHECK_FUNCS([setutxdb setutxent utmpxname])
 dnl    Checks for lastlog functions
 AC_CHECK_FUNCS([getlastlogxbyname])
+dnl    Various system functions
+AC_CHECK_FUNCS([mmap])
```
daztucker, line 2459: Test that aren't platform specific should go in the big alphabetized AC_CHECK_FUNCS list where possible.

daztucker (commented):
This is close enough that I can rejigger this locally to apply without updating the PR, but what system are you building on that doesn't have mmap()?

daztucker:
Applied my modified version of your PR, please confirm that it works for your use case.

vapier:
> but what system are you building on that doesn't have mmap()?

[WASM/WASI](https://wasi.dev/) for running OpenSSH on the web or in [Chromium browsers](https://chromewebstore.google.com/detail/iodihamcpbpeioajjeobimgagajmlibd).

I could hack up my own mmap emulation, but would only be able to support very limited settings (basically a thin wrapper on top of malloc).  I debated going that route, but considering the usage here and other checks OpenSSH already has, seemed easier to check for it since I know the semantics this particular bit of code wants.

> Applied my modified version of your PR, please confirm that it works for your use case.

thx, will try to check it later today.

vapier:
it doesn't quite work because it's unconditionally including sys/mman.h which doesn't exist, but if i have to stub that header out, OK.

there is a bug in `sshkey_prekey_alloc` though ... the final `return 0` should be moved to the bottom of the func.

```
../sshkey.c:741:1: warning: non-void function does not return a value [-Wreturn-type]
  741 | }
      | ^
```

daztucker:
> it doesn't quite work because it's unconditionally including sys/mman.h

oh, I typoed the header when I added it to configure.ac :-(

fixed that and the return, does this now work for your use case?

vapier:
thx, can confirm it works now

PR #616, opened 2025-12-18
Providing support of PKCS#11 URI

beldmit:
This is a rebase of patch provided in #2817 by @Jakuje to OpenSSH 10.2/master branches. 
The patch has been tested in RHEL/Fedora for a while and I'm not aware of any problems

### DIFF
--- a/ssh.c
+++ b/ssh.c
@@ -859,6 +859,14 @@ main(int ac, char **av)
 			options.gss_deleg_creds = 1;
 			break;
 		case 'i':
+#ifdef ENABLE_PKCS11
+			if (strlen(optarg) >= strlen(PKCS11_URI_SCHEME) &&
+			    strncmp(optarg, PKCS11_URI_SCHEME,
+			    strlen(PKCS11_URI_SCHEME)) == 0) {
+				add_identity_file(&options, NULL, optarg, 1);
+				break;
+			}
+#endif
 			p = tilde_expand_filename(optarg, getuid());
 			if (stat(p, &st) == -1)
 				fprintf(stderr, "Warning: Identity file %s "
@@ -1831,6 +1839,7 @@ main(int ac, char **av)
 
 #ifdef ENABLE_PKCS11
 	(void)pkcs11_del_provider(options.pkcs11_provider);
+	pkcs11_terminate();
 #endif
 
  skip_connect:
@@ -2342,6 +2351,45 @@ ssh_session2(struct ssh *ssh, const struct ssh_conn_info *cinfo)
 	    options.escape_char : SSH_ESCAPECHAR_NONE, id);
 }
 
+#ifdef ENABLE_PKCS11
+static void
+load_pkcs11_identity(char *pkcs11_uri, char *identity_files[],
+    struct sshkey *identity_keys[], int *n_ids)
+{
+	int nkeys, i;
+	struct sshkey **keys;
+	struct pkcs11_uri *uri;
+
+	debug("identity file '%s' from pkcs#11", pkcs11_uri);
+	uri = pkcs11_uri_init();
+	if (uri == NULL)
+		fatal("Failed to init PKCS#11 URI");
+
+	if (pkcs11_uri_parse(pkcs11_uri, uri) != 0)
+	fatal("Failed to parse PKCS#11 URI %s", pkcs11_uri);
+
+	/* we need to merge URI and provider together */
+	if (options.pkcs11_provider != NULL && uri->module_path == NULL)
+		uri->module_path = strdup(options.pkcs11_provider);
+
+	if (options.num_identity_files < SSH_MAX_IDENTITY_FILES &&
+	    (nkeys = pkcs11_add_provider_by_uri(uri, NULL, &keys, NULL)) > 0) {
+		for (i = 0; i < nkeys; i++) {
+			if (*n_ids >= SSH_MAX_IDENTITY_FILES) {
+				sshkey_free(keys[i]);
+				continue;
+			}
+			identity_keys[*n_ids] = keys[i];
+			identity_files[*n_ids] = pkcs11_uri_get(uri);
+			(*n_ids)++;
+		}
+		free(keys);
+	}
+
+	pkcs11_uri_cleanup(uri);
+}
+#endif /* ENABLE_PKCS11 */
+
 /* Loads all IdentityFile and CertificateFile keys */
 static void
 load_public_identity_files(const struct ssh_conn_info *cinfo)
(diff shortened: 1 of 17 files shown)


beldmit:
@bob-beck I hopefully resolved the conflict - can we please move this forward?

beldmit:
@bob-beck any chances to move forward here?

beldmit:
@bob-beck I've splitted the commit into several independent parts for better reviewability

bob-beck:
OK so revisiting this - Today, you can link a pkcs11 provider library that will hand you pkcs11 keys via the helper in the existing way.  

Obviously you want to add a way to fetch such keys directly via a pkcs11 URI scheme. Is that necessary to be in here?

Or should this be hidden away in your pkcs11 library you are linking and all ssh and the helper sees are the keys as it sees today? 

So, It's hard for me to review this and potentially justify it to other openssh developers without more detail on what
specifically you are trying to accomplish that the current way of linking a library and the helper does not let you do. 

IF you had some sort of internal design document when you did this that had "man, the current ssh pkcs11 helper does not let us do X Y and Z" - we need to add features, and here's what we considered, that would be very helpful to link here, or alternatively, Gimme some help understanding what it is you want here that does not work in the current helper/shlib model? 

So you can either add that detail here or toss in a link to somehwere, don't particularly care, but I think I need a little
more info on what capabilites you specifically want this to let you do, that does not work now.

bob-beck:
> OK so revisiting this - Today, you can link a pkcs11 provider library that will hand you pkcs11 keys via the helper in the existing way.
> 
> Obviously you want to add a way to fetch such keys directly via a pkcs11 URI scheme. Is that necessary to be in here?
> 
> Or should this be hidden away in your pkcs11 library you are linking and all ssh and the helper sees are the keys as it sees today?
> 
> So, It's hard for me to review this and potentially justify it to other openssh developers without more detail on what specifically you are trying to accomplish that the current way of linking a library and the helper does not let you do.
> 
> IF you had some sort of internal design document when you did this that had "man, the current ssh pkcs11 helper does not let us do X Y and Z" - we need to add features, and here's what we considered, that would be very helpful to link here, or alternatively, Gimme some help understanding what it is you want here that does not work in the current helper/shlib model?
> 
> So you can either add that detail here or toss in a link to somehwere, don't particularly care, but I think I need a little more info on what capabilites you specifically want this to let you do, that does not work now.

and a reminder, especially in this crowd, "Do X as per RFC ABCD" doesnt' really tell us what you want out of it, RFC's have a habit of being kitchen sinks of every functionality someone thought sounded good in an ivory tower, and overall by RFC volume there is, IMO,  generally a low probability an individual one won't just be a bad idea in general :)

ssh.c
```diff
@@ -2373,33 +2416,46 @@ load_public_identity_files(const struct ssh_conn_info *cinfo)
...
+	pkcs11_init(!options.batch_mode);
+
+	if (options.pkcs11_provider != NULL) {
+		struct pkcs11_uri *uri;
+
+		uri = pkcs11_uri_init();
+		if (uri == NULL)
+			fatal("Failed to init PKCS#11 URI");
+
+		/* Construct simple PKCS#11 URI to simplify access */
```
bob-beck, line 2429: So unless  I am reading this wrong, and very well may be, this takes the provider as today, and instead of just using it add an identity file and fetch the keys from it, it adds it's url to the list (along with potentially some others you have added in the config) and then loads them all by the URI. 

Assuming I'm correct in that (if not tell me that's not how it works) What's the reasoning for
exposing this in ssh, and not handing off the potential existence of multiple uri's to get keys to the
pkcs11 provider, pulling them all in there from whatever bag of pkcs11 uri locations are desired, and
then feeding it all to ssh the way it is now?  (I.e. why should this complexity be in ssh and not the provider?)

beldmit:
@bob-beck OK, I will prepare the information you requested in a digestable form and will reach you again

Jakuje:
The original idea was a convenience standard way of referencing objects on tokens. Note, that this does not really change the current helper/shlib model, but extends it to have better mechanism of selecting objects.

The main motivation was:
 * selecting specific key on token -- if you have more keys on token (for example issued by third party), you can easily run out of authentication tries as the existing code tries all the keys provided by the pkcs11 module. I think there was some workaround of having the selected public key in local file and using IdentitiesOnly to prefer that specific key, but it was undocumented and awkward.
 * having multiple tokens/cards -- by selecting specific key (or token/slot) you can avoid sending wrong keys or input wrong pin and locking card/token. Without this change, openssh blindly iterates over all the slots and potentially prompts for PIN.
 * the object selection could be probably achieved by different means, but using standard PKCS#11 URI scheme looked like obvious solution, rather than inventing different OpenSSH-specific selection mechanism. You can easily get these URIs from existing tools (pkcs11-tool, p11tool, p11-kit ...) and most of the tools and applications use these in their CLI or configuration (for example curl, wget, nginx, apache).

> Or should this be hidden away in your pkcs11 library you are linking and all ssh and the helper sees are the keys as it sees today?

the pkcs11 library does not have any way of selecting objects -- the application has to do that through the PKCS#11 API.

I tried to describe this also in some blob posts couple of years back with some examples, but I would be happy to clarify more if needed:

https://fedoramagazine.org/fedora-28-better-smart-card-support-openssh/

ssh.c
Jakuje: > So unless I am reading this wrong, and very well may be, this takes the provider as today, and instead of just using it add an identity file and fetch the keys from it, it adds it's url to the list (along with potentially some others you have added in the config) and then loads them all by the URI.

I think this is correct. 

I think there is a bit confusion about the PKCS11Provider configuration option in OpenSSH, which is just a path to the pkcs11 module (.so) and pkcs11-provider from OpenSSL. The current approach was written some 10 years ago before OpenSSL got any provider concept. While I would love to use the pkcs11-provider here, the process separation in OpenSSH would make it very complicated and it would not work for other forks (I think they did not get providers yet?).

As mentioned in the other comment, the idea is not to flood the application with all objects the pkcs11 module provides. There is no other (clean) mechanism to do that in the existing architecture.

beldmit:
@bob-beck I believe your questions are answered

Jakuje:
And one more thing I wanted to mention is the usability -- the existing approach requires the user to juggle in the configuration or in CLI paths to /usr/lib where the specific pkcs11 module lives (and what the user should not bother with). This approach specifies also default (non-mandatory) PKCS#11 module (p11-kit-proxy), which is used to juggle these paths for the user. This is very useful in the cases where more different pkcs11 modules are present and configured in the system, where you just throw `pkcs11:` to `ssh` and all works, rather than figuring out the path to the `.so` under `/usr/lib` (or `/lib64`? Or is it under `/usr/lib/x86_64-linux-gnu`?).

gitmpr:
Adding a real-world motivation for this: with two PIV YubiKeys plugged in at once, with different PINs, ssh-add -s <module> currently tries the single entered PIN against every slot the module reports, one after another, before any URI-style filtering exists. If the PIN doesn't match a given token, that's a failed C_Login on it, i.e. a burned PIV PIN retry attempt on a key you never intended to touch. Three wrong attempts locks that token's PIN and forces a PUK recovery.

This patch fixes exactly that: pkcs11_register_provider_by_uri() skips non-matching slots by serial/token/manufacturer before pkcs11_open_session()/C_Login is ever called, so an unrelated token is never touched. Confirmed this by reading the diff against current master: `ssh-pkcs11.c`, `ssh-add.c`, and reproducing the underlying problem locally. `ssh-add -s` on vanilla OpenSSH 10.4p1 has no URI parsing and will always loop-login every enumerated slot.

Multi-token PIV setups are increasingly common, for examle with a personal + work YubiKeys, or backup keys kept plugged in. Would be good to get this merged, or at least get maintainer feedback on what's blocking it.

PR #621, opened 2026-01-14
optimize random calls

vapier:
randombytes is implemented on top of arc4random_buf which, depending on the system, might involve a syscall to get bytes.  Linux can often optimize this with a vsyscall to lessen the pain.

The urandom32 function will grab 4 bytes and then manually convert them to a 32-bit number instead of just writing directly.

The Short_random & Small_random both call urandom32 p times (where p is defined as 761) to fill another buffer.  This means 1500+ calls to get 4 random bytes at a time.  Both functions can simply fill the buffer in one go which means 2 calls to get 4*761 bytes each, and then drop the unoptimized urandom32 function entirely.

### DIFF
--- a/sntrup761.c
+++ b/sntrup761.c
@@ -1961,25 +1961,17 @@ static void Hash_prefix(unsigned char *out, int b, const unsigned char *in, int 
   for (i = 0; i < 32; ++i) out[i] = h[i];
 }
 
-static uint32_t urandom32(void) {
-  unsigned char c[4];
-  uint32_t result = 0;
-  int i;
-  randombytes(c, 4);
-  for (i = 0; i < 4; ++i) result += ((uint32_t)c[i]) << (8 * i);
-  return result;
-}
-
 static void Short_random(small *out) {
   uint32_t L[p];
-  int i;
-  for (i = 0; i < p; ++i) L[i] = urandom32();
+  randombytes(L, sizeof(L));
   Short_fromlist(out, L);
 }
 
 static void Small_random(small *out) {
   int i;
-  for (i = 0; i < p; ++i) out[i] = (((urandom32() & 0x3fffffff) * 3) >> 30) - 1;
+  uint32_t L[p];
+  randombytes(L, sizeof(L));
+  for (i = 0; i < p; ++i) out[i] = (((L[i] & 0x3fffffff) * 3) >> 30) - 1;
 }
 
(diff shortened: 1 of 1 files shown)


daztucker:
This code comes directly from supercop with only minimal massaging (mostly types) done by sntrup761.sh, so these changes would be lost on the next update.  I agree it's not the most efficient way to do this, but does this actually make a measurable difference?

vapier:
it's measurable for me under WASM because arc4random_buf is a direct syscall which means 1500+ getrandom syscalls during initial connect.  for comparison, all other syscalls combined barely make it to 500.

I'm certainly not reading this file for fun :).  it's really rough.

i've never heard of supercop before.  do you think this is something i could get merged there (wherever "there" is) ?

daztucker:
> i've never heard of supercop before. do you think this is something i could get merged there (wherever "there" is) ?

Main page is https://bench.cr.yp.to/supercop.html, mirrored at https://github.com/jedisct1/supercop.  As for feeding back changes, dunno, I'm not sure if we've ever tried that.

vapier:
welp that was a pita, but I think I managed to send an e-mail to `ebats@list.cr.yp.to` with a patch to all the relevant kmem files.  no idea where/if they have an archive to check if it made it through.

djmdjm:
We could also work this change into `sntrup761.sh` and apply it locally while we wait

daztucker:
> We could also work this change into `sntrup761.sh` and apply it locally while we wait

It was a bit painful to do in sed, but this does that: http://www.dtucker.net/tmp/ssh-sntrup-rand.patch

djmdjm:
ok djm

(though I'm surprised this code doesn't zero stack allocations that seem like private data - this is preexisting though)

daztucker:
> (though I'm surprised this code doesn't zero stack allocations that seem like private data - this is preexisting though)

Well the array L on the stack in Small_random() is new, previously there would only have been a single uint32_t that was overwritten.  I've added an explicit_bzero(L, sizeof(L)) at the end of both functions.

vapier:
do you guys want me to update the PR, or are you on top of things and I won't sweat it?

daztucker:
> do you guys want me to update the PR, or are you on top of things and I won't sweat it?

no it's fine as is.  I'm about to commit it to OpenBSD upstream then sync the changes back into Portable.

daztucker:
This is now in.  Thanks for your work.

PR #643, opened 2026-03-22
vm.yml: fix solaris PAM tests and add retry for VM hangs

renaudallard:
Add missing SSHD_CONFOPTS="UsePam yes" to the solaris PAM test step so it actually tests PAM functionality instead of re-running the default tests with a PAM-enabled binary.

The Solaris QEMU VM intermittently hangs during tests (a VM-level issue also affecting upstream master).  Add retry logic: each test step has a 30-minute timeout with continue-on-error, followed by a retry step that runs if the first attempt failed.  This handles the intermittent hang rate reliably without masking real test failures.

Tested successfully 7 times in a row (including runs where the retry mechanism was exercised after a VM hang).

### DIFF
--- a/.github/workflows/vm.yml
+++ b/.github/workflows/vm.yml
@@ -349,7 +349,17 @@ jobs:
       shell: solaris {0}
       run: cd $GITHUB_WORKSPACE && sudo -u builder make
     - name: make tests
+      id: test1
       shell: solaris {0}
+      timeout-minutes: 30
+      continue-on-error: true
+      run: |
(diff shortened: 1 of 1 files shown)


daztucker:
Hi!  I've applied the PAM fix part since that was definitely wrong.  I didn't apply the retry part, because while it's clever the config file is already a bit of a monster.   Ideally I'd like to see the underlying hang fixed, but failing that I think a shell loop with a `timeout` command would fix it in a simpler way?

renaudallard:
Yes, `timeout` is another idea.

PR #647, opened 2026-03-30
Fix unsafe atoi() usage in get_sock_port()

NotUnHackable:
Replace atoi() with strtonum() for safer port number conversion. atoi() has undefined behavior on overflow and no error detection, while strtonum() provides proper bounds validation and error reporting.


djmdjm:
I'm not sure this is worth fixing. The port string in this case comes from getnameinfo(3), so if it is returning garbage then the system is already hosed.

PR #648, opened 2026-03-31
MaxStartups parsing fix and refactor

xypiie:
The commit 683d0ab introduced a conversion issue, which converted
"MaxStartups 3" into "10:30:3".  This is wrong, as value1 > value3 and
it causes side effects leading to crash of sshd when MaxStartups
functionality is triggered with this error:
"error: mm_reap: child terminated by signal 15"

Please find fix proposal and a proposal for refactoring the parsing part to be more robust against such issues.

### DIFF
--- a/servconf.c
+++ b/servconf.c
@@ -2015,12 +2015,12 @@ process_server_config_line_depth(ServerOptions *options, char *line,
 				    filename, linenum, keyword);
 		} else if (n == 1) {
 			value3 = value;
-			value = value2 = -1;
+			value2 = -1;
 		} else {
 			fatal("%s line %d: Invalid %s spec.",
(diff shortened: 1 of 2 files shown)


djmdjm:
This was merged as 78d549857e0cc - thanks

xypiie:
> This was merged as [78d5498](https://github.com/openssh/openssh-portable/commit/78d549857e0cc480c3cbb0a3571078920e3b79c5) - thanks

@djmdjm , thanks. what about the refactoring improvement (https://github.com/openssh/openssh-portable/pull/648/changes/cfc7c0748f8860b983c78a9b97c8256f7ef4770c)? has this been considered?

PR #649, opened 2026-04-02
Fix error handling in fill_default_options

ZoltanFridrich:
Found during a bug investigation in CentOS 10 where our downstream patch removes kex algorithm in FIPS mode but ignores the error when encountered in `kex_names_valid` on purpose. This propagated the error further eventually hitting an error in `fill_default_options` but instead of exiting, the error is ignored and ssh keeps running in broken state.

This seems to be a bug in upstream code where `ret` is never set in `fill_default_options` even when error is encountered.
Looking at the history the `r` variable was introduced in commit 1b9dd4a where the error was fatal. Then in commit 43026da the error handling was moved outside of the function. The commit introduced `ret` variable which is never assigned into. This makes the function always return 0, even on error.

### DIFF
--- a/readconf.c
+++ b/readconf.c
@@ -2849,7 +2849,7 @@ fill_default_options(Options * options)
 {
 	char *all_cipher, *all_mac, *all_kex, *all_key, *all_sig;
 	char *def_cipher, *def_mac, *def_kex, *def_key, *def_sig;
-	int ret = 0, r;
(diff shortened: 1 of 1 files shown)


djmdjm:
Thanks, I went with a slightly different fix in c8716138281d3dac33c2c7e98615c52d2b606d85 but the result should be the same.

PR #655, opened 2026-04-16
ci: match all OpenSSL versions in setup_ci.sh, not just 3.x

### LINKED ISSUE #1933: ci: fix interop openssh pipeline

npajkovsky:
The case pattern for constructing the OpenSSL target only matched versions starting with 3. Change to a wildcard default so future OpenSSL major versions (e.g. 4.x) work without further edits.


npajkovsky:
Failing openssl interop pipeline is here https://github.com/openssl/openssl/actions/runs/24491252112/job/71576524065

daztucker:
There was an earlier PR that did the same thing that we hadn't got to (PR#639) so I've applied that one.  Thanks.

PR #656, opened 2026-04-16
Make ci work^Wattempt to run for OpenSSL 4 variants

bob-beck:
Assuming I did this right this is correct for 4 going forward and maintains the bypass for the 3.X version
breakage.


bob-beck:
@daztucker, @quarckster - I think this is needed.

.github/install_libcrypto.sh
bob-beck, line 70: Without the above change we attempt to check out "4.0"  which does not exist.

daztucker:
Applied, thanks.

PR #657, opened 2026-04-16
sk-usbhid: skip unsupported key types in read_rks()

akhilesharora:
## Summary

- In `read_rks()`, encountering a resident credential with an unsupported COSE key type (anything other than ES256 or EdDSA) causes `goto out`, aborting the entire resident key enumeration. All valid keys - both already collected and remaining - are discarded.
- This means a single unsupported credential on a FIDO2 token causes `ssh-add -K` / `ssh-keygen -K` to return zero keys, even if other valid ecdsa-sk or ed25519-sk credentials exist.
- The fix moves the key type check before the per-credential allocation, so unsupported types are skipped with `continue` instead of aborting the loop. No memory is allocated before the check, so `continue` is safe without additional cleanup.
- The existing `XXX` comment at the old `goto out` site (`/* XXX free rk and continue */`) documents the intended fix.

## Details

The old code allocated `srk` (including `key_handle`, `application`, `user_id`) *before* checking the key type. When the switch hit the default case, `goto out` jumped past the `ret = 0` assignment, returning `SSH_SK_ERR_GENERAL`. The caller `sk_load_resident_keys()` then freed all previously collected keys and returned an error.

The new code checks `fido_cred_type()` first, stores the result in a local `uint32_t alg`, and only proceeds with allocation if the type is supported. For supported types, behavior is unchanged.

## Test plan

- Compiled with `./configure --with-security-key-builtin && make` - zero warnings with `-Wall -Wextra -Wuninitialized -Wsign-compare`
- Existing regression test suite passes
- Wrote a standalone C test modeling the exact loop structure that proves: old code aborts on unsupported key type (losing valid keys), new code skips and continues (preserving valid keys)
- Verified no behavior change when all credentials use supported types

### DIFF
--- a/sk-usbhid.c
+++ b/sk-usbhid.c
@@ -1262,6 +1262,7 @@ read_rks(struct sk_usbhid *sk, const char *pin,
     struct sk_resident_key ***rksp, size_t *nrksp)
 {
 	int ret = SSH_SK_ERR_GENERAL, r = -1, internal_uv;
+	uint32_t alg;
 	fido_credman_metadata_t *metadata = NULL;
 	fido_credman_rp_t *rp = NULL;
 	fido_credman_rk_t *rk = NULL;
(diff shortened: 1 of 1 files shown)


akhilesharora:
openbsd-current-upstream fails in `cd /usr/src/usr.bin/nc && make` at `netcat.c:1660: undeclared IPTOS_DSCP_VA`. This check was green on the last 15 master runs (through 2026-05-13). The failure is in OpenBSD's nc source, not in any file this PR touches.

djmdjm:
Thanks, I've merged this upstream and synced it back as 684d26a3c4d07e78027517fd6e98875ae5dff71d

akhilesharora:
Thanks!

PR #658, opened 2026-04-25
CircleCI Commit

Ikonic1z420:


daztucker:
What is this and why would we want it?

Ikonic1z420 (commented):
connects ssh to cloud gpu service

daztucker:
> connects ssh to cloud gpu service

openssh doesn't use GPUs or cloud services so no thanks.

Ikonic1z420 (commented):
apparently some people want ot

PR #659, opened 2026-05-02
Timing Attack Vulnerability in Password Authentication

JobLuxe-prod:
image

### DIFF
--- a/auth-passwd.c
+++ b/auth-passwd.c
@@ -79,48 +79,22 @@ auth_password(struct ssh *ssh, const char *password)
 	Authctxt *authctxt = ssh->authctxt;
 	struct passwd *pw = authctxt->pw;
 	int result, ok = authctxt->valid;
-	size_t i;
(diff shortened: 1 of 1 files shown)


daztucker:
This whole thing is based on a false premise: the password isn't secret to an "attacker" since it's supplied by them, and MAX_PASSWORD_LEN is a *CONSTANT* so by definition not secret either.  Which tool/LLM did you use to generate this?

JobLuxe-prod:
> This whole thing is based on a false premise: the password isn't secret to an "attacker" since it's supplied by them, and MAX_PASSWORD_LEN is a _CONSTANT_ so by definition not secret either. Which tool/LLM did you use to generate this?

Thanks for the clarification. I agree that MAX_PASSWORD_LEN and the input password are not secret. My concern was about the hash comparison using strcmp(), but since hashing is done first, the timing difference isn’t really exploitable.
So this is not a real vulnerability, just a minor point.

PR #660, opened 2026-05-09
channels: use computed poll events for distinct socket fd

Komzpa:
## Summary

channel_prepare_pollfd() computes a pollfd event mask for a channel socket, but the branch for a socket fd distinct from rfd stores zero in the pollfd events field.

This stores the computed mask instead, matching the rfd, wfd, and efd branches.

## Details

With events set to zero, poll()/ppoll() will not report normal read or write readiness for that socket; only exceptional conditions such as HUP, ERR, or NVAL can wake the entry. channel_after_poll() already filters readiness through the corresponding SSH_CHAN_IO_SOCK_R and SSH_CHAN_IO_SOCK_W masks, so this only arms the events the channel already requested.

This was found while reviewing an sshd-session channel wait-loop symptom, but this PR only claims the local channel polling bug fixed here.

## Validation

- autoreconf
- ./configure
- make channels.o
- make -j2
- git diff --check -- channels.c

### DIFF
--- a/channels.c
+++ b/channels.c
@@ -2858,7 +2858,7 @@ channel_prepare_pollfd(Channel *c, u_int *next_pollfd,
 		if (ev != 0) {
 			c->pfds[3] = p;
 			pfd[p].fd = c->sock;
-			pfd[p].events = 0;
+			pfd[p].events = ev;
 			dump_channel_poll(__func__, "sock", c, p, &pfd[p]);
 			p++;
 		}


djmdjm:
This seems correct, but now I'm wondering why we haven't seen problems from this before. What was the symptom that led you to this?

Komzpa:
Thanks. The symptom was not a crash or a common user-visible failure; it came
from tracing a forwarding path that was repeatedly waking without useful fd
work.

While reducing that, I noticed this narrower issue: when `channel->sock` is a
distinct fd, its pollfd can be armed from the generic channel state instead of
the events computed for that fd. In the usual cases where the fds are the same,
or `sock` is not separately used, the mistake is masked, which is probably why
it has not shown up as an obvious long-standing failure.

I can add a small regression test if you would prefer one before merge.

djmdjm:
Thanks, this has been committed upstream and synced back as de24573efc60c

PR #663, opened 2026-05-17
auth2: avoid delay amplification after slow auth

Komzpa:
When an authentication attempt takes longer than its requested user-specific failure delay, `ensure_minimum_time_since()` currently doubles the requested delay until it exceeds the elapsed runtime. That can add another artificial sleep even though the slow authentication work has already satisfied the minimum delay.

Return immediately when the elapsed runtime already meets the requested delay, and keep the existing sleep path only for the remaining time below that minimum.

Validation:
- `autoreconf && ./configure && make -j2 auth2.o sshd-auth`
- `git diff --check -- auth2.c`
- Started `make tests`; build, file-tests, and a substantial regression prefix passed (connect/proxy/protocol/rekey/dhgex/scp/sftp/keygen/agent blocks) before I stopped the full suite manually to avoid running the entire long regression set in this local worktree.

Fixer observed evidence for this was a sampled `sshd-auth` process spending CPU in the failed-auth timing path; I have not independently reproduced that exact trace outside the Fixer capture.

### DIFF
--- a/auth2.c
+++ b/auth2.c
@@ -222,7 +222,6 @@ input_service_request(int type, uint32_t seq, struct ssh *ssh)
 }
 
 #define MIN_FAIL_DELAY_SECONDS 0.005
-#define MAX_FAIL_DELAY_SECONDS 5.0
 static double
 user_specific_delay(const char *user)
 {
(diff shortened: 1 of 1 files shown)


Komzpa:
Closing this because the changed code is part of OpenSSH's failed-authentication timing behavior. The current PR frames the change as a straightforward delay-amplification fix, but it does not sufficiently prove that the existing bucket/scale-up behavior is unintended rather than security-relevant anti-enumeration/throttling behavior. This needs a more careful security-sensitive analysis or bug report before proposing a code change.

PR #671, opened 2026-05-28
openbsd-compat: preserve fractional precision in scan_scaled()

metsw24-max:
Fix incorrect fractional parsing in `scan_scaled()` for large-unit suffixes such as `E` and `P`.

The previous implementation truncated the fractional component before scaling to avoid overflow:

```c
while (fpart >= LLONG_MAX / scale_fact ||
    fpart <= LLONG_MIN / scale_fact) {
	fpart /= 10;
	fract_digits--;
}
```

For large scale factors (especially `E`), this could reduce `fpart` to zero before multiplication, causing fractional inputs like `0.9E` to parse incorrectly as `0`.

This patch replaces that behavior with an interleaved multiply/divide approach that preserves significantly more precision while keeping intermediate values bounded.

---

### `openbsd-compat/fmt_scaled.c`

Reworked fractional scaling logic in `scan_scaled()` to:

- avoid premature truncation of the fractional component
- preserve precision for large-unit suffixes
- interleave `*1024` and `/10` operations during scaling

The mathematical result remains:

```text
fpart * scale_fact / 10^(fract_digits - 1)
```

but intermediate computation is now performed incrementally instead of truncating first and scaling afterward.

### `openbsd-compat/regress/Makefile.in`

Added:

```make
scan_scaledtest$(EXEEXT)
```

to `TESTPROGS`.

---

## Examples

Previously broken inputs:

| Input | Pre-fix behavior |
|------|------|
| `0.9E` | parsed as `0` |
| `1.9E` | heavily truncated |
| `3.7E` | precision loss |
| `7.9E` | precision loss |

After this change, fractional values for large-unit suffixes retain their
expected scaled contribution instead of collapsing toward zero.
---

## Testing

Regression coverage was added for previously failing fractional inputs and
general sanity cases.

### DIFF
--- a/openbsd-compat/fmt_scaled.c
+++ b/openbsd-compat/fmt_scaled.c
@@ -186,18 +186,40 @@ scan_scaled(char *scaled, long long *result)
 			/* scale whole part */
 			whole *= scale_fact;
 
-			/* truncate fpart so it doesn't overflow.
-			 * then scale fractional part.
+			/*
+			 * Scale fractional part: compute
+			 *   fpart * scale_fact / 10^(fract_digits-1)
+			 * without intermediate overflow. scale_fact is 1024^i,
+			 * i.e. a power of 1024 = 2^10. Interleave
+			 * multiply-by-1024 with divide-by-10 so the running
+			 * value stays within long long range while preserving
+			 * precision -- the old code truncated fpart all the
+			 * way to zero for any fractional input with the E
+			 * (and sometimes P) unit, because for those units
+			 * LLONG_MAX/scale_fact is smaller than a single digit.
 			 */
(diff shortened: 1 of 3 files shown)


djmdjm:
Thanks, this code came from OpenBSD so I've submitted it upstream for review there. We'll pick up the change if/when they adopt it.

https://marc.info/?l=openbsd-tech&m=178030157126359&w=2

metsw24-max:
Got it thanks for the update.

metsw24-max:
@djmdjm any update on this?

djmdjm:
no, not yet

djmdjm:
Thanks, this was committed to OpenBSD's libutil and merged back as 47af21eb8d480c0f368a629ab58e31b993ed28f2

PR #679, opened 2026-05-31
Fix a critical vulnerability that could lead to system dumping and in…

MrSecurityGuy:
…fo disclosure via ssh -X

A vulnerability that dates back to 1999 that led to a null crash vulnerability causing info disclosure from a crash dump file from an unauthenticated user when you try ssh-X to connect to a server. someone running a malicious ssh server can trigger this by sending SSH2_MSG_CHANNEL_OPEN(x11) before the client sends x11-req, causing strlen(NULL) in x11_open_helper(). ... Leading to key's being leaked, an potential system hijack.

NULL Pointer Dereference → SIGSEGV → Information Disclosure → Key Theft → RCE

[exploit.py](https://github.com/user-attachments/files/28431889/exploit.py)

### DIFF
--- a/channels.c
+++ b/channels.c
@@ -1376,6 +1376,13 @@ x11_open_helper(struct ssh *ssh, struct sshbuf *b)
 	u_char *ucp;
 	u_int proto_len, data_len;
 
+	/* Null check. */
+    if (sc->x11_saved_proto == NULL) {
+        debug2("X11 open received before x11-req sent");
+        return -1;
+    }
+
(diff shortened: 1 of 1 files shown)


djmdjm:
Thanks, I've committed a fix in a1dd1c80687fbfbdca19d3a314a8176b90ae471b but none of the consequences you list of a simple local crash are true. There is no information disclosure, let alone key theft or RCE possible from this bug.

PR #681, opened 2026-06-02
fix compatibility-layer correctness and robustness issues

metsw24-max:
This PR fixes several correctness and robustness issues in `openbsd-compat/`.

The changes are limited to portability-layer code and include fixes for error handling, bounds validation, ownership cleanup, type-safety, and edge-case handling.

### bsd-misc.c

* Return the result of `kill()` from the `raise()` compatibility wrapper instead of falling off the end of a non-void function.

### port-aix.c

* Guard `aix_remove_embedded_newlines()` against empty strings to avoid decrementing past the start of the buffer.
* Improve `snprintf()` result handling in `aix_usrinfo()` by using a signed return type and validating the result before passing it to `usrinfo()`.

### fake-rfc2553.c

* Validate that the supplied socket address length is large enough for `struct sockaddr_in` before accessing address fields.

### port-prngd.c

* Reject requests larger than the maximum length accepted by `get_random_bytes_prngd()` to prevent problematic size narrowing from `size_t` to `int`.

### bsd-cygwin_util.c

* Free allocated memory when pattern conversion fails in `_match_pattern()`.
* Use `size_t` instead of `u_int` for string-length based iteration.

### getrrsetbyname-ldns.c

* Clarify ownership transfer after successful assignment to `*res`.
* Ensure resources are released correctly on error paths.

## Testing

* Verified that the changes are localized to `openbsd-compat/`.
* Reviewed affected control flow, ownership handling, and boundary conditions.
* Platform-specific code paths (AIX, Cygwin, LDNS, PRNGD compatibility code) were inspected for correctness, but were not compile-tested on their respective target platforms.

### DIFF
--- a/openbsd-compat/bsd-cygwin_util.c
+++ b/openbsd-compat/bsd-cygwin_util.c
@@ -198,8 +198,10 @@ _match_pattern(const char *s, const char *pattern)
 		return 0;
 	ws = (wchar_t *) xcalloc(len + 1, sizeof (wchar_t));
 	mbstowcs(ws, s, len + 1);
-	if ((len = mbstowcs(NULL, pattern, 0)) == (size_t) -1)
+	if ((len = mbstowcs(NULL, pattern, 0)) == (size_t) -1) {
+		free(ws);
 		return 0;
+	}
 	wpattern = (wchar_t *) xcalloc(len + 1, sizeof (wchar_t));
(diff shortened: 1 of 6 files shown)


metsw24-max:
gentle ping

djmdjm:
Thanks, most of these have been committed. There were a couple that were redundant or incorrect and we didn't commit those.

PR #683, opened 2026-06-04
Fix aesctr IV setup.

loganaden:


djmdjm:
I think we should just delete `cipher_set_keyiv()` entriely - nothing uses it

loganaden:
@djmdjm  diff updated.

djmdjm:
Fixed by deleting the code in 0d08d384a6783d3f3bc5a01d10015f527e0792ba - thanks :)

PR #688, opened 2026-06-05
Fix path truncation vulnerability in sftp-server readdir

sahvx655-wq:
Fix silent pathname truncation in sftp-server readdir that could cause lstat() to operate on an unintended file.
snprintf() was used to build path + "/" + d_name without checking buffer capacity. If the combined length exceeded pathname size, the result was silently truncated (see “XXX OVERFLOW ?” comment).

Add a pre-check to ensure the full pathname fits in the buffer. If it does not fit, skip that directory entry.

Testing:
Added a reproducible test (test_sftp_readdir_overflow.c) that:

shows truncation in the vulnerable version
confirms correct skipping after fix
Tested locally.

### DIFF
--- a/SFTP_PATH_TRUNCATION_FIX.md
+++ b/SFTP_PATH_TRUNCATION_FIX.md
@@ -1,0 +1,205 @@
+# OpenSSH SFTP Path Truncation Vulnerability - Fix Report
+
+## Vulnerability Summary
+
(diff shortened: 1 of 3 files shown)


djmdjm:
Thanks for the report. This is fixed in 70d7044a966a541d9df6002fbeb62d005555cc78

However, this is not a vulnerability. It most it might report incorrect stat information during ls-like operations, but this lstat is not used toactually grant or deny access to anything.

PR #690, opened 2026-06-12
Update monitor.c mm_answer_gss_userok: enforce prior GSSCHECKMIC

MrSecurityGuy:
mm_answer_gss_userok can be reached without mm_answer_gss_checkmic having been called first.

### DIFF
--- a/monitor.c
+++ b/monitor.c
@@ -90,6 +90,7 @@
 
 #ifdef GSSAPI
 static Gssctxt *gsscontext = NULL;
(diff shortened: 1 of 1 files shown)


djmdjm:
Thanks. Maybe we should just remove the no-mic path entirely? It's pretty legacy by now...

MrSecurityGuy:
Up to you what you do, if people still use it i'd say to fix just this, but if its mostly unused it could be removed.

PR #692, opened 2026-06-23
Add -U flag to enforce PID -> UID matchup before allowing forwarding

pschou:
Currently, the SSH `-D` flag enables a dynamic SOCKS proxy tunnel that accepts TCP connections from local clients and forwards them through the encrypted SSH connection. Currently, **all** connections to the dynamic tunnel are accepted regardless of who is connecting.  This means that processes running under a different UID (e.g., a system service running as `nobody`, or a container process with a different user) can also connect to the tunnel, which may be undesirable for security-sensitive use cases.

There is no built-in mechanism to restrict the dynamic tunnel to only accept connections from processes sharing the same UID as the SSH client process. The user may want to enable this restriction to prevent unauthorized processes from using the tunnel, similar to how `SO_PEERCRED` is used for agent socket permissions.

Add a new `enforce_same_uid` flag to the `ForwardOptions` struct that, when enabled, checks the peer's UID via `SO_PEERCRED` on every accepted connection to the dynamic tunnel. Connections from non-matching UIDs are silently rejected (the socket is closed before a new channel is created).

- **misc.h** — Add `enforce_same_uid` field to `struct ForwardOptions`. This stores the flag alongside other forwarding options (`gateway_ports`, `streamlocal_bind_mask`, `streamlocal_bind_unlink`).

- **readconf.h** / **readconf.c** — Add `oEnforceSameUid` to the `OpCodes` enum and the config keyword table. Add the `case` handler in `parse_config()`, the initialization in `initialize_options()`, the normalization in `fill_default_options()`, and the dump entry in `dump_client_config()`.

- **ssh.c** — Add the `-U` CLI flag that sets `options.fwd_opts.enforce_same_uid = 1`. Update the getopt short option string to include `U`. The config option `EnforceSameUid yes|no` is also supported via the standard config parser.

- **channels.c** — Two key changes:

  1. **GatewayPorts + EnforceSameUid constraint check**: In `channel_setup_fwd_listener_tcpip()`, after determining the bind address, check whether both `GatewayPorts` and `enforce_same_uid` are set. If so, `fatal()` with an error message explaining that opening the port globally (0.0.0.0) while enforcing UID makes the peer's PID/UID determination meaningless — when the port is global, remote peers cannot have their UID reliably determined via `SO_PEERCRED`. This prevents the semantic contradiction of a globally-opened port that only accepts same-UID connections (which cannot be verified).

  2. **enforce_same_uid applies to `-L` listening port connections too**: The existing `enforce_same_uid` check in `channel_post_port_listener()` uses `SO_PEERCRED` to get the peer's UID on every accepted connection. This check now applies to both `-D` (dynamic SOCKS) tunnels and `-L` (local forwarding) listening port connections when `-U` is specified. The peer's UID is compared against the current listener process's UID. If the peer's UID does not match, the socket is closed and the connection is rejected. This check is skipped for Unix domain socket listeners (which already have UID information via filesystem permissions).

### DIFF
--- a/channels.c
+++ b/channels.c
@@ -2062,6 +2062,23 @@ channel_post_port_listener(struct ssh *ssh, Channel *c)
 			c->notbefore = monotime() + 1;
 		return;
 	}
+	if (options.fwd_opts.enforce_same_uid &&
+	    c->type != SSH_CHANNEL_UNIX_LISTENER &&
+	    c->type != SSH_CHANNEL_RUNIX_LISTENER) {
+		struct ucred cred;
+		socklen_t len = sizeof(cred);
+
+		if (getsockopt(newsock, SOL_SOCKET, SO_PEERCRED, &cred,
+		    &len) == 0) {
+			if (cred.uid != (uid_t)getuid()) {
+				debug("channel %d: UID mismatch: peer %u != %u, "
+				    "rejecting", c->self,
+				    (u_int)cred.uid, (u_int)getuid());
+				close(newsock);
+				return;
+			}
+		}
+	}
 	if (c->host_port != PORT_STREAMLOCAL)
 		set_nodelay(newsock);
 	nc = channel_new(ssh, rtype, nextstate, newsock, newsock, -1,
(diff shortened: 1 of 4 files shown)


spagoveanu:
> Connections from non-matching UIDs are silently rejected (the socket is closed before a new channel is created).

since `SO_PEERCRED` only supports unix-domain sockets[1], and ssh's `-D` flag does not support unix-domain sockets, this either doesn't do anything or silently rejects **any** connection.

have you actually tested this at all? was this LLM-generated?

> with an error message explaining that opening the port globally (0.0.0.0) while enforcing UID makes the peer's PID/UID determination meaningless — when the port is global, remote peers cannot have their UID reliably determined via SO_PEERCRED.

wtf?

[1] on linux `getsockopt(SO_PEERCRED)` does succeed on TCP sockets, but always sets the pid to 0 and the uid/gid to -1

pschou:
It is not ready for review.

PR #693, opened 2026-06-29
sftp: set CLOEXEC on ssh subprocess IPC descriptors

peripherium:
This patch sets the relevant descriptors with `FD_CLOEXEC`, ensuring they are automatically closed when the shell is execl()'d.

I investigated the existing XXX comment and tested the behavior with multiple shells.

With `tcsh`, descriptor inheritance depends on how OpenSSH was built. `tcsh` closes inherited `pipe()` descriptors during startup, but does not close descriptors referring to `socketpair()`. As a result, when OpenSSH is built without `USE_PIPES` (using `socketpair()` instead), the IPC descriptor remain open. The relevant `tcsh` code is here:

https://github.com/tcsh-org/tcsh/blob/master/sh.misc.c#L285

With `fish`, neither inherited `pipe()` descriptors nor `socketpair()` descriptors are closed, so the IPC descriptors remain open in both configurations.

The inherited descriptors can be observed after invoking `!<cmd>`  using tools such as:
`lsof -p <pid>` or `ls -l /proc/<pid>/fd/`

### DIFF
--- a/sftp.c
+++ b/sftp.c
@@ -326,7 +326,6 @@ local_do_shell(const char *args)
 		fatal("Couldn't fork: %s", strerror(errno));
 
 	if (pid == 0) {
-		/* XXX: child has pipe fds to ssh subproc open - issue? */
 		if (args) {
 			debug3("Executing %s -c \"%s\"", shell, args);
 			execl(shell, shell, "-c", args, (char *)NULL);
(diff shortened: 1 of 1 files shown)


djmdjm:
Thanks, committed upstream and synced back as 710cd5abcea22aeddbd5d5d661c224eaa05e47b9

This will be in OpenSSH 10.4, due very soon

PR #694, opened 2026-06-29
add missing entries for MLDSA44ED25519 hybrid for sshfp

loganaden:


dns.h
```diff
@@ -33,7 +33,8 @@ enum sshfp_types {
 	SSHFP_KEY_RSA = 1,
 	SSHFP_KEY_DSA = 2,
 	SSHFP_KEY_ECDSA = 3,
-	SSHFP_KEY_ED25519 = 4
+	SSHFP_KEY_ED25519 = 4,
+	SSHFP_KEY_MLDSA44ED25519 = 5
```
daztucker, line 37: That value has not yet been assigned by IANA so adding this seems premature.

https://www.iana.org/assignments/dns-sshfp-rr-parameters/dns-sshfp-rr-parameters.xhtml

PR #695, opened 2026-06-30
Add missing mldsa44-ed25519 option description

loganaden:


djmdjm:
Similar patch applied - thanks

PR #697, opened 2026-07-01
AIX: PAM enablement to set  options.use_pam based on AIX system autht…

ajaykini:
…ype setting

 On AIX systems, the operating system's own authentication framework can be configured to use PAM (Pluggable Authentication Modules) or a native AIX authentication stack. This setting is held in the AIX system-wide login configuration attribute SC_AUTHTYPE (under SC_SYS_LOGIN).

Previously, OpenSSH's PAM behaviour on AIX was controlled solely by the UsePAM directive in sshd_config — independently of what the host OS was actually configured to use. This mismatch could cause authentication to succeed via the SSH daemon but still fail (or behave unexpectedly) at the OS session/credential level if the two settings diverged.

This change adds an AIX-specific post-authentication check that reads the host OS authtype setting at runtime and uses it as an additional gate on options.use_pam — PAM is only active when both the OS is configured for PAM_AUTH and sshd_config has UsePAM yes. If either says no, PAM is disabled.


sshd-session.c
```diff
@@ -1275,6 +1289,23 @@ main(int ac, char **av)
...
+#ifdef _AIX
+	retval = getconfattr(SC_SYS_LOGIN, SC_AUTHTYPE, &authType, SEC_CHAR);
+	if (retval != 0 || authType == NULL) {
+		error("AIX getconfattr(SC_AUTHTYPE) failed (ret=%d, errno=%d: %s, authtype=%s): "
+		    "using sshd_config UsePAM setting", retval, errno, strerror(errno),
+		    authType != NULL ? authType : "(null)");
+	} else {
+		debug3("AIX system authtype: %s", authType);
+		if (strcmp(authType, "PAM_AUTH") == 0 && options.use_pam)
+			options.use_pam = 1;
```
daztucker: This seems much too late in the process to be doing this kind of config tweaking.

If you have UsePAM=no and SC_AUTHTYPE=PAM_AUTH it'll force PAM on here and call the PAM setcred and session but won't have called the PAM auth stack even for password authentication.  If you have UsePAM=yes and SC_AUTHTYPE!=PAM_AUTH then it'll do the PAM auth part but not the PAM session part.

It seems like they should be either both done or neither done.

Also, putting platform-specific code in the middle of code we need to keep in sync with OpenBSD is a maintenance headache.  For this reason AIX specific code should go in openbsd-compat/port-aix.c where possible, and calls to it should go via the platform-specific hooks in the platform*.c files.  You could probably call this check from platform_pre_listen().

ajaykini: All the above review comments has been adressed @daztucker.

ajaykini:
@daztucker , I have refactored the code base as per review comments and tested. Please let me know if the changes looks good. 
Also not sure why one of the CI tests are failing. I guess the failure has to do nothing w.r.t the Fix.

ajaykini:
Hi @daztucker , following up on the PR. The review comments have been addressed and the updated changes have been available for some time. Please let me know if any further action is required from my side ? else, can we proceed with merging this out? Thanks

Ajay Kini K

ajaykini:
Hi,

Just following up on this PR.

PR #697 proposes the AIX-specific PAM enablement needed to correctly set options.use_pam based on the AIX system authentication configuration.

The PR has been kept limited to the AIX-specific changes.

Could a maintainer please take a look when possible and let me know whether this approach is suitable for upstream, or if there are any changes you'd prefer before considering it?

In particular, I'm happy to rework the implementation if there is a preferred approach for handling the AIX PAM/authentication configuration.

Thanks.
Ajay Kini K

PR #702, opened 2026-07-28
Fix error condition in ssh_ed25519_generate

xnox:
The call to generate ed25519 keypair can fail, ensure to return error
instead of writing out a key.


xnox:
https://github.com/openssh/openssh-portable/commit/4b5860ed27f758c9d8156714cbd8a62562420907

PR #712, opened 2026-09-16
auth-options: only serialise placeholder strings when option is non-NULL in untrusted mode

vpashaiev:
When authenticating in untrusted mode, `sshauthopt_serialise()` unconditionally serialised "true" for `force_command` and "yes" for `cert_principals` regardless of whether the option was actually set.

This caused the unprivileged child process to deserialise `force_command` as a valid non-NULL string even when no forced command was present on the key in `authorized_keys`. Under configurations like `PermitRootLogin forced-commands-only` + `UsePAM yes`, the child considered root login accepted for forced command and proceeded with PAM account management (`MONITOR_REQ_PAM_ACCOUNT` / msg 102). 

However, the privileged monitor still held `force_command = NULL` and rejected root login, leading to a fatal assertion:
`fatal: monitor_child_preauth: unexpected authentication from 102`

This PR fixes the issue by only serialising "true" or "yes" when the corresponding field in `sshauthopt` is non-NULL. If the field is NULL, NULL is serialised and the child correctly detects that no forced command is configured.

Tested with `PermitRootLogin forced-commands-only` + `UsePAM yes`:
- Key without forced command: correctly rejected without fatal monitor crash.
- Key with forced command: correctly accepted and executed.


vpashaiev:
Note: the two failures in `xcode-27` (`default` and `pam`) are caused by the known flaky `regress/scp3.sh` test (`Broken pipe` during remote-to-remote scp copy), and are unrelated to the `auth-options.c` serialization changes (all other 116 matrix builds passed cleanly).

